PR #20, opened 2017-07-04
Updated input dataset for integration tests using latest NEURON version.

pramodk:
We were using legacy hoc scripts (bluron trunk) for data generation. This
is now replaced by latest python ringtest model. IClamp test is removed
and replaced by gap juction test.

P.S. We will merge this after bbp jenkin nodes work correctly (tomorrow morning)


pramodk:
- Jenkins configuration done and working
- test datasets added are smaller size (~50KB, similar to old tests)

This can be now reviewed.

Issue #3, opened 2016-10-13
InterleaveInfo breaks rule of 3 (5)


juanchopanza:
By pure chance, I came across class [`InterleaveInfo`](https://github.com/BlueBrain/CoreNeuron/blob/ba1fc53e2cd6fe8ac06e9afeba9b7504035f7fb8/coreneuron/nrniv/cellorder.h#L9), which seems to do some resource management. Its [destructor releases a host of resources via `delete[]`](https://github.com/BlueBrain/CoreNeuron/blob/ba1fc53e2cd6fe8ac06e9afeba9b7504035f7fb8/coreneuron/nrniv/cellorder.cpp#L36), however, no special care is taken of copy construction or assignment, which could result in memory errors at runtime (specifically, double-deletes, but formally, undefined behaviour.) I suggest either making the class non-copyable, or implementing the correct copy and assignment semantics.

---

See [the rule of three/five/zero](http://en.cppreference.com/w/cpp/language/rule_of_three).

pramodk:
Sorry, didn't get time to look into this before due to other stuff. Addressed in #22.

PR #22, opened 2017-07-17
Implement copy constructor and assignment operator for InterleaveInfo

pramodk:
Probably InterleaveInfo won't be used in other contexts but this addresses seggestions in #3 (rule of 3).


pramodk:
understood, agree. I will merge this to address suggestions in #3.

PR #26, opened 2017-07-27
Added option to keep ion_global_map after nrn_cleanup and use nrnmpi_under_nrncontrol in mpi initialisation

brunomaga:
Some fixes and added options related to clean-up and init of data

### DIFF
--- a/coreneuron/nrniv/nrn_setup.cpp
+++ b/coreneuron/nrniv/nrn_setup.cpp
@@ -787,20 +787,19 @@ void nrn_cleanup(bool clean_ion_global_map) {
     gid2in.clear();
     gid2out.clear();
 
-    //clean ezOpt parser allocated memory (if any)
+    // clean ezOpt parser allocated memory (if any)
     nrnopt_delete();
 
-    //clean ions global maps
-    if (clean_ion_global_map)
-    {
-      for (int i=0; i<nrn_ion_global_map_size; i++)
-        free(nrn_ion_global_map[i]);
-      free(nrn_ion_global_map);
-      nrn_ion_global_map = NULL;
-      nrn_ion_global_map_size=0;
+    // clean ions global maps
+    if (clean_ion_global_map) {
+        for (int i = 0; i < nrn_ion_global_map_size; i++)
+            free(nrn_ion_global_map[i]);
+        free(nrn_ion_global_map);
+        nrn_ion_global_map = NULL;
+        nrn_ion_global_map_size = 0;
     }
 
-    //clean NrnThreads
+    // clean NrnThreads
     for (int it = 0; it < nrn_nthread; ++it) {
         NrnThread* nt = nrn_threads + it;
         NrnThreadMembList* next_tml = NULL;
--- a/coreneuron/nrniv/nrn_setup.cpp
+++ b/coreneuron/nrniv/nrn_setup.cpp
@@ -269,7 +269,6 @@ void read_phase1(data_reader& F, int imult, NrnThread& nt) {
     // the extra netcon_srcgid will be filled in later
     netcon_srcgid[nt.id] = new int[nt.n_netcon + nrn_setup_extracon];
     F.read_array<int>(netcon_srcgid[nt.id], nt.n_netcon);
-    F.close();
 
 #if 0
   // for checking whether negative gids fit into the gid space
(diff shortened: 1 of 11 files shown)


pramodk:
I didn't understand the option "...under_nrn_control". 

Also have used same clang format for code formatting?

pramodk:
I think tests are failing with this change but I can't cerify this without laptop. Check with make test locally on your laptop (before/after this PR).

brunomaga:
Tests were failing due to `coreneuron/nrnmpi/mpispike.c` : In the `void nrnmpi_multisend_comm()`, the `MPI_Comm_dup` should be applied to the global `nrnmpi_world_comm` variable which is the `MPI_Comm` the is being used everywhere else.

Yet, it only works with `MPI_COMM_WORLD` instead, which I believe is wrong but passes the tests, so I left it that way.

brunomaga:
The `nrnmpi_undernrncontrol` was already part of Coreneuron, hard-coded as `1`. See the previous call to `nrnmpi_init`. It serves as a *hint* to tell CN that the MPI initialisation is being handled anywhere else. I parameterised it by making it a user-defined option, defaulted to 1.

pramodk:
@brunomaga :
 
> Yet, it only works with MPI_COMM_WORLD instead, which I believe is wrong but passes the tests, so I left it that way.

Your error is coming from : 

```C
MPI_Comm_dup(&nrnmpi_world_comm, &multisend_comm);
```

Note that `MPI_Comm_dup` takes `MPI_Comm` as an argument and not `MPI_Comm*`.

pramodk:
@brunomaga : as I have fixed clang-format issue, you have to rebase your branch to master, and do `make formatsource` and update the branch. (clang-format should be there on your PC).

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1632,6 +1646,7 @@ for (int i=0; i < nt.end; ++i) {
         }
         nt._vecplay[i] = new VecPlayContinuous(ml->data + ix, yvec, tvec, NULL, nt.id);
     }
+    F.close();
```
pramodk: why we need to close it here? If I understood correctly, it's taken care by nrn_setup.h:

```c
file_reader_w[i].close();
```

coreneuron/nrniv/nrnoptarg.cpp
```diff
@@ -112,7 +112,7 @@ static param_str param_str_args[] = {
 
 static void graceful_exit(int);
 
-static ez::ezOptionParser* opt;
+static ez::ezOptionParser* opt = NULL;
```
pramodk: this doesn't hurt but `static pointer is always initialised to null` by default.

coreneuron/nrnoc/register_mech.c
```diff
@@ -150,6 +150,7 @@ void alloc_mech(int n) {
         int i;
         for (i = 0; i < memb_func_size_; ++i) {
             nrn_mech_data_layout_[i] = 1;
+            memb_func[i].alloc = NULL; //mark as non-instantiated
```
pramodk: not necessary because it's `calloc`ed and hence zero/null initialised already.

coreneuron/nrnmpi/nrnmpi.c
```diff
@@ -113,11 +107,10 @@ void nrnmpi_init(int nrnmpi_under_nrncontrol, int* pargc, char*** pargv) {
             nrn_assert(MPI_Init(pargc, pargv) == MPI_SUCCESS);
 #endif
         }
-
-        { nrn_assert(MPI_Comm_dup(MPI_COMM_WORLD, &nrnmpi_world_comm) == MPI_SUCCESS); }
     }
     grp_bbs = MPI_GROUP_NULL;
     grp_net = MPI_GROUP_NULL;
+    nrn_assert(MPI_Comm_dup(MPI_COMM_WORLD, &nrnmpi_world_comm) == MPI_SUCCESS);
```
pramodk: as Michael suggested this should be `nrnmpi_world_comm`

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -784,10 +784,24 @@ inline void mech_layout(data_reader& F, T* data, int cnt, int sz, int layout) {
  * allocated with malloc(). This is not the case here, so let's try and fix
  * things up first. */
 
-void nrn_cleanup() {
+void nrn_cleanup(bool clean_ion_global_map) {
```
pramodk: just to understand: is this change because you want to call `nrn_cleanup` and still want to have access to `ion_global_map`?

brunomaga: yes

coreneuron/nrniv/nrn_setup.cpp
brunomaga: ok

PR #27, opened 2017-08-02
Fix clang-format indentation for OpenMP and OpenACC pragmas

pramodk:
OpenMP/OpenACC pragma is more like programming cnstruct rather
than just hint and hence we would like to have those indented.
clang-format doesn't recognize these "special" pragmas and end-up
removing all indendation. Currently there is no way to fix
this rather than using clang-format off/on comments.


pramodk:
Adding 

```
// clang-format off
....
// clang-format one
```

is still bit ugly but I gone through mailing list discussion and there is *no* other elegant way (as far as I remember, [discussion](https://bugs.llvm.org/show_bug.cgi?id=17362)) 

With this PR, at least everyone will be able to do `make formatsource` on the code before creating pull request. This will avoid current situation of different indentation from different people.

PR #28, opened 2017-08-02
Fix compilation issue on Cray platforms (shared builds)

pramodk:
Allow building shared/static variants woth Cray / Intel compiler.
Cray platform is detected by checking /opt/cray which seeems sufficient
on all available platforms like Daint, Titan and Theta.


pramodk:
> I can test it on Theta I guess ?

I encountered these issues while compiling on Theta. If you are in middle of any other task, go ahead with that. Once data transfer complete we can discuss benchmarks for Theta.

PR #29, opened 2017-08-03
update RPath config for Cray (static linkage)

fouriaux:
cmake try to setup rpath of coreneuron when doing make install, however we build coreneuron statically on cray systems.


pramodk:
@fouriaux : just to confirm, could you check if everything works fine with `SHARED` compile library type flag? I don't see any reason though...

pramodk:
@fouriaux : your PR has 3 commits (including from Bruno which are already merged). So I would suggest to just push single relevant commit (?)

pramodskumbhar:
cherry-picking your relevant commit and replacing this with #33

PR #30, opened 2017-08-03
Checkpoint restart prototype

fouriaux:


fouriaux:
Hi, Is it possible to allow comit squash for this repository ? 
That will ease process when merging..

PR #35, opened 2017-08-21
Mem info fix

fouriaux:


fouriaux:
when trying to perform make formatsource I got the following error:
clang-format: Unknown command line argument '-fallback-style=none'.  Try: '/usr/bin/clang-format -help'

pramodk:
The issue is with version of `clang-format`. I think we needed `>=3.9`. Could you check what you have?

fouriaux:
ok, I have just fixed that using clang-format-3.9

coreneuron/utils/memory_utils.cpp
```diff
@@ -70,12 +73,26 @@ double nrn_mallinfo(void) {
...
+#else
+    std::ifstream file;
+    std::stringstream buffer;
+    file.open("/proc/self/statm");
+    if (file.is_open()) {
+        buffer << file.rdbuf();
+        file.close();
+        unsigned long long int total_virtual_memory;
+        buffer >> total_virtual_memory;
+        mbs = (total_virtual_memory * sysconf(_SC_PAGESIZE)) / (1024.0 * 1024.0);
```
pramodk: I am wondering if we should use VM or RSS. (e.g. in hpctools  we considered only RSS. While using mallinfo I have seen that considering `mmap` and shared library related memory usage gives very different numbers).

How your measurements look for RSS vs VM?

coreneuron/utils/memory_utils.cpp
```diff
@@ -70,12 +73,26 @@ double nrn_mallinfo(void) {
...
-#elif defined HAVE_MALLOC_H
-    struct mallinfo m;
-    m = mallinfo();
-    mbs = (m.hblkhd + m.uordblks) / (1024.0 * 1024.0);
+#else
+    std::ifstream file;
+    std::stringstream buffer;
+    file.open("/proc/self/statm");
+    if (file.is_open()) {
+        buffer << file.rdbuf();
```
pramodk: not necessary to read entire file : `file >> total_virtual_memory` should do?

coreneuron/utils/memory_utils.cpp
fouriaux: They will for sure provide very different numbers:
On a Linux x86 workstation, RSS will provides what have been accessed, not what have been allocated.
On an HPC system with KNL: I have to check but it seems that we try to keep everything in Physical memory and the policy might be to reserve physical space directly. 
I will check if this behavior is correct.
Also I have checked that "data" section in statm is growing properly with allocation, so instead of full virtual memory, we can take this figure (without code and shared).

pramodk: ok, fine with me. Go ahead and update it.

fouriaux:
I have just updated with changes required

pramodk:
@fouriaux : I was testing this today on Lugano viz cluster and inclined for using RSS instead of VM. Here are numbers:

For Small dataset of 25MB:

```
# dataset size
$ du -h /gpfs/bbp.cscs.ch/scratch/gss/viz/kumbhar/ring/coreneuron_data_25mb/
23M	/gpfs/bbp.cscs.ch/scratch/gss/viz/kumbhar/ring/coreneuron_data_25mb/

# using this PR (VM)
 Memory (MBs) :          After nrn_setup  : Max 196.0391, Min 196.0391, Avg 196.0391

# using mallinfo
 Memory (MBs) :          After nrn_setup  : Max 25.9347, Min 25.9347, Avg 25.9347

# using RSS (i.e. 2nd value in statm)
 Memory (MBs) :          After nrn_setup  : Max 43.0586, Min 43.0586, Avg 43.0586
```

For slightly larger dataset of 180MB:

```
# dataset size
$ du -h /gpfs/bbp.cscs.ch/scratch/gss/viz/kumbhar/ring/coreneuron_data_180mb/
186M	/gpfs/bbp.cscs.ch/scratch/gss/viz/kumbhar/ring/coreneuron_data_180mb/

# using this PR (VM)
 Memory (MBs) :          After nrn_setup  : Max 468.9609, Min 468.9609, Avg 468.9609

# using mallinfo
 Memory (MBs) :          After nrn_setup  : Max 203.2387, Min 203.2387, Avg 203.2387

# using RSS (i.e. 2nd value in statm)
 Memory (MBs) :          After nrn_setup  : Max 318.0586, Min 318.0586, Avg 318.0586

# using massif 

time=5446904827
mem_heap_B=214147430         => ~ 210 MB
mem_heap_extra_B=79250
mem_stacks_B=0
heap_tree=empty

# DDT 

~ 220 MB
```

For now we don't need very precise numbers anyway. So I am updating this PR with RSS. Feel free to revert if you see anything missing.

PR #36, opened 2017-08-22
Spike buffer realloc

fouriaux:
* use std::vector instead of new allocated memory, permitting easy realloc. 
* filter negative gid inside spike_buffer (shouldnt it be already filtered in setup anyway ?)


coreneuron/nrnoc/multicore.c
```diff
@@ -444,22 +444,23 @@ void nrn_thread_table_check() {
 void nrn_multithread_job(void* (*job)(NrnThread*)) {
     int i;
 #if defined(_OPENMP)
-
+    printf("\n..openmp \n");
```
pramodk: these changes for `multicore.c` are coming from your local changes...

fouriaux:
I have just reverted those changes on multicore.c

fouriaux:
After validation, everything is working as expected.

fouriaux:
I have just updated with changes required

PR #51, opened 2017-10-11
Mod2c as submodule

fouriaux:


fouriaux:
I haven't changed the findmod2c as it is not possible to add ExternalProject_add in it without significant changes, that seems to be something aside to think about. Not on current path.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "external/mod2c"]
+	path = external/mod2c
+  url = https://github.com/BlueBrain/mod2c
+  branch = coreneuron_submodule
```
pramodskumbhar: to master

coreneuron/CMakeLists.txt
```diff
@@ -71,7 +62,7 @@ macro(mod2c_target name input)
     list(APPEND MOD2C_${name}_MODS "${mod2c_modname_}")
 
     add_custom_command(OUTPUT "${mod2c_output_}"
-      DEPENDS ${input} "${mod2c_install_prefix}/mod2c_core"
+      DEPENDS ${input} mod2c
```
pramodskumbhar: could you check if change in mod2c recompile mod files?

fouriaux: It dont recompile when there is changes to mod2c

PR #53, opened 2017-10-27
Minor RPath change

sharkovsky:
The Coreneuron library get installed in a location defined at line 78 of `CMake/config/DefineInstallationPaths.cmake`
```cmake
SET(LIB_INSTALL_DIR
    "${INTERNAL_BASE_PREFIX}/lib${LIB_SUFFIX}"
    CACHE PATH "The subdirectory relative to the install prefix where libraries will be in    stalled (default is prefix/lib)"
)
```
However, the correct path was not being picked up by the file `CMake/config/SetRpath.cmake`, because it assumed that the path always ended with `/lib`. 

I am fixing this by using the variable `LIB_INSTALL_DIR` defined in `CMake/config/DefineInstallationPaths.cmake`.

P.S. There is **no need to test this on BG/Q and on Cray Systems**, becase these changes are within a 
```cmake
IF( NOT BLUEGENE AND NOT CRAY_SYSTEM)
```
conditional block.


pramodk:
looks good to me but jenkins going funny! lets see tomorrow (once bg-q will be back).

PR #59, opened 2017-11-08
Refactor CoreNeuron / mod2c CMake chain

adevress:
- Remove ExternalProject needs, avoid the problems related to it
- Add a "DISABLE_BUNDLE" mode where external mod2c can be found and use
- Add mod2c as a `add_subdirectory`
- Properly propagate mod2c targets to the generators to avoid parallel build dependencies issues
- Validate the build with NinnnnnnJa (Tim's style) on both viz cluster and BGQ
- Follow the same pattern for frontend / backend compiler than before

Adev


fouriaux:
@adevress : seems great. If you have access to Theta, can you try to build on it, so that we ensure it is working on Cray machines ?

fouriaux:
@adevress : actually sorry, we can check that directly :-)

adevress:
@fouriaux  Thx Jeremy.

Can you type "add to whitelist" in a comment ? That would allow the stupid Jenkins bot to build my commits

pramodskumbhar:
@adevress : build is failing because of mod2c tests. Let me have a look.

fouriaux:
after discussion: we will do some refactoring in order to get cross-compilation on theta.

PR #62, opened 2017-11-15
Fix bug in gpu implementation (fixes #61)

### LINKED ISSUE #61: Bug in CoreNEURON GPU implementation
Creating separate issue reported by @pkuzyc in #60 : 

```
I tried to use the current commit of CoreNEURON to simulate this PC model, but there is a segmentation fault when I run it. The same error also happens when I run ringtest with this commit. The error message is :

--- Copying to Device! ---
-----------COPYING 0'th NrnThread TO DEVICE ---------------
GPU MEMORY USAGE (MB) : Used = 66.375000, Free = 11373.500000 MB, Total = 11439.875000zsh: segmentation fault (core dumped)  ./coreneuron_exec -e 100 -d  --gpu
is there anything wrong?
```

pramodskumbhar:
Avoid dereferencing gpu pointer on CPU (this was
introduced from changes in #14).

### DIFF
--- a/coreneuron/nrniv/nrn_acc_manager.cpp
+++ b/coreneuron/nrniv/nrn_acc_manager.cpp
@@ -118,7 +118,7 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
         dptr = d__data + 5 * ne;
         acc_memcpy_to_device(&(d_nt->_actual_area), &(dptr), sizeof(double*));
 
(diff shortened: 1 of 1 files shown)


pramodskumbhar:
I have added ticket #63. I actually we should execute all tests on GPU.

pramodk:
Manually ran all tests but BBP Jenkin queue has an issue on viz cluster. Will create an issue for Jenkins, merging this.

PR #65, opened 2017-11-29
fix pattrnStim when using more mpi ranks than cells

fouriaux:
fix segfault when setting more mpi ranks than nb cells and that a patternStim file is specified


pramodk:
#67 will handles all changes related to patternstim.

PR #67, opened 2017-12-02
Fix issues with spike replay / patternstim

pramodk:
- pattern.mod expects spike raster to be sorted (to avoid injecting
   large events at the begining of simulation itself). Refactor
   patternstim implementation to sort spike raster
 - if rank is empty then don't create patternstim (avoid segfault)
 - prcellstate file name time format changed simular to NEURON

### DIFF
--- a/coreneuron/nrniv/main1.cpp
+++ b/coreneuron/nrniv/main1.cpp
@@ -218,12 +218,12 @@ void call_prcellstate_for_prcellgid(int prcellgid, int compute_gpu, int is_init)
             if (is_init)
                 sprintf(prcellname, "%s_gpu_init", prprefix);
             else
-                sprintf(prcellname, "%s_gpu_t%g", prprefix, t);
+                sprintf(prcellname, "%s_gpu_t%f", prprefix, t);
         } else {
             if (is_init)
                 strcpy(prcellname, "cpu_init");
             else
-                sprintf(prcellname, "cpu_t%g", t);
+                sprintf(prcellname, "cpu_t%f", t);
         }
         update_nrnthreads_on_host(nrn_threads, nrn_nthread);
         prcellstate(prcellgid, prcellname);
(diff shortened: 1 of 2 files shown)


nrnhines:
This seems ok from the perspective of formally working. But it transiently uses twice the size
of the file to do the sort, and from a legacy perspective an already sorted file was a precondition.
PatternStim was originally envisioned for retrospective stimulation of one or a small number of
cells on a serial machine. I'm no longer sure what are the use cases in a parallel coreneuron environment.

pramodk:
test failure is due to bg-q queue unavailability during weekend (which we can ignore)

pramodk:
Hi Michael,

> But it transiently uses twice the size of the file to do the sort

that's true. I was considering to implement in-place sorting but then we still need to "store" permutation vector in order to sort the second vector. Hence thought few MBs for transient usage would be ok. (for introducing `namespace` in coreneuron (to build as library to link with neuron/neurodamus), we have to change all files to `.cpp` and thought we could refactor `struct Info` with C++ `std::vector<std::pair>`. And hence used above simple implementation for now.)

> PatternStim was originally envisioned for retrospective stimulation of one or a small number of
cells on a serial machine. I'm no longer sure what are the use cases in a parallel coreneuron environment.

Currently we are going through all our neuron+neurodamus simulations and validating them via CoreNEURON. Two main use cases for replay : 
- thalamocortical sims where presyn gids doesn't exist 
- Replay with diverse random number seeds

Neurodamus users are using BlueConfig to specify spike raster file. Neurodamus uses HOC interface to read+sort spike file and then trigger events. In CoreNEURON we thought we should use  patternstim to achieve the same functionality. 

>  from a legacy perspective an already sorted file was a precondition.

I was debugging patternstim and everything was working for small test example and were not able to reproduce the issue. But for production sims, users are using unsorted spike raster in neurodamus and hence thought would be better to sort it in CoreNEURON to avoid errors.

PR #68, opened 2017-12-03
Added support for euler (similar to thread safe support in NEURON)

pramodk:
- updated kinderiv.py to generate
    - added euler_thread from neuron and updated for aos/soa layout

@nrnhines : could you check if below changes belong to `dimplic.c` ? At least I tested the functionality/ results with plasticity model of Giuseppe and produces same results as NEURON+Neurodamus.

### DIFF
--- a/coreneuron/kinderiv.py
+++ b/coreneuron/kinderiv.py
@@ -1,6 +1,9 @@
 #!/usr/bin/python
 
 # read the translated mod files and construct _kinderiv.h
+# this is implemented to avoid function callbacks using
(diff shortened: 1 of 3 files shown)


pramodk:
@nrnhines : these are changes after our todays discussion. Tested with Giuseppe's model (GluSyn+Euler). (we should squash all commits while merging).

```
$  diff neurodamus_output/out.spk coreneuron_output/out.spk
1d0
< /scatter

$ wc -l coreneuron_output/out.spk
120184 coreneuron_output/out.spk
```

PR #69, opened 2017-12-05
fix segfault when getting more mpi ranks than nb cells 

fouriaux:
when a patternStim file is specified on command line, ranks without cells should do nothing concerning patternStim


pramodk:
duplicate of #67

PR #72, opened 2018-01-05
Implementation of synapse report in coreneuron

fouriaux:
Add command line option
  Add mapping between synapses and variable names to register
  Add a GID to compartment mapping functionality
make coreneuron compiling in C++11


nrnhines:
In general I like the improvements,  And in the future, the data_layuot.hpp can be used to improve
the clarity of the many places where an explicit 1 or 0 is used instead of AOS_LAYOUT, SOA_LAYOUT.
I'm perplexed by the removal of  nrniv_mf.h . How can that be?

fouriaux:
@michael, we have noticed that this file is not used anywhere in coreneuron, however, reviewing the direction we agreed to go (coreneuron as a library) this file might be useful to expose public API after refactoring. I can keep it as is for now on, also it is not strictly needed in coreneuron yet...

nrnhines:
You are correct. The substance of nrniv_mf.h has been moved into nrnoc/membfunc.h and with
better typedefs. I would remove it using a separate changeset as it has nothing to do with reporting.

pramodk:
Notes from discussion:
* validation of full mc2_column (31k cells)
* reports are binary identical with neuron+reportinglib?
* is selection of target from BlueConfig + neurodamus working?

fouriaux:
partial answer:
 -- it is not binary identical to neuron+reportinglib ( I am currently exploring why) however, brion tool dump_report.py shows identical outputs for neuron and coreneuron.

-- scaling the simulation provides some warnings, results dumps are not visible as brion segfault on coreneuron results.

PR #77, opened 2018-03-02
Bug fix: skip initial block for capac.c and eion.c (necessary for checkpoint-restore)

pramodk:


pramodk:
sure, updated comments

PR #78, opened 2018-03-14
Dryrun mechanisms

fouriaux:
for getting a better load balancing, we need more accurate measurement of mechanisms execution in coreneuron. Previous version was using estimation from Neuron, however we have experienced some different figures when directly estimating time in coreneuron. 
This version is executing each mechanism in coreneuron instead of relying on neuron estimation.

### DIFF
--- a/coreneuron/nrniv/dryrun.h
+++ b/coreneuron/nrniv/dryrun.h
@@ -1,0 +1,40 @@
+/*
+Copyright (c) 2018, Blue Brain Project
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


coreneuron/nrniv/dryrun.h
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ *  Dryrun mode of coreneuron, computing execution statistics and estimators 
+ *  without a full simulation execution
+ */
+
+#ifndef dryrun_h
+#define dryrun_h
+
+extern "C"
```
ohm314: why is this needed?

pramodk:
dry-run mode will be implemented later after planned refactoring of the code base.

PR #79, opened 2018-03-21
Centralize global variables definitions

fouriaux:
in perspective of making coreneuron as a library, we need to be careful of potential symbol conflicts with Neuron.
This first step clean up potentially conflictual declarations.

### DIFF
--- a/coreneuron/nrniv/nrnoptarg.cpp
+++ b/coreneuron/nrniv/nrnoptarg.cpp
@@ -372,7 +372,6 @@ static void graceful_exit(int err) {
 #if defined(nrnoptargtest)
 // for testing, compile with: g++ -g -I../.. nrnoptarg.cpp
 
-int nrnmpi_myid;
 
 int main(int argc, const char* argv[]) {
     nrnopt_parse(argc, argv);
--- a/coreneuron/nrniv/nrnoptarg.cpp
+++ b/coreneuron/nrniv/nrnoptarg.cpp
@@ -368,25 +368,3 @@ static void graceful_exit(int err) {
 #endif
     exit(nrnmpi_myid == 0 ? err : 0);
 }
-
-#if defined(nrnoptargtest)
-// for testing, compile with: g++ -g -I../.. nrnoptarg.cpp
-
-
-int main(int argc, const char* argv[]) {
-    nrnopt_parse(argc, argv);
-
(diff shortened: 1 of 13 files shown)


coreneuron/nrnoc/eion.c
```diff
@@ -67,7 +68,7 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #define _PRAGMA_FOR_SEC_ORDER_CUR_ACC_LOOP_ _Pragma("")
 #endif
 
-extern void hoc_register_prop_size(int, int, int);
+//extern void hoc_register_prop_size(int, int, int);
```
pramodk: remove this

coreneuron/nrnoc/fadvance_core.c
```diff
@@ -32,10 +32,10 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #include "coreneuron/nrnoc/nrnoc_decl.h"
 #include "coreneuron/nrniv/nrn_acc_manager.h"
 #include "coreneuron/coreneuron.h"
-
+//#include "coreneuron/utils/reports/nrnreport.h"
```
pramodk: remove

coreneuron/nrniv/nrnoptarg.cpp
```diff
@@ -372,7 +372,6 @@ static void graceful_exit(int err) {
 #if defined(nrnoptargtest)
 // for testing, compile with: g++ -g -I../.. nrnoptarg.cpp
 
-int nrnmpi_myid;
```
nrnhines: This must exist in the context of a test as mentioned in the comment above it. If it is removed one should remove
the entire test. If not removed I guess it could be helpful to add a comment as to why it is there.

fouriaux: @nrnhines: thanks Michael I overlooked this one. 
This is currently compiling correctly as I have moved the definition into nrnmpi.h. 
Nevertheless, I will move this file as a test case, which will clarify the intend of it.

pramodk: @nrnhines : Jeremy is working on namespace with c++ to make it as library. With c++, above definition result into duplicate definition error (but not with c). As we have tests for this in `tests/unit/cmdline_interface/test_cmdline_interface.cpp`, we can remove this from here.

PR #80, opened 2018-03-22
enforce c++ rules before migration

fouriaux:
* We enforce explicit casts: C is more tolerant than C++ on casts.
* We change variable names from C++ keywords into none conflicting names.
* explicit return type and arguments in declaration mandatory in C++

### DIFF
--- a/coreneuron/nrnoc/register_mech.c
+++ b/coreneuron/nrnoc/register_mech.c
@@ -214,11 +214,11 @@ int register_mech(const char** m,
 #if VECTORIZE
     memb_func[type].vectorized = vectorized ? 1 : 0;
     memb_func[type].thread_size_ = vectorized ? (vectorized - 1) : 0;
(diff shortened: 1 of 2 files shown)


coreneuron/nrnoc/register_mech.c
```diff
@@ -214,11 +214,11 @@ int register_mech(const char** m,
 #if VECTORIZE
     memb_func[type].vectorized = vectorized ? 1 : 0;
     memb_func[type].thread_size_ = vectorized ? (vectorized - 1) : 0;
-    memb_func[type].thread_mem_init_ = (void*)0;
-    memb_func[type].thread_cleanup_ = (void*)0;
-    memb_func[type].thread_table_check_ = (void*)0;
+    memb_func[type].thread_mem_init_ = NULL;
```
pramodk: after migration we will do nullptr.

PR #81, opened 2018-03-25
Trying to centralize includes and extern declaration visible to mod2c

fouriaux:
We centralize all declarations needed by mod2c in coreneuron headers

### DIFF
--- a/coreneuron/mech/mod2c_core_thread.h
+++ b/coreneuron/mech/mod2c_core_thread.h
@@ -57,6 +57,7 @@ extern int nrn_newton_steer(int, _threadargsproto_);
 extern int nrn_euler_steer(int, _threadargsproto_);
 #define eulerfun(arg) nrn_euler_steer(arg, _threadargs_);
 
+
 #endif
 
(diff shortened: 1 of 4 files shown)


coreneuron/mech/mod2c_core_thread.h
```diff
@@ -57,6 +57,7 @@ extern int nrn_newton_steer(int, _threadargsproto_);
 extern int nrn_euler_steer(int, _threadargsproto_);
 #define eulerfun(arg) nrn_euler_steer(arg, _threadargs_);
 
+
```
pramodk: remove extra line ?

coreneuron/scopmath_core/sparse_thread.c
```diff
@@ -1,6 +1,5 @@
 #include "coreneuron/mech/cfile/scoplib.h"
 #include "coreneuron/nrniv/nrn_assert.h"
-
```
pramodk: keep line here? and also same for register_mech.c?

PR #90, opened 2018-05-17
Use the value calculated mindelay by the MPI min reduction

ohm314:
Also remove superfluous mpi wrapper function

### DIFF
--- a/coreneuron/nrniv/netpar.cpp
+++ b/coreneuron/nrniv/netpar.cpp
@@ -801,15 +801,14 @@ double set_mindelay(double maxdelay) {
 
     // printf("%d netpar_mindelay local %g now calling nrnmpi_mindelay\n", nrnmpi_myid, mindelay);
     //	double st = time();
-    mindelay_ = nrnmpi_mindelay(mindelay);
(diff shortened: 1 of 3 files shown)


nrnhines:
https://www.neuron.yale.edu/neuron/static/new_doc/modelspec/programmatic/network/parcon.html#ParallelContext.set_maxstep

ohm314:
yes, i think this can be merged afaik

coreneuron/nrniv/netpar.cpp
```diff
@@ -801,15 +801,14 @@ double set_mindelay(double maxdelay) {
...
+    mindelay_ = nrnmpi_dbl_allmin(mindelay);
     //	add_wait_time(st);
     // printf("%d local min=%g  global min=%g\n", nrnmpi_myid, mindelay, mindelay_);
     errno = 0;
-    return mindelay;
 #else
     mindelay_ = mindelay;
-    return mindelay;
 #endif  // NRNMPI
+    return mindelay_;
```
pramodk: as commented previously, neurones implementation return "local mindelay". In coreneuron context, its ok/make sense to return global mindelay (as there is no scripting interface).

PR #96, opened 2018-06-05
clang-format after moving to c++

pramodk:
- previous PR from Jeremy didn't update source style
 - disable indentation of namespaces
 - upgrade clang-format  to 5.0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,8 +163,8 @@ add_definitions(-DCORENEURON_BUILD)
 option (ENABLE_SELECTIVE_GPU_PROFILING "Enable GPU profiling only for Solver" ON)
 option (ENABLE_OPENACC "Enable use of OpenACC" OFF)
 
-set(CLANG_FORMAT_MIN_VERSION "5.0")
-set(CLANG_FORMAT_MAX_VERSION "5.9")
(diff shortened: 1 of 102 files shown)


CMakeLists.txt
```diff
@@ -163,8 +163,8 @@ add_definitions(-DCORENEURON_BUILD)
 option (ENABLE_SELECTIVE_GPU_PROFILING "Enable GPU profiling only for Solver" ON)
 option (ENABLE_OPENACC "Enable use of OpenACC" OFF)
 
-set(CLANG_FORMAT_MIN_VERSION "4.0")
-set(CLANG_FORMAT_MAX_VERSION "4.9")
+set(CLANG_FORMAT_MIN_VERSION "5.0")
```
fouriaux: Does changing CLANG_FORMAT_MIN version is changing some formating or correcting C++ formating ?

pramodk: obviously this is typo...I wanted to set min to 5.0 and max to 5.9.

pramodk:
> Why are we changing clang minimal version ?

no specific reason but 4.0 is quite old now and when we install clang-format (say with brew), clang-format 5 is installed. Hence I thought we should start using latest version.

pramodk:
p.s. note that most of changes in this PR are not specific to 5.0

PR #97, opened 2018-06-08
Reportinglib integration tests

fouriaux:
Setting up integration tests for reportinglib


tests/integration/ring/globals.dat
```diff
@@ -22,6 +22,320 @@ ko0_k_ion 2.5
 ib_IntFire4 0
 eps_IntFire4 9.9999999999999995475e-07
 taueps_IntFire4 0.010000000000000000208
+Dt_BinReportHelper 0.10000000000000000555
+cai0_ca_ion 5.0000000000000002396e-05
+cao0_ca_ion 2
```
pramodk: Jeremy: could you generate dataset from non-neurodamus binary? 
All this globals are not related to ringtest. It would be clean to use mod files from ringtest only.

PR #99, opened 2018-06-13
allow one to check if mech has been instantiated by checking alloc function

brunomaga:


coreneuron/nrnoc/register_mech.cpp
```diff
@@ -159,6 +159,7 @@ void alloc_mech(int n) {
         int i;
         for (i = 0; i < memb_func_size_; ++i) {
             nrn_mech_data_layout_[i] = 1;
+            memb_func[i].alloc = NULL; //mark as non-instantiated
```
pramodk: `members_func` is allocated with `calloc` so it's already initialised to 0 (unless I am missing something here...)

brunomaga: @pramodk this commit fixes verification of the instantiation for non-`calloc` allocation of `memb_func`

pramodk: could you check why build was failed? might need close and reopen.

pramodk:
`alloc_mech` allocates with calloc so I don't see how this is relevant here now.

PR #104, opened 2018-07-02
inlining euler_thread function

fouriaux:


coreneuron/mech/mod2c_core_thread.h
```diff
@@ -112,8 +112,23 @@ extern void _nrn_destroy_sparseobj_thread(SparseObj* so);
 extern int nrn_kinetic_steer(int, SparseObj*, double*, _threadargsproto_);
 #define spfun(arg1, arg2, arg3) nrn_kinetic_steer(arg1, arg2, arg3, _threadargs_);
 
+// derived from nrn/src/scopmath/euler.c
+// updated for aos/soa layout index
 #pragma acc routine seq
-extern int euler_thread(int, int*, int*, EULFUN, _threadargsproto_);
+inline int euler_thread(int neqn, int* var, int* der, DIFUN fun, _threadargsproto_) {
```
pramodk: this should be static inline to avoid possible duplication error at link time.

PR #107, opened 2018-07-30
Minor changes to compile with OpenACC and latest cuda + pgi compiler:

pramodk:
- acc routine seq in header causes pgi internal errors
    - sm_20 is outdated, keep minimum sm_35


pramodk:
@fouriaux  : yes, you are right. There are few cuda files in coreneuron that don't do large compute:

- starting stopping profiling
- pick random123 stream

Hence for arch compatibility / avoid failure, minimum sm version was used. 

Now I used sm_30 and sm_60. If sm_60 is not supported, it will fallback to older version to avoid failure. Is that reasonable?

PR #121, opened 2018-10-08
Patternstim fix : allow restore phases to provide new pattern-stim

pramodk:
This allows specifying separate patternstim file for restore part:
  - start event chain by sending event at t+0
  - discard events from file that arrive before t

Here is internal ticket: https://bbpteam.epfl.ch/project/issues/browse/CNEUR-313

```
Current implementation of coreneuron expects whole spike-replay file provided on command line. We have fixed this by https://github.com/BlueBrain/CoreNeuron/pull/121.
But this might result in loss of spikes in case of:
  * user specify tstop = 100 for part0
  * simulator stop at 100.1
  * pattern.dat in part1 has event at 100.05
One way to avoid this error is to always provide whole pattern.dat file to coreneuron. Is that feasible solution? (note that the implementation in coreneuron needs to be updated to avoid storing whole file)
```

@nrnhines : do we need to consider above corner case?

### DIFF
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -265,7 +265,7 @@ void nrn_init_and_load_data(int argc,
 
     // Invoke PatternStim
     if (!corenrn_param.patternstim.empty()) {
-        nrn_mkPatternStim(corenrn_param.patternstim.c_str());
+        nrn_mkPatternStim(corenrn_param.patternstim.c_str(), corenrn_param.tstop);
     }
 
     /// Setting the timeout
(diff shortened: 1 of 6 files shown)


pramodk:
@nrnhines : could you review this?

coreneuron/mech/modfile/pattern.mod
```diff
@@ -116,6 +116,7 @@ void pattern_stim_setup_helper(int size, double* tv, int* gv, _threadargsproto_)
 	info->size = size;
 	info->tvec = tv;
 	info->gidvec = gv;
+	artcell_net_send ( _tqitem, -1, (Point_process*) _nt->_vdata[_ppvar[1*_STRIDE]], t +  0.0 , 1.0 ) ;
```
nrnhines: A comment would be helpful. Is this needed for initialization after restore?

pramodk: I am testing it now.

pramodk:
This should be merged after having test in CI

coreneuron/mech/modfile/pattern.mod
pramodk: Done

PR #133, opened 2019-02-18
Fixing nrnivmodl-core: Avoid redefined symbols + script startup fixes

ferdonline:
**Fixes to nrnivmodl-core and required changes in the build of coreneuron**

Current implementation of nrnivmodl-core suffers from a complex problem, redefinition of some symbols which depend on the mechanism set: `modl_reg()` and those defined in `dimplic.cpp`

This PR fixes that problem (and other little ones) from the root which required changes in the organization of coreneuron libraries and executables. 

**Short summary:**

- libcoreneuron shall be a "kernel" lib, i.e. it shall not contain extra mechanisms and their dependencies, but "complete", i.e.: It contains the standard mechanisms and everything (no undefined symbols) to be able to run on standard simulations.
  - `modl_reg()` call was accordingly moved out `main1.cpp`
  - `scopmath_core` package excluded from the `libcoreneuron` too
- libcoreneuron's `solve_core` was split into `mk_mech_init` and `run_solve_core`
- A new (coreneuron+mechanisms) lib exists (`corenrnmech`) for the previous purpose, which includes
  - mechanisms per se, `modl_reg()`. `solve_core()` (which calls `modl_reg()`) and full scopmath_core (+ dimplic.cpp)
  - links against `libcoreneuron`
- coreneuron main therefore only has to do a bare `solve_core()` from corenrnmech lib (importing the new `enginemech.h` header)

When building mods with `nrnivmodl-core` the process is the same! 
Note that in order to optionally build-up on previous mechanism, the ADDITIONAL mechamisms are exported as cpp files to `share/mod2c` and bundled together with the new mods cpp's.
See the diagram below.


ferdonline:
Updated nrnivmodl-core diagram:

ferdonline:
@pramodk I don't think we expected this much work :p but it was a nice challenge!
We might need to adapt some software though in case they link to the previous `libcoreneuron` and expect to have the extra machanisms included

extra/CMakeLists.txt
```diff
@@ -33,6 +33,16 @@ set(BUILD_TYPE_C_FLAGS "${CMAKE_C_FLAGS_${_BUILD_TYPE}}")
 set(BUILD_TYPE_CXX_FLAGS "${CMAKE_CXX_FLAGS_${_BUILD_TYPE}}")
 message("CXX Compile Flags from BUILD_TYPE: ${BUILD_TYPE_CXX_FLAGS}")
 
+############################################################
+# Disable new-dtags with newer linkers
+############################################################
+
+if (UNIX AND NOT APPLE)
```
ferdonline: If you want to avoid this workaround I think we could live with RUNPATH by setting it also in the shared libraries (apparently it allows that). Shall I experiment with that?

pramodk:
> @pramodk I don't think we expected this much work :p but it was a nice challenge!

yeah....too much headache though! :)

extra/CMakeLists.txt
pramodk: without that --disable-dtags option, the RUNPATH is populated correctly but still get library not found error. Didnt understand that quite well.

ferdonline: It might be related to the fact that they are used a bit differently by the dynamic loader: See [Bug #1253638 “dynamic linker does not use DT_RUNPATH for transitive dependencies](https://bugs.launchpad.net/ubuntu/+source/eglibc/+bug/1253638)
If I understood well, RUNPATH is only valid for direct dependencies, but sub-dependencies can also define RUNPATH which gets also checked. I kind of agree with the new behavior, but true, breaks things

ferdonline:
@pramod I linked using gold, checked it produces RUNPATH, but still it all works in bb5. Is it something specific to ubuntu?
I noticed that libcoreneuron is directly linked so it should indeed work.
To make the test, in the last commit I added rpath also to the lib (it seems RUNPATH supports that in libs) and removed your patch. Check if that also works with you, otherwise please discard that commit.

coreneuron/CMakeLists.txt
```diff
@@ -315,8 +312,8 @@ add_library(coreneuron ${COMPILE_LIBRARY_TYPE} ${coreneuron_all_headers} ${coren
                        ${coreneuron_all_c_files} ${cudacorenrn_objs} ${MOD2C_STDMECH_OUTPUTS})
 add_dependencies(coreneuron kinderiv)
 
-target_link_libraries(coreneuron ${MPI_CXX_LIBRARIES} 
-                      ${link_reportinglib} ${CUDA_LIBRARIES})
+target_link_libraries(coreneuron ${MPI_CXX_LIBRARIES}
```
ferdonline: ${cudacorenrn_objs} doesnt exist now, right?

coreneuron/CMakeLists.txt
```diff
@@ -297,8 +300,20 @@ if(ENABLE_REPORTINGLIB)
...
+
+############################################################
+# Target libs
+############################################################
+
+#
+# The coreneuron lib (only internal mechanisms)
+#
+add_library(coreneuron ${COMPILE_LIBRARY_TYPE} ${coreneuron_all_headers} ${coreneuron_all_templates}
+                       ${coreneuron_all_c_files} ${cudacorenrn_objs} ${MOD2C_STDMECH_OUTPUTS})
```
ferdonline:
```suggestion
                       ${coreneuron_all_c_files} ${MOD2C_STDMECH_OUTPUTS})
```

coreneuron/utils/reports/nrnreport.h
```diff
@@ -68,7 +68,7 @@ struct ReportConfiguration {
     double start;                           // start time of report
     double stop;                            // stop time of report
     int num_gids;                           // total number of gids
-    size_t buffer_size;                     // hint on buffer size used for this report
+    int buffer_size;                     // hint on buffer size used for this report
```
ferdonline:
```suggestion
    int buffer_size;                        // hint on buffer size used for this report
```

PR #134, opened 2019-02-19
Fixing large stdout issue

### LINKED ISSUE #116: Avoid large stdout from long running simulations due to progressbar
Currently progress bar result into large stdout. Change the implementation similar to python-neurodamus.

iomaganaris:
- Added an interval between progress bar prints
- Max number of progress bar prints set to 1000
- Keeps ETA updated regularly
- Fixes #116

### DIFF
--- a/coreneuron/utils/progressbar/progressbar.c
+++ b/coreneuron/utils/progressbar/progressbar.c
@@ -37,18 +37,8 @@ enum {
     /// The maximum number of bar redraws (to avoid frequent output in long runs)
     BAR_DRAW_COUNT_MAX = 500,
 
-    /// Reference bar redraw interval (seconds)
-    BAR_DRAW_INTERVAL = 5,
-
-    /// The Initial period, when BAR_DRAW_INTERVAL is always respected (seconds)
-    BAR_DRAW_INIT_PERIOD = 120,
-
-    /// Number of progress bar redraws after the Initial period to draw maximum of
-    /// BAR_DRAW_COUNT_MAX in total
-    BAR_DRAW_COUNT_REMAINING =
-        BAR_DRAW_COUNT_MAX -
-        BAR_DRAW_INIT_PERIOD / BAR_DRAW_INTERVAL  // assumes that the time that 4-5 simulation steps
-                                                  // take is less than BAR_DRAW_INTERVAL
+    /// Number of ETA samples taken to calculate the time intervals between progress bar redraws
+    NUMBER_OF_ETA_SAMPLES = 1000
 };
 
 /// Models a duration of time broken into hour/minute/second components. The number of seconds
@@ -77,12 +67,12 @@ progressbar* progressbar_new_with_format(const char* label, unsigned long max, c
 
     new->max = max;
     new->value = 0;
-    new->prev_value = 0;
-    new->value_init_period_last = 0;
+    new->draw_time_interval = 2;
+    new->prev_sample_value = 0;
+    new->drawn_count = 0;
     new->t = 0;
     new->start = time(NULL);
     new->prev_t = time(NULL);
-    new->eta_init_period_last = 0;
     assert(3 == strlen(format) && "format must be 3 characters in length");
     new->format.begin = format[0];
     new->format.fill = format[1];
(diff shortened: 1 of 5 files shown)


coreneuron/utils/progressbar/progressbar.c
```diff
@@ -94,7 +98,13 @@ void progressbar_free(progressbar* bar) {
 void progressbar_update(progressbar* bar, unsigned long value, double t) {
     bar->value = value;
     bar->t = t;
-    progressbar_draw(bar);
+    // The progress bar will be printed only if the minimum interval
+    // to print max BAR_UPDATES_NUMBER bars has passed from the
+    // previous progress bar printing.
+    if ((bar->value - bar->prev_value) >= bar->max / BAR_UPDATES_NUMBER) {
```
ferdonline: `max->max` and `BAR_UPDATES_NUMBER` are both integral types, division will be 0 if `BAR_UPDATES_NUMBER > bar->max`, so it's gonna always print.
Some casts are needed

coreneuron/utils/progressbar/progressbar.c
```diff
@@ -32,6 +32,9 @@ enum { WHITESPACE_LENGTH = 2 };
 /// The amount of width taken up by the border of the bar component.
 enum { BAR_BORDER_WIDTH = 2 };
 
+/// Progress Bar max number of updates
+enum { BAR_UPDATES_NUMBER = 1000 };
```
ferdonline: I find this number too large to be a constant. (I don't know if you already agreed with someone else), but 1000 prints is a lot of output.
To my mind we could instead print after N seconds have elapsed

coreneuron/utils/progressbar/progressbar.c
iomaganaris: You are right, however if bar->max, which is the number of simulation steps, is smaller that the BAR_UPDATE_NUMBER there are still going to be less than BAR_UPDATE_NUMBER of bars printed, which is the intended behavior.

coreneuron/utils/progressbar/progressbar.c
iomaganaris: This particular number was chosen after a small discussion with @pramodk .
I see your point and I will try to improve this part as you suggested.
Thank you very much! :+1:

ferdonline: Another idea. Updates in the first minutes is important so that the user gets to know if something is wrong. 
If the simulation lasts for three days, even with BAR_UPDATES_MAX=1000 he only knows anything after 4.5min. 
@pramodk I don't know if this is overcomplicating, how cool we would like this. Maybe the first X minutes we force updates at rate BAR_UPDATE_INTERVAL, and only after that, we might reduce the rate based on BAR_UPDATES_MAX?

iomaganaris: I have made a change in the print mechanism based on @ferdonline comments.
Please see the next commit.

coreneuron/utils/progressbar/progressbar.c
```diff
@@ -32,6 +32,18 @@ enum { WHITESPACE_LENGTH = 2 };
...
+enum { BAR_UPDATES_NUMBER = 500 };
+
+/// Progress Bar interval of updates (seconds)
+enum { BAR_UPDATES_INTERVAL = 5 };
+
+/// Progress Bar time of change of updates (seconds)
+enum { BAR_UPDATES_CHANGE = 120 };
+
+/// Progress Bar number of slow updates
+enum { BAR_UPDATES_SLOW = BAR_UPDATES_NUMBER - BAR_UPDATES_INTERVAL * BAR_UPDATES_CHANGE };
```
ferdonline: I'm not getting the logic, sorry, maybe tomorrow you can explain me. But this is resulting in a constant -100.

coreneuron/utils/progressbar/progressbar.c
```diff
@@ -32,6 +32,18 @@ enum { WHITESPACE_LENGTH = 2 };
...
 enum { BAR_BORDER_WIDTH = 2 };
 
+/// Progress Bar max number of updates
+enum { BAR_UPDATES_NUMBER = 500 };
+
+/// Progress Bar interval of updates (seconds)
+enum { BAR_UPDATES_INTERVAL = 5 };
+
+/// Progress Bar time of change of updates (seconds)
+enum { BAR_UPDATES_CHANGE = 120 };
```
ferdonline: more suggestive names: 
```
/// The reference bar redraw interval
BAR_DRAW_INTERVAL = 5    // maybe (isatty()? 1 : 5)

/// The maximum number of updates (avoid frequent output on very long runs ) 
BAR_DRAW_COUNT_MAX = 1000

/// The Initial period, where BAR_DRAW_INTERVAL is always respected
BAR_DRAW_FREQUENT_PERIOD = 120

File: coreneuron/utils/progressbar/progressbar.c [hunk 5]
```diff
@@ -190,7 +214,7 @@ static void progressbar_draw(const progressbar* bar) {
     // Draw the ETA
     fputc(' ', stdout);
     fprintf(stdout, ETA_FORMAT, bar->t, eta.hours, eta.minutes, eta.seconds);
-    fputc('\r', stdout);
+    fputc('\n', stdout);
```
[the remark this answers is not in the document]
iomaganaris: I was doing some checks using the new line char and forgot to change it again. It should be '\r' normally.

coreneuron/utils/progressbar/progressbar.c
iomaganaris: Thank you very much. I will take care of this and try to be more descriptive the next time 👍

iomaganaris: @ferdonline Please check the next commit and give me some feedback if you want on the variables' names descriptiveness.

coreneuron/utils/progressbar/progressbar.c
```diff
@@ -118,35 +108,36 @@ void progressbar_free(progressbar* bar) {
...
-        bar->eta_init_period_last / BAR_DRAW_INTERVAL <= BAR_DRAW_COUNT_REMAINING;
-    int is_more_than_time_interval = difftime(cur_time, bar->prev_t) >= BAR_DRAW_INTERVAL;
-    int is_more_than_steps_interval =
-        (bar->value - bar->prev_value) >=
-        (bar->max - bar->value_init_period_last) / BAR_DRAW_COUNT_REMAINING;
+    if (is_more_than_sample_interval) {
+        bar->prev_sample_value = bar->value;
+        bar->draw_time_interval = (eta_s) / (BAR_DRAW_COUNT_MAX - bar->drawn_count);
+        if (bar->draw_time_interval < 1)
+            bar->draw_time_interval = 1;  // avoid output in every time step
```
ferdonline: Better to always use `{ }` even in single line statements. (easy to add a second line and screw the logic :p )

iomaganaris: Thank you very much for the extensive and detailed comments @ferdonline . I really appreciate them, so that I can learn as much as I can. 
Just some further development on your thoughts:
> * 1000 samples seems a lot for me. Maybe ask one of our mathematicians, but I have the impression that to get a good average from a distribution we need actually very few samples (kind, less than 50...).

It's true that a large number of samples is not needed to get a good understanding of the ETA. However, I kept the sampling number high, so that in experiments with a large number of timesteps the first sampling is done quickly to avoid a large number of bars being redrawn(every 2 seconds) during this time. 
> * If the simulation gets heavier and heavier with time, even a lot of samples won't represent the mean. (Maybe Pramod knows if that's a common scenario or not). In case it is, a moving average would be better (of course, this is if we really want to be fancy ;) )
> * If we can stay on the simple side, maybe we could have `sample_interval = draw_time_interval` and simplify those if's.
You might be right in that. I want to get some more feedback on the different kinds of simulation behaviors to check if the sampling is okay or there are any problems with than in any case.

Thank you again and I will also correct the single line if-statement brackets. Good observation :+1:

ferdonline: another advantage of `sample_interval = draw_time_interval` is that we compute less ETA samples for small simulations.
For the "running average" consider simply:
- `projected_ETA = ETA_according_to_last_interval * 0.5 + previous_projected_ETA * 0.5`

If it works well it's what I call elegant: little code, great value! :)
What do you think?

iomaganaris: Thanks for your input again Fernando.
`sample_interval = draw_time_interval` is a very good choice indeed I think and I will add it after testing it.
Regarding the moving average, after some consideration I have found it may not be that necessary but please correct me if I am wrong as I may not have understood correctly the idea. ETA is the `simulation_time / number_of_timesteps_passed * number_of_time_steps_remaining`, so it is already based on the average time a simulation step takes and every time it is calculated it is already making a "better" forecast, as it is based in more timesteps. According to my understanding the moving average would have a greater impact in the resulting ETA if there was a huge difference in the average time needed to calculate a simulation timestep before and after the previous ETA calculation.  If I had understood correctly the way you have written your formula, I think that it should be `projected_ETA = (ETA_according_to_last_interval / current_simulation_step * 0.5 + previously_projected_ETA / previous_ETA_calc_simulation_step * 0.5) * simulation_steps_remaining` as ETA is time needed to finish the simulation, which is something different in every timestep.
Sorry for the long text and I hope I made my thoughts clear. Please tell me if I made any mistake in my train of thought or you want to develop further any point, as I see that the progress bar drawing has captivated the interest of both of us :P

ferdonline: Very clear, thanks! I see you introduced the elapsed time component in the weighted average, which indeed was required to be correct! But maybe with `ETA_according_to_last_interval` you meant the "time of the last sample", no? But yeah, that's about it. 
If it's a matter of changing the formula and you don't need to store more vars I'd give it a shot, otherwise keep the simple implementation.

iomaganaris: @ferdonline I would like to thank you very much for all your feedback! :+1: 
After discussing about the moving average with Pramod we reached a conclusion that the moving average is too much, as the time needed to calculate a single timestep is not changing a lot throughout the simulation, so it is not needed to be implemented for now.

PR #135, opened 2019-02-21
Removed warning for absence of default config file with name "config"

### LINKED ISSUE #93: --read-config option message
When using --read-arg with a non-default name for the config file, an information message appears which seems unclear about what information it provides.
```
>coreneuron_exec --read-config 1.conf
Info: No default arguments file provided: config
```

iomaganaris:
- No default configuration file needed with the exact name "config"
- Fixes #93

### DIFF
--- a/coreneuron/nrniv/nrnoptarg.cpp
+++ b/coreneuron/nrniv/nrnoptarg.cpp
@@ -160,16 +160,6 @@ int nrnopt_parse(int argc, const char* argv[]) {
         }
     }
 
-    // last the default config file
-    std::string cfg("config");
-    if (!opt->importFile(cfg.c_str(), '#')) {
-        if (nrnmpi_myid == 0)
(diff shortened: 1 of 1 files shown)


coreneuron/nrniv/nrnoptarg.cpp
```diff
@@ -160,16 +160,6 @@ int nrnopt_parse(int argc, const char* argv[]) {
...
-    // last the default config file
-    std::string cfg("config");
-    if (!opt->importFile(cfg.c_str(), '#')) {
-        if (nrnmpi_myid == 0)
-            std::cerr << "Info: No default arguments file provided: " << cfg << std::endl;
-    } else {
-        if (nrnmpi_myid == 0)
-            std::cout << "Read default configuration file: " << cfg << std::endl;
-    }
-
```
pramodk: @nrnhines  : we are only using option via `-read-config`. I think we can get read of this default `config` file?

PR #136, opened 2019-02-21
Fix CMake issue with PGI and README formatting.

st4rl3ss:
Minor fix of CMake file for PGI compilation. Also minor changes to readme due to formatting and inconsistencies.


CMake/config/CompilerFlagsHelpers.cmake
```diff
@@ -141,6 +141,9 @@ foreach(COMPILER_LANGUAGE ${SUPPORTED_COMPILER_LANGUAGE_LIST})
 			## Intel O3 is extreme
 			set(CMAKE_${COMPILER_LANGUAGE}_OPT_AGGRESSIVE "-O2")
 		endif()
+		if (CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID STREQUAL "PGI")
+			set(CMAKE_${COMPILER_LANGUAGE}_WARNING_ALL "")
+		endif()
```
pramodk: After looking at this I realise that `-Wall` shouldn't be set in this block...set empty this on line 126 and remove this block.

st4rl3ss: Fixed.

CMake/config/CompilerFlagsHelpers.cmake
```diff
@@ -123,7 +123,7 @@ foreach(COMPILER_LANGUAGE ${SUPPORTED_COMPILER_LANGUAGE_LIST})
 
 	## rest of the world
 	else()
-		set(CMAKE_${COMPILER_LANGUAGE}_WARNING_ALL "-Wall")
+		set(CMAKE_${COMPILER_LANGUAGE}_WARNING_ALL "")
```
tristan0x: You should create a section dedicated to PGI compiler where you remove `-Wall` instead of removing the flag in the default case.
This can help: https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html

st4rl3ss: That was my original approach. If you read above it's the opposite of what @pramodk told me to do. Which approach should I follow?

tristan0x: I can't see the first patch because branch has been forced pushed.
But I would create a `elseif(compiler_is_pig)` section line 124 before the final `else`.

CMake/config/CompilerFlagsHelpers.cmake
st4rl3ss: Regarding the "Endian" error, If I comment out the line that causes issues the tests are correctly compiled. The test then fail but I'm not sure it's connected to that. Am I the only one that is having that kind of errors while compiling tests?

st4rl3ss: Endian error fixed, created separate pull request.

CMake/config/CompilerFlagsHelpers.cmake
pramodk: > I can't see the first patch because branch has been forced pushed.
I didnt understand how GitHub allowed my local changes to get push forced....I think I fixed settings yesterday.

@tristan0x  is right.. I was thinking about cray and other compilers where -Wall may be right...but as Tristan suggested, you can unset Wall for PGI for now. See few lines below where we have done same for Intel.

st4rl3ss: Pushed new version with separate PGI block after Tristan and Pramod comments.

PR #139, opened 2019-02-24
Update cmake options 

pramodk:
- to build with Caliper framework
    - MOD2C command line option


pramodk:
Will re-submit with Caliper based instrumentation.

PR #140, opened 2019-02-26
Run simulation with different mod files if they are not used

iomaganaris:
Fixes issue with simulation not running if a mod file is different between Neuron and CoreNeuron even if it is not used in the simulation.

### DIFF
--- a/coreneuron/nrniv/nrn_setup.cpp
+++ b/coreneuron/nrniv/nrn_setup.cpp
@@ -263,6 +263,9 @@ std::vector<NetCon*> netcon_in_presyn_order_;
 /// Only for setup vector of netcon source gids
 std::vector<int*> netcon_srcgid;
 
+/// Vector storing indexes (IDs) of different mod files between Neuron and CoreNeuron
+extern std::vector<int> different_mod_files;
+
 // Wrap read_phase1 and read_phase2 calls to allow using  nrn_multithread_job.
 // Args marshaled by store_phase_args are used by phase1_wrapper
 // and phase2_wrapper.
@@ -1140,6 +1143,11 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
         for (int i = 0; i < nmech; ++i) {
             tml_index[i] = F.read_int();
             ml_nodecount[i] = F.read_int();
+            auto found_different_mod_file =
+                std::find(different_mod_files.begin(), different_mod_files.end(), tml_index[i]);
+            if (found_different_mod_file != different_mod_files.end()) {
+                exit(1);
+            }
         }
         nt._nidata = F.read_int();
         nt._nvdata = F.read_int();
--- a/coreneuron/nrniv/nrn_setup.cpp
+++ b/coreneuron/nrniv/nrn_setup.cpp
@@ -263,8 +263,10 @@ std::vector<NetCon*> netcon_in_presyn_order_;
 /// Only for setup vector of netcon source gids
 std::vector<int*> netcon_srcgid;
 
-/// Vector storing indexes (IDs) of different mod files between Neuron and CoreNeuron
-extern std::vector<int> different_mod_files;
+/// Vector storing indexes (IDs) of different mechanisms of mod files between Neuron and CoreNeuron
+extern std::vector<int> different_mechanism_type;
+
+static void graceful_exit(int);
 
 // Wrap read_phase1 and read_phase2 calls to allow using  nrn_multithread_job.
 // Args marshaled by store_phase_args are used by phase1_wrapper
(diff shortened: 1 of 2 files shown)


coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1137,9 +1140,20 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
...
             ml_nodecount[i] = F.read_int();
+            auto found_different_mod_file =
+                std::find(different_mod_files.begin(), different_mod_files.end(), tml_index[i]);
+            if (found_different_mod_file != different_mod_files.end()) {
+                printf("Error: %s is different version of MOD file than the one used by NEURON!\n",
+                       nrn_get_mechname(tml_index[i]));
+                diff_modfiles_count++;
+            }
+        }
+        if (diff_modfiles_count > 0) {
```
pramodk: here you can print message from rank 0 : `Error : NEURON and CoreNEURON must use same mod files for compatibility, x different mod files found. Re-compile special and special-core!`

coreneuron/nrnoc/register_mech.cpp
```diff
@@ -86,6 +87,9 @@ static int ion_write_depend_size_;
 static int** ion_write_depend_;
 static void ion_write_depend(int type, int etype);
 
+/* Vector keeping the types (IDs) of different mod files between Neuron and CoreNeuron */
+std::vector<int> different_mod_files;
```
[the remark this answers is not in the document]
iomaganaris: I agree and I am going to change it. I wasn't sure about this in the first place.

coreneuron/nrnoc/register_mech.cpp
```diff
@@ -277,9 +281,7 @@ void hoc_register_prop_size(int type, int psize, int dpsize) {
     if (psize != pold || dpsize != dpold) {
         printf("%s prop sizes differ psize %d %d   dpsize %d %d\n", memb_func[type].sym, psize,
```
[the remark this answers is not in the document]
iomaganaris: I will remove it since it is not needed. I thought that maybe the number of variables would be useful for debugging in a case of a real error.

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1137,9 +1140,20 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
...
         tml_index = new int[nmech];
         ml_nodecount = new int[nmech];
+        int diff_modfiles_count = 0;
         for (int i = 0; i < nmech; ++i) {
             tml_index[i] = F.read_int();
             ml_nodecount[i] = F.read_int();
+            auto found_different_mod_file =
+                std::find(different_mod_files.begin(), different_mod_files.end(), tml_index[i]);
+            if (found_different_mod_file != different_mod_files.end()) {
+                printf("Error: %s is different version of MOD file than the one used by NEURON!\n",
```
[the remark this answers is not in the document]
iomaganaris: Done :+1:

coreneuron/nrniv/nrn_setup.cpp
iomaganaris: Done :+1:

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1137,9 +1140,20 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
         nmech = F.read_int();
         tml_index = new int[nmech];
         ml_nodecount = new int[nmech];
+        int diff_modfiles_count = 0;
         for (int i = 0; i < nmech; ++i) {
             tml_index[i] = F.read_int();
             ml_nodecount[i] = F.read_int();
+            auto found_different_mod_file =
+                std::find(different_mod_files.begin(), different_mod_files.end(), tml_index[i]);
+            if (found_different_mod_file != different_mod_files.end()) {
```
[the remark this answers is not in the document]
iomaganaris: I have added the `std::find() != different_mechanism_type.end()` expression inside the if and I hope it is still understandable.
I have also got a question regarding this check only by rank 0. If `read_phase2()` is called by all threads, shouldn't all the ranks calculate the number of different files, so that they can all call `MPI_Finalize()` if there is a different mechanism and then `exit()`? 
Please correct me if I am wrong, but from a bit of study I found out that this is the case and added a `graceful_exit()` function that all ranks call if they need to exit. 
I have corrected the other parts of the code and only rank 0 prints to stdout.

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -263,6 +263,11 @@ std::vector<NetCon*> netcon_in_presyn_order_;
 /// Only for setup vector of netcon source gids
 std::vector<int*> netcon_srcgid;
 
+/// Vector storing indexes (IDs) of different mechanisms of mod files between Neuron and CoreNeuron
+extern std::vector<int> different_mechanism_type;
+
+static void graceful_exit(int);
```
pramodk: not needed anymore?

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1137,10 +1142,26 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
...
         for (int i = 0; i < nmech; ++i) {
             tml_index[i] = F.read_int();
             ml_nodecount[i] = F.read_int();
+            if (nrnmpi_myid == 0 &&
+                std::find(different_mechanism_type.begin(), different_mechanism_type.end(),
+                          tml_index[i]) != different_mechanism_type.end()) {
+                printf("Error: %s is a different MOD file than used by NEURON!\n",
+                       nrn_get_mechname(tml_index[i]));
+                diff_modfiles_count++;
+            }
```
pramodk: you were right, this if block doesn't look good to read...

you can compute error count from all rank but just do printf error conditional

coreneuron/nrniv/nrn_setup.cpp
```diff
@@ -1137,10 +1142,26 @@ void read_phase2(FileHandler& F, int imult, NrnThread& nt) {
...
         }
+
+        if (nrnmpi_myid == 0 && diff_modfiles_count > 0) {
+            printf(
+                "Error : NEURON and CoreNEURON must use same mod files for compatibility, %d different mod file(s) found. Re-compile special and special-core!\n",
```
pramodk: printf error from rank 0 only but everyone can all abort

coreneuron/nrniv/nrn_setup.cpp
iomaganaris: Okay. I will change all the if statements so that all ranks make the check but only rank 0 prints the messages. :+1:

PR #144, opened 2019-03-06
Created Jenkins pipeline for CoreNeuron

iomaganaris:
- Uses Jenkinsfile found in /tests/jenkins
- Builds AoS and SoA configuration of CoreNeuron
- Tests simple ringtest with both of them
- Clones master branch of https://github.com/pramodk/testcorenrn
  to load more tests
- Uses scripts found in /tests/jenkins directory to run all tests

### DIFF
--- a/tests/integration/CMakeLists.txt
+++ b/tests/integration/CMakeLists.txt
@@ -29,30 +29,28 @@ set(COMMON_ARGS "--tstop 100. --celsius 6.3 -mpi")
 set(RING_DATASET_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ring")
 set(RING_COMMON_ARGS "--datpath ${RING_DATASET_DIR} ${COMMON_ARGS}")
 set(RING_GAP_COMMON_ARGS "--datpath ${CMAKE_CURRENT_SOURCE_DIR}/ring_gap ${COMMON_ARGS}")
+set(PERMUTE1_ARGS "--cell-permute 1 --nwarp 8")
+set(PERMUTE2_ARGS "--cell-permute 2 --nwarp 16")
 if(ENABLE_OPENACC)
-    set(PERMUTE1_ARGS "--cell-permute 1 --nwarp 8 --gpu")
-    set(PERMUTE2_ARGS "--cell-permute 2 --nwarp 16 --gpu")
-else(ENABLE_OPENACC)
-    set(PERMUTE1_ARGS "--cell-permute 1 --nwarp 8")
-    set(PERMUTE2_ARGS "--cell-permute 2 --nwarp 16")
+  set(GPU_ARGS "--nwarp 8 --gpu")
 endif()
 
 # List of tests with arguments
 set(TEST_CASE_ARGS
-    "ring!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring"
-    "ring_binqueue!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_binqueue --binqueue"
-    "ring_multisend!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_multisend --multisend"
-    "ring_spike_buffer!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_spike_buffer -b 1"
-    "ring_duplicate!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_duplicate -z 3"
-    "ring_permute1!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute1 ${PERMUTE1_ARGS}"
-    "ring_permute2!${RING_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute2 ${PERMUTE2_ARGS}"
-    "ring_gap!${RING_GAP_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap"
-    "ring_gap_binqueue!${RING_GAP_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_binqueue --binqueue"
-    "ring_gap_multisend!${RING_GAP_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_multisend --multisend"
-    "ring_gap_permute1!${RING_GAP_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_permute1 ${PERMUTE1_ARGS}"
-    "ring_gap_permute2!${RING_GAP_COMMON_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_permute2 ${PERMUTE2_ARGS}")
+    "ring!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring"
+    "ring_binqueue!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_binqueue --binqueue"
+    "ring_multisend!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_multisend --multisend"
+    "ring_spike_buffer!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_spike_buffer -b 1"
+    "ring_duplicate!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_duplicate -z 3"
+    "ring_permute1!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute1 ${PERMUTE1_ARGS}"
+    "ring_permute2!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute2 ${PERMUTE2_ARGS}"
+    "ring_gap!${RING_GAP_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap"
+    "ring_gap_binqueue!${RING_GAP_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_binqueue --binqueue"
+    "ring_gap_multisend!${RING_GAP_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_gap_multisend --multisend"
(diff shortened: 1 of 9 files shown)


deploy/Jenkinsfile
```diff
@@ -0,0 +1,227 @@
+pipeline {
+    agent {
```
pramodk: Could you move these files to test directory rather than new deploy?

iomaganaris: I will create a folder in `/tests` called `jenkins` and move all scripts used in the pipeline there

deploy/Jenkinsfile
```diff
@@ -0,0 +1,227 @@
...
+                    userRemoteConfigs: [[url: 'https://github.com/pramodk/testcorenrn.git']]
+                ])
+            }
+        }
+
+        stage('nrnivmodl'){
+            parallel{
+                stage('nrnivmodl'){
+                    steps{
+                        sh 'sh deploy/nrnivmodl.sh'
```
ferdonline: I have the impression that these scripts work in the same directory, even in running in parallel. Did I miss smtg?

deploy/Jenkinsfile
```diff
@@ -0,0 +1,227 @@
...
+                }
+                stage('nrnivmodl-core SoA'){
+                    steps{
+                        sh 'sh deploy/nrnivmodl-core.sh SoA'
+                    }
+                }
+            }
+        }
+
+        stage('testcorenrn'){
```
ferdonline: These stages can be created in a for loop. Something like:
```
script {
    def cmds = [:]
    for (vspec in VERSIONS_SPEC) {
        for (testname in TESTS) {
            def taskname = "${vspec.key}-${testname}"
            cmds[taskname] = """
                source .jenkins/testutils.sh
                run_test ${testname} ${vspec.value}
                """.toString()
        }
    }
    parallel cmds.collectEntries{
        key, cmd -> return [(key): { stage(key){sh(cmd)} } ] }
}
```

deploy/Jenkinsfile
iomaganaris: Yes, they all work in the same directory in parallel but do different tasks.

deploy/Jenkinsfile
iomaganaris: After a lot of tries and research I could not find a way to construct with a `for-loop` the same pipeline design with running in parallel a branch with more than one stages, so I suppose that I will leave this part of the code like that. :disappointed:

pramodk:
@iomaganaris  : I have some suggestions about enabling GPU builds in Jenkins test suite (see #145), lets discuss.

tests/integration/CMakeLists.txt
```diff
@@ -113,7 +111,9 @@ endif()
     elseif(${SUBTEST_NAME} MATCHES "spikereplay")
       set(TEST_ARGS "--datpath ${CMAKE_CURRENT_SOURCE_DIR}/${TEST_NAME} --outpath ${CMAKE_CURRENT_BINARY_DIR}/${SIM_NAME} --pattern ${CMAKE_CURRENT_SOURCE_DIR}/${SIM_NAME}/out.std --tstop 100. -mpi")
     endif()
-
+    if(ENABLE_OPENACC)
+      set(TEST_ARGS "${TEST_ARGS} --nwarp 8 --gpu")
```
pramodk: `--nwarp 8 --gpu` => can be `${GPU_ARGS}` ?

tests/jenkins/nrnivmodl-core.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+. /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
+module load intel/18.0.3 hpe-mpi/2.16
```
pramodk: remove versions `/18.0.3` & `/2.16` so that if new modules are available, we dont have to change this.

tests/jenkins/nrnivmodl-core.sh
```diff
@@ -0,0 +1,8 @@
...
+CORENRN_TYPE=$1
+cd $WORKSPACE/testcorenrn
+rm -rf $CORENRN_TYPE
```
pramodk: `$CORENRN_TYPE` => could you use here full path? `$WORKSPACE/testcorenrn/$CORENRN_TYPE`

tests/jenkins/nrnivmodl.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+. /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
+module load neuron/2018-10/python3/parallel
```
pramodk: same here

tests/jenkins/ringtest.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/bash
+
+module load hpe-mpi/2.16
```
pramodk: same

tests/jenkins/run_corenrn.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/bash
+
+module load hpe-mpi/2.16
```
pramodk: same

tests/jenkins/run_neuron.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/bash
+
+module load hpe-mpi/2.16
```
pramodk: same

tests/jenkins/nrnivmodl-core.sh
iomaganaris: Should the version definition be removed from `pgi`, `cuda`, `cmake` and `boost` as well?

tests/jenkins/nrnivmodl.sh
iomaganaris: I used this particular version because `module load neuron` loads by default `neuron/2018-10/python3/serial` and running `nrnivmodl` of this version to build the `special` for the `testcorenrn` fails with the following error:
```
libtool: link: unable to infer tagged configuration
libtool: link: specify a tag with `--tag'
```

tests/jenkins/nrnivmodl-core.sh
pramodk: yeah...I would remove them now. I believe we would be deploying compatible ones.

tests/jenkins/nrnivmodl.sh
pramodk: ok..understood. for neuron you can specify full module then.

tests/integration/CMakeLists.txt
```diff
@@ -113,7 +111,9 @@ endif()
     elseif(${SUBTEST_NAME} MATCHES "spikereplay")
       set(TEST_ARGS "--datpath ${CMAKE_CURRENT_SOURCE_DIR}/${TEST_NAME} --outpath ${CMAKE_CURRENT_BINARY_DIR}/${SIM_NAME} --pattern ${CMAKE_CURRENT_SOURCE_DIR}/${SIM_NAME}/out.std --tstop 100. -mpi")
     endif()
-
+    if(ENABLE_OPENACC)
```
ferdonline: No need for the `if`. $GPU_ARGS will be empty

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/bash
+
```
ferdonline: Add `set -e` to break on error

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
...
+. /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
+
+CORENRN_TYPE=$1
+
+if [ ${CORENRN_TYPE} = "GPU" ]
```
ferdonline: Use quotes to avoid the comparison to break when there are spaces in the var
`if [ "$CORENRN_TYPE" = "GPU" ]`

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
...
+then
+    module load pgi cuda hpe-mpi cmake
+
+    mkdir build_${CORENRN_TYPE}
+else
+    module load boost intel hpe-mpi cmake
+    export CC=mpicc
```
ferdonline: We dont need these exports for the GPU version?

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/bash
+
+. /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
+
+CORENRN_TYPE=$1
```
ferdonline: Quote, otherwise assignment might fail
`CORENRN_TYPE="$1"`

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
...
+    export CC=mpicc
+    export CXX=mpicxx
+
+    mkdir build_${CORENRN_TYPE} build_intel_${CORENRN_TYPE}
+fi
+
+cd $WORKSPACE/build_${CORENRN_TYPE}
+
+echo "${CORENRN_TYPE} build"
+if [ ${CORENRN_TYPE} = "GPU" ]
```
ferdonline: Quotes. Also, as a convention, it is more common to see
`if [...]; then`

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
...
+
+    mkdir build_${CORENRN_TYPE}
+else
+    module load boost intel hpe-mpi cmake
+    export CC=mpicc
+    export CXX=mpicxx
+
+    mkdir build_${CORENRN_TYPE} build_intel_${CORENRN_TYPE}
+fi
+
```
ferdonline: Maybe good to control the input
```
else if [ "$CORENRN_TYPE" = .... ]; then
    ....
else
   echo "Not a valid mode"
   exit 1
fi
```

tests/jenkins/install.sh
```diff
@@ -0,0 +1,57 @@
...
+      -G 'Unix Makefiles'  \
+      -DCMAKE_INSTALL_PREFIX=$WORKSPACE/install_${CORENRN_TYPE}/ \
+      -DCMAKE_BUILD_TYPE=Debug  \
+      -DENABLE_SOA=$ENABLE_SOA \
+      -DCORENEURON_OPENMP=$ENABLE_OPENMP \
+      -DTEST_MPI_EXEC_BIN="mpirun" -DTEST_EXEC_PREFIX="mpirun;-n;2" -DAUTO_TEST_WITH_SLURM=OFF -DAUTO_TEST_WITH_MPIEXEC=OFF \
+      $WORKSPACE/
+fi
+make VERBOSE=1 -j8
+make install
```
ferdonline: Newline at the of file(s)

tests/jenkins/nrnivmodl-core.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
```
ferdonline: Add `set -e`
Also in the next script files

tests/jenkins/install.sh
iomaganaris: I didn't know about this. Thank you for the addition!

tests/jenkins/install.sh
iomaganaris: You are right. The exports are needed to use the `hpe-mpi` compiler and are included in the tutorial of building `CoreNeuron for GPU`, however if they are not set the `PGI` compiler is used and `CoreNeuron` is still built without problems and without any big difference in performance.

ferdonline: CMake will handle both cases for you. (sorry I lost the source of this info...)
If you `export CC=mpicc` it will use the mpi compiler wrapper, if not, it will use the CXX compiler and detect the MPI flags for you.
I believe it would be better to be consistent, but I don't know if there are any gotchas. @pramodk ?

iomaganaris:
please retest

PR #146, opened 2019-03-08
Fix cmake options for MOD2C

pramodk:


pramodk:
Ok..I will close this then.

PR #155, opened 2019-03-19
Some minor fixes that I missed in the last PR

ohm314:
- add pythonpath for nmodl
- fix include directory finding


CMakeLists.txt
```diff
@@ -116,6 +116,11 @@ if (ENABLE_NMODL)
     set(NMODL_EXTRA_FLAGS "" CACHE STRING "Extra NMODL options such as passes")
     separate_arguments(NMODL_EXTRA_FLAGS UNIX_COMMAND "${NMODL_EXTRA_FLAGS}")
     include_directories(${nmodl_INCLUDE})
+    if("$ENV{PYTHONPATH}" STREQUAL "")
```
pramodk: may be this condition check is not needed? if you think its ok append empty string to python path

PR #156, opened 2019-03-19
Update instrumentation with Caliper

pramodk:


coreneuron/nrniv/main1.cpp
```diff
@@ -396,7 +399,10 @@ extern "C" int run_solve_core(int argc, char** argv) {
     }
 
     // initializationa and loading functions moved to separate
+    Instrumentor::phase_begin("load-model");
     nrn_init_and_load_data(argc, argv, configs.size() > 0);
```
tristan0x: s/configs.size() > 0/!config.empty()/

CMakeLists.txt
```diff
@@ -183,6 +184,18 @@ if(ENABLE_REPORTINGLIB)
     endif(reportinglib_FOUND)
 endif(ENABLE_REPORTINGLIB)
 
+# Code instrumentation
+if(ENABLE_CALIPER)
+    find_package(caliper REQUIRED)
+    include_directories(${caliper_INCLUDE_DIR})
+    add_definitions("-DCALIPER")
```
tristan0x: I would use a `CORENEURON_CALIPER` macro instead of `CALIPER`

coreneuron/nrnoc/fadvance_core.cpp
```diff
@@ -34,6 +34,9 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #include "coreneuron/nrniv/nrn_acc_manager.h"
 #include "coreneuron/utils/reports/nrnreport.h"
 #include "coreneuron/utils/progressbar/progressbar.h"
+#include "coreneuron/nrniv/profiler_interface.h"
+
+#include <sstream>
```
tristan0x: standard include should be before coreneuron ones. clang-format should do it automatically.

tristan0x:
If I am not mistaken, most of the end statements are in the same scope than the begin. So what an API that would leverages scopes like `scoped_lock` for instance:

Instead of:
```cpp
if (nrn_have_gaps) {
    Instrumentor::phase_begin("gap-v-transfer");
    nrnthread_v_transfer(_nt);
    Instrumentor::phase_end("gap-v-transfer");
}
```

have something like:
```cpp
if (nrn_have_gaps) {
    Instrumentor::phase p("gap-v-transfer");
    nrnthread_v_transfer(_nt);
}
```

pramodk:
> If I am not mistaken, most of the end statements are in the same scope than the corresponding begin. So what an API that would leverages scopes like `scoped_lock` for instance:
> 
> Instead of:
> 
> ```c++
> if (nrn_have_gaps) {
>     Instrumentor::phase_begin("gap-v-transfer");
>     nrnthread_v_transfer(_nt);
>     Instrumentor::phase_end("gap-v-transfer");
> }
> ```
> 
> have something like:
> 
> ```c++
> if (nrn_have_gaps) {
>     Instrumentor::phase p("gap-v-transfer");
>     nrnthread_v_transfer(_nt);
> }
> ```

yes, c++ example of caliper works like this : https://llnl.github.io/Caliper/AnnotationAPI.html#function-annotations

I think we need timer & profiler interface that we can use across different codes. That more generic implementation can be tackled in separate PR ?

I have addressed rest of the comments.

tristan0x:
> I think we need timer & profiler interface that we can use across different codes. That more generic implementation can be tackled in separate PR ?

Yes sure. No problem. I will create an issue about that to discuss it.

PR #168, opened 2019-04-26
Added instrumentor::phase struct API implementation

tapaswenipathak:
As there is existing implementation of `instrumentor` struct, I have created a `namespace` named `instrumentor` other than `detail` and kept `phase` struct inside it. Please let me know if you think the design should be different.

Fix https://github.com/BlueBrain/CoreNeuron/issues/158.

Issue #158, opened 2019-03-20
Use scope-like API for instrumentor


tristan0x:
Unless I am not mistaken, most of the `end` statements are in the same scope than the corresponding `begin`. So I suggest to use API that would leverages scopes like scoped_lock for instance.

Instead of:

```cpp
if (nrn_have_gaps) {
    Instrumentor::phase_begin("gap-v-transfer");
    nrnthread_v_transfer(_nt);
    Instrumentor::phase_end("gap-v-transfer");
}
```
have something like:

```cpp
if (nrn_have_gaps) {
    Instrumentor::phase p("gap-v-transfer");
    nrnthread_v_transfer(_nt);
}
```

tristan0x:
This improvement comes from #156

tapaswenipathak:
Hi @tristan0x, I read the occurrence of all such API usage, can I work on this?

tristan0x:
I will ask @pramodk to perform the triage of the issues so that it is easier for you to know what you can do to help us. But yes, I think you can start implementing this. Do not hesitate to ask questions.
Thanks!

tapaswenipathak:
I completed `Instrumentor` in my [fork branch](https://github.com/tapaswenipathak/CoreNeuron/tree/api). I'm writing [`Caliper`](https://github.com/BlueBrain/CoreNeuron/blob/master/coreneuron/nrniv/profiler_interface.h#L44-L56) using https://llnl.github.io/Caliper/AnnotationAPI.html#function-annotations then  `CUDA_PROFILING`, `Tau`, `CRAYPAT`. I think `NullInstrumentor` shouldn't have a implementation? Please let me know if I am missing anything.

tapaswenipathak:
I completed `Instrumentor` in my [fork branch](https://github.com/tapaswenipathak/CoreNeuron/tree/api). I'm writing [`Caliper`](https://github.com/BlueBrain/CoreNeuron/blob/master/coreneuron/nrniv/profiler_interface.h#L44-L56) using https://llnl.github.io/Caliper/AnnotationAPI.html#function-annotations. Can you please let me know if  the implementation is required in  `CUDA_PROFILING`, `Tau`, `CRAYPAT` and `NullInstrumentor` as well. There is nearly no usage of these?

tapaswenipathak:
[...]

PR #170, opened 2019-04-30
Add Phase API

tapaswenipathak:
This adds Caliper, CudaProfiling, CrayPat and Tau `phase` API.

I didn't find any Caliper, CudaProfiling, CrayPat and Tau API usage as such so just added the API implementation. I created `Caliper`,  `CudaProfiling`, `CrayPat` and `Tau` namespace, please let me know if you would prefer different design.

### DIFF
--- a/coreneuron/nrniv/profiler_interface.h
+++ b/coreneuron/nrniv/profiler_interface.h
@@ -55,6 +55,17 @@ struct Caliper {
     inline static void stop_profile(){};
 };
 
+namespace Caliper {
+    struct phase {
+        phase() {
+            detail::Caliper::phase_begin();
+				}
+        ~phase() {
+            detail::Caliper::phase_end();
+        }
+    }
+}
+
 #endif
 
 #if defined(CUDA_PROFILING)
@@ -73,6 +84,17 @@ struct CudaProfiling {
     };
 };
 
+namespace CudeProfiling {
+    struct phase {
+        phase() {
+            detail::CudeProfiling::phase_begin();
+        }
+        ~phase() {
+            detail::CudeProfiling::phase_end();
+        }
+    }
+}
+
 #endif
 
 #if defined(CRAYPAT)
@@ -90,6 +112,18 @@ struct CrayPat {
         PAT_record(PAT_STATE_OFF);
     };
 };
+
+namespace CrayPat {
+    struct phase {
+        phase() {
+            detail::CrayPat::phase_begin();
+        }
+        ~phase() {
+            detail::CrayPat::phase_end();
+        }
+    }
+}
+
 #endif
 
 #if defined(TAU)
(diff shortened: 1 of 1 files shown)


tristan0x:
> I didn't find any Caliper, CudaProfiling, CrayPat and Tau API usage a

They are used here:
https://github.com/BlueBrain/CoreNeuron/blob/6635cf5a3938256dccc8e20f48c59a8e480861db/coreneuron/nrniv/profiler_interface.h#L167-L180

> I created `Caliper`, `CudaProfiling`, `CrayPat` and `Tau` namespace, please let me know if you would prefer different design.

`coreneuron::detail::Instrumentor` class provides an abstraction layer on top of the different instrumentors (`coreneuron::detail::Caliper`, `coreneuron::detail::CudaProfiling`, ...).

So the scope API this PR should provide may use this `coreneuron::detail::Instrumentor` class instead.

tapaswenipathak:
I don't think I get you. `template <class... TProfilerImpl>` is being implemented like

```c++
using Instrumentor = detail::Instrumentor<
#if defined CORENEURON_CALIPER
    detail::Caliper,
#endif
#if defined(CUDA_PROFILING)
    detail::CudaProfiling,
#endif
#if defined(CRAYPAT)
    detail::CrayPat,
#endif
#if defined(TAU)
    detail::Tau,
#endif
    detail::NullInstrumentor>;
```

W/ usage I meant where you just call the API w/ `phase::begin` and `phase::end` and would replace with `phase::p` scoped_lock mechanism. `phase::p` calling the function of implemented `phase::begin` and `phase::end`, right?

Now the non scoped API implementation should be under separate namespace? I [PRed](https://github.com/BlueBrain/CoreNeuron/pull/168) this as well.

tapaswenipathak:
I don't think I get you. `template <class... TProfilerImpl>` is being implemented as

```c++
using Instrumentor = detail::Instrumentor<
#if defined CORENEURON_CALIPER
    detail::Caliper,
#endif
#if defined(CUDA_PROFILING)
    detail::CudaProfiling,
#endif
#if defined(CRAYPAT)
    detail::CrayPat,
#endif
#if defined(TAU)
    detail::Tau,
#endif
    detail::NullInstrumentor>;
```

W/ usage I meant where you just call the API w/ `phase::begin` and `phase::end` and would replace with `phase::p` scoped_lock mechanism. `phase::p` calling the function of implemented `phase::begin` and `phase::end`, right?

The non scoped API implementation should be under separate namespace? I [PRed](https://github.com/BlueBrain/CoreNeuron/pull/168) this as well.

tapaswenipathak:
@tristan0x ping, do you think I should add anything in this PR? Please let me know.

tapaswenipathak:
hi @tristan0x, do you think I enhance anything else in this PR? Please let me know.

tristan0x:
Hello @tapaswenipathak 
AFAIK, there is no use case for the C++ classes added by this PR. However #168 seems to implement #158 perfectly. So I am in favor of closing this PR

PR #174, opened 2019-05-08
Likwid instrumentor

sharkovsky:
This Pull Request introduces the possibility to use the [Likwid Marker API](https://github.com/RRZE-HPC/likwid/wiki/TutorialMarkerC) to instrument the code for profiling.

I had to introduce two significant changes:
1. `Findlikwid.cmake` to automatically discover Likwid;
2. Adding `init_profile` and `finalize_profile` to the `Instrumentor`'s API, because Likwid requires special init and finalize functions to be called before any memory allocation is done.


coreneuron/nrniv/profiler_interface.h
```diff
@@ -37,6 +41,12 @@ struct Instrumentor {
     inline static void stop_profile() {
         std::initializer_list<int>{(TProfilerImpl::stop_profile(), 0)...};
     }
+    inline static void init_profile() {
```
tristan0x: The `Instrumentor` static methods are missing documentation. I might be mistaken but it seems to me that `start_profile` static function serves the same purpose than the new `init_profile`, same for `stop_profile` / `finalize_profile`.

CMake/packages/Findlikwid.cmake
```diff
@@ -0,0 +1,18 @@
+# from git@github.com:PolyJIT/polli.git
```
tristan0x: This CMake module doesn't follow the _standards_ [requirements](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules)

This module is also missing documentation in prelude that details the list of variables defined. I also recommend using the [find_package_handle_standard_args ](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) CMake function to take care of most things.

You can take inspiration from supported modules in CMake install directory.

CMakeLists.txt
```diff
@@ -195,6 +196,13 @@ if(ENABLE_CALIPER)
     endif()
 endif(ENABLE_CALIPER)
 
+# Code instrumentation
+if(ENABLE_LIKWID)
+    find_package(likwid REQUIRED)
+    include_directories(${LIKWID_INCLUDE_DIR})
```
tristan0x: The macro should be prefixe by `CORENEURON_` to avoid possible collision.

apps/coreneuron.cpp
```diff
@@ -27,7 +27,10 @@ THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 #include <coreneuron/enginemech.h>
+#include "coreneuron/nrniv/profiler_interface.h"
 
 int main(int argc, char** argv) {
-    return solve_core(argc, argv);
+    coreneuron::Instrumentor::init_profile();
+    solve_core(argc, argv);
```
tristan0x: the `main` function should return what the `solve_core` function returns.

coreneuron/nrniv/profiler_interface.h
sharkovsky: Could you please take a look at c4c7a4f538baef7317eb7de4ee6250f47fa6047a?

apps/coreneuron.cpp
sharkovsky: Should be solved by 45a48f64aa4640131ac94ce21eec8f1919bd5559

CMakeLists.txt
sharkovsky: could you please take a look at 3f280985ac66b40f152a558dcee97d374c40be42?

CMake/packages/Findlikwid.cmake
sharkovsky: could you please take a look at 3f280985ac66b40f152a558dcee97d374c40be42?

CMakeLists.txt
tristan0x: That is very good! Thank you.
Please apply `cmake-format` utility (I think the variables list in the comments should be prefixed with '*' to prevent being reformated)

apps/coreneuron.cpp
tristan0x: Thank you!

coreneuron/nrniv/profiler_interface.h
tristan0x: This commit is amazing! Thank you so much for improving this.

CMakeLists.txt
ohm314: We currently don't use `cmake-format` in this repository. So, I think this is quite alright with the latest changes :)

coreneuron/nrniv/profiler_interface.h
```diff
@@ -106,6 +209,42 @@ struct Tau {
...
+    inline static void start_profile() {};
+
+    inline static void stop_profile() {};
+
+    inline static void init_profile(){
+        LIKWID_MARKER_INIT;
+
+#pragma omp parallel
+        {
+            LIKWID_MARKER_THREADINIT;
```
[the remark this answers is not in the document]
sharkovsky: When threading is **enabled**:
    - you must call **both** `LIKWID_MARKER_INIT` and `LIKWID_MARKER_THREADINIT`.
When threading is **disabled**:
    - you can potentially call only `LIKWID_MARKER_INIT`, but I've confirmed with a simple test that calling also `LIKWID_MARKER_THREADINIT` does not seem to pose a problem.

CMakeLists.txt
tristan0x: That's fine then indeed.

PR #177, opened 2019-05-22
Fixing soname/install_name mac linking probs

ferdonline:
**Problem**
When creating libnrnmech in mac the subsequent link would bring part of
the original output dir into the link entry. That is because mac writes
all outpyt libname in the install_name (similar to soname). Further, to
be foundable via rpath, `@rpath` had to be in the name.

This patch fixes both issues and special-core now works in MAC


ferdonline:
retest this please

Issue #95, opened 2018-06-05
Streamlining exit on error procedures


fouriaux:
create_report_configuration can fail and raise error using abort(), this lead to an error not necessary easy to interpret from a normal user without a debugger.

this abort and potentially other ones must be replaced by proper error management.

tapaswenipathak:
Hi @fouriaux, if you have branch let me know otherwise I will PR.

pramodk:
@tapaswenipathak : I dont think we started anything on this. But I would keep this in queue for time being as we are going to change overall execution workflow soon.

tapaswenipathak:
Hi @pramodk, can I work on this or is this internal? I can create a draft PR for proper error management?

pramodk:
@tapaswenipathak : Sure. We should consider parallel aspects while error handling (i.e. MPI executions). I haven't thought about this in detail but if you already thinking about this / some abstraction, let me know how.

tapaswenipathak:
@pramodk Can I go ahead and implement https://github.com/gabime/spdlog? Do you have any package in mind which would work w/ MPI executions?

tapaswenipathak:
I just saw `coreneuron/scopmath_core/abort.cpp` file but `abort_run` is being used at very few places, should I implement that completely?

tapaswenipathak:
ping @pramodk, do you think I can enhance the PR? https://github.com/BlueBrain/CoreNeuron/pull/171, please let me know.

olupton:
See also: #316.

PR #180, opened 2019-06-17
Implement logging

tapaswenipathak:
Fix https://github.com/BlueBrain/CoreNeuron/issues/95.

- [ ] Run and write few test logs


coreneuron/logging/easylogging.h
```diff
@@ -0,0 +1,6663 @@
+//
```
ohm314: We don't favor simply including external libraries. Much better would be to add git submodules for such libraries

coreneuron/logging/logging.hpp
```diff
@@ -0,0 +1,477 @@
...
+#endif
+#include <string>
+using namespace std;
+
+#ifdef WITH_MPI
+  #include <sstream>
+  #include <mpi.h>
+#endif
+
+#include "CORENEURON/config.hpp"
```
ohm314: Not sure what this is referring to?

coreneuron/logging/logging.hpp
```diff
@@ -0,0 +1,477 @@
...
+    //! @{
+    static const string bold;
+    static const string underline;
+    //! @}
+
+    //! resets both, colour and style
+    static const string reset;
+};
+
+const string OUT::black = "\033[30m";
```
ohm314: Why would those assignments not be done inside the struct?

coreneuron/logging/logging.hpp
```diff
@@ -0,0 +1,477 @@
...
+  #define ML_CVLOG(verbose_level, logger_id, x) CVLOG(verbose_level, logger_id) << x
+
+  /**
+   * same as `CVLOG_IF(condition, verbosity, logger, x)` from easylogging++
+   *
+   * @see @ref logging_macros
+   */
+  #define ML_CVLOG_IF(condition, verbose_level, logger_id, x) CVLOG_IF(condition, verbose_level, logger_id) << x
+#else
+  #define ML_LOG(level, x)
```
ohm314: I see here and further below that in your implementation (maybe that's the easylogging way to do it) you heavily rely on macros. I think that's a bad idea. We are trying to modernize this code, moving away from macros and defines as possible. C++ offers other, more powerful and flexible features to build and use APIs.

coreneuron/logging/logging.hpp
```diff
@@ -0,0 +1,477 @@
...
+ *
+ * @see CORENEURON::log::add_custom_logger()
+ */
+#define LOGGER_ID_LENGTH 10
+
+
+namespace CORENEURON
+{
+  /**
+   * Logging facilities for CORENEURONT++.
```
ohm314: I'm a bit puzzled about this naming?

coreneuron/logging/mpi_communicator_impl.hpp
```diff
@@ -0,0 +1,165 @@
+#include "coreneuron/mpi_communicator.hpp"
+
+#include "coreneuron/logging.hpp"
+
+
+namespace coreneuron
+{
+  namespace mpi
```
ohm314: As in many other places in this PR it's not entirely clear where your choice comes from and what this particular change is there for. There is a lot of documentation / commenting missing

coreneuron/nrniv/main1.cpp
```diff
@@ -437,9 +439,7 @@ extern "C" int run_solve_core(int argc, char** argv) {
         double tstop = nrnopt_get_dbl("--tstop");
 
         if (tstop < t && nrnmpi_myid == 0) {
-            printf("Error: Stop time (%lf) < Start time (%lf), restoring from checkpoint? \n",
-                   tstop, t);
-            abort();
+            ML_LOG(ERROR, "Error: Stop time" << tstop << "Start time" << "restoring from checkpoint? \n";
```
ohm314: This code cannot have been compiled. The line above could not compile, I believe.

coreneuron/nrniv/tqueue.cpp
```diff
@@ -518,8 +518,8 @@ void splay(SPBLK* n, SPTREE* q) {
 #ifdef DEBUG
     if (q->root != prev) {
         /*	fprintf(stderr, " *** bug in splay: n not in q *** " ); */
-        abort();
-    }
+        ML_LOG(FATAL, "fatal error");
+		}
```
ohm314: bad indenting...

coreneuron/nrnmpi/nrnmpi.cpp
```diff
@@ -198,7 +198,7 @@ void nrn_abort(int errcode) {
     } else
 #endif
     {
-        abort();
+			ML_LOG(FATAL, "fatal error");
```
ohm314: The program will still need to be aborted.

tapaswenipathak:
Thanks for all the review comments @ohm314, there is a lot of writing fixes I should do, my target would be finishing this by end of day.

tapaswenipathak:
Thanks for all the review comments @ohm314, there is a lot of writing fixes I should do, my target would be finishing this out in a day.

PR #181, opened 2019-06-18
Changed pinned parallel version of neuron to generic parallel version

iomaganaris:
Instead of using a specific version of neuron, query the available modules to find a parallel neuron module

### DIFF
--- a/tests/jenkins/nrnivmodl.sh
+++ b/tests/jenkins/nrnivmodl.sh
@@ -7,7 +7,7 @@ TEST_DIR="$1"
 
 . /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
 module load intel
-module load `module av 2>&1 | grep -o 'neuron.*/parallel'`
+module load `module av neuron 2>&1 | grep -o -m 1 'neuron.*/parallel' | awk -F' ' '{print $1}'`
 
 unset $(env|awk -F= '/^(PMI|SLURM)_/ {if ($1 != "SLURM_ACCOUNT") print $1}')
 
--- a/tests/jenkins/nrnivmodl.sh
+++ b/tests/jenkins/nrnivmodl.sh
@@ -1,13 +1,20 @@
 #!/usr/bin/bash
 
 set -e
-set +x
+set -x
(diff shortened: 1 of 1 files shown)


tests/jenkins/nrnivmodl.sh
```diff
@@ -6,7 +6,8 @@ set +x
 TEST_DIR="$1"
 
 . /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
-module load neuron/2018-10/python3/parallel intel
+module load intel
+module load `module av neuron 2>&1 | grep -o -m 1 'neuron.*/parallel' | awk -F' ' '{print $1}'`
```
tristan0x: The script should check that the crazy `module av ...`  command actually finds something, and properly fails otherwise.
Also maybe use ^ $ in the regex to enforce it a bit.

iomaganaris: That's a good improvement :+1: 
I added a check if a neuron version is found. Hope that this is sufficient

tests/jenkins/nrnivmodl.sh
```diff
@@ -1,12 +1,20 @@
 #!/usr/bin/bash
 
 set -e
-set +x
+set -x
```
pramodk: this will be just make module load verbose, may be we can turn back +x?

tests/jenkins/nrnivmodl.sh
```diff
@@ -1,12 +1,20 @@
...
-module load neuron/2018-10/python3/parallel intel
+module load intel
+
+neuron_version=$(module av neuron 2>&1 | grep -o -m 1 '^neuron.*/parallel$' | awk -F' ' '{print $1}')
+if [[ $neuron_version ]]; then
+    module load $neuron_version
+else
+    echo "No compatible neuron version found."
+    exit 1
+fi
```
pramodk: Here we can do `module list` so that we see what's loaded.

iomaganaris: I added this in both cases to check what is loaded in any case

PR #182, opened 2019-06-26
Direct1

nrnhines:
CoreNEURON direct mode trajectory requests allow CoreNEURON to send back information
during (or after) a simulation to allow plotting of trajectories and movies in NEURON as well as
Vector.record. Should be extended/tested for GPU but this pull request is needed to be consistent
with current communication layer in NEURON.

### DIFF
--- a/coreneuron/nrnoc/fadvance_core.cpp
+++ b/coreneuron/nrnoc/fadvance_core.cpp
@@ -235,7 +235,7 @@ void nrncore2nrn_send_values(NrnThread* nth) {
     // if per time step transfer, need to call nrn_record_init() in NEURON.
     // if storing full trajectories in CoreNEURON, need to initialize
     // vsize for all the trajectory requests.
-    (*nrn2core_trajectory_values_)(-1, 0, NULL, 0.0, 0, NULL);
+    (*nrn2core_trajectory_values_)(-1, 0, NULL, 0.0);
     for (int tid = 0; tid < nrn_nthread; ++tid) {
       NrnThread& nt = nrn_threads[tid];
       if (nt.trajec_requests) {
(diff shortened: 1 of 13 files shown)


coreneuron/nrnoc/fadvance_core.cpp
```diff
@@ -229,6 +230,44 @@ void nrn_ba(NrnThread* nt, int bat) {
     }
 }
 
+void nrncore2nrn_send_values(NrnThread* nth) {
+  if (nrn2core_trajectory_values_ == nullptr) {
+      // standalone execution : no callbacks
+      return;
+  }
```
pramodk: @nrnhines : I added above. Otherwise standalone CoreNEURON runs were segfaulting (as nrn2core_trajectory_values_ == nullptr).

coreneuron/nrnoc/fadvance_core.cpp
```diff
@@ -229,6 +230,44 @@ void nrn_ba(NrnThread* nt, int bat) {
...
+
+  if (nth == NULL) {
```
pramodk: @nrnhines : if I understood correctly we are using `nth==NULL` as special case here. Wouldn't to better to turn that into separate function?

Intuitively, if `nt==NULL`, we typically don't do anything. May be you can add this block to separate function like `initialize_trajectories`?

coreneuron/nrnoc/fadvance_core.cpp
```diff
@@ -289,8 +328,11 @@ void* nrn_fixed_step_lastpart(NrnThread* nth) {
 
         fixed_play_continuous(nth);
         nonvint(nth);
+        nrncore2nrn_send_values(nth);
         nrn_ba(nth, AFTER_SOLVE);
         nrn_ba(nth, BEFORE_STEP);
+    }else{
+        nrncore2nrn_send_values(nth);
```
pramodk: when `nt==NULL`, calling `nrncore2nrn_send_values` seems confusing at first sight.

coreneuron/nrnoc/fadvance_core.cpp
nrnhines: I agree. This was an oversight on my part.

coreneuron/nrnoc/fadvance_core.cpp
nrnhines: This was only needed on per step value transfer to tickle the gui in NEURON to initialize
in preparation to plot first points on next call.  There is also the claim for whole trajectory transfer
mode that one needs to initialize vector sizes.  The fragment could easily be put into a separate
function in coreneuron. However it is tedious to declare new callbacks into NEURON so it
seems simplest not to introduce one in the separate function.

coreneuron/nrnoc/fadvance_core.cpp
nrnhines: This is a case, not for nt==NULL but for nth->ncell  != 0 . I think I am imagining that someone may
want to plot a variable in an artificial cell.

pramodk: Ok understood !

PR #187, opened 2019-07-19
Added neuron direct test in CoreNeuron Jenkins pipeline

iomaganaris:
- Install neuron master branch with spack and run all the tests requiring neuron with it


tests/jenkins/install_neuron.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+
+export SPACK_INSTALL_PREFIX="${SPACK_INSTALL_PREFIX:-${WORKSPACE}/INSTALL_HOME}"
+BUILD_HOME="${WORKSPACE}/BUILD_HOME"
+export SPACK_ROOT="${BUILD_HOME}/spack"
+export PATH=$SPACK_ROOT/bin:/usr/bin:\$PATH
```
pramodk: Is this necessary? we typically `source spack/...setup_env.sh`.

tests/jenkins/neuron_direct.py
```diff
@@ -0,0 +1,27 @@
...
+tvstd = tv.cl()
+
+#h.CoreNeuronRun[0].run()
+pc = h.ParallelContext()
+h.cvode.cache_efficient(1)
+h.stdinit()
+pc.nrncore_run("-e %g"%h.tstop, 0)
+
+print ("Voltage times same: " + str(bool(tv.eq(tvstd))))
+print ("Voltage difference less than 1e-10: " + str(v.cl().sub(vstd).abs().max() < 1e-10))
```
pramodk: Curious how this test fails. e.g. if `False` is printed, does this test fails?

tests/jenkins/nrnivmodl.sh
```diff
@@ -6,17 +6,10 @@ set +x
 TEST_DIR="$1"
 
 . /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
-module load intel
+export SPACK_INSTALL_PREFIX="${SPACK_INSTALL_PREFIX:-${WORKSPACE}/INSTALL_HOME}"
+export MODULEPATH=$SPACK_INSTALL_PREFIX/modules/tcl/linux-rhel7-x86_64:$MODULEPATH
```
pramodk: `linux-rhel7-x86_64` this will change with OS change. I don't remember command to get modules directory but could you check in spack issues?

tests/jenkins/run_neuron_direct.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -e
+set -x
+
+CORENRN_TYPE="$1"
+
+unset MODULEPATH
+. /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
+export SPACK_INSTALL_PREFIX="${SPACK_INSTALL_PREFIX:-${WORKSPACE}/INSTALL_HOME}"
+export MODULEPATH=$SPACK_INSTALL_PREFIX/modules/tcl/linux-rhel7-x86_64:$MODULEPATH
```
pramodk: same

tests/jenkins/run_neuron_direct.sh
```diff
@@ -0,0 +1,31 @@
...
+    nrniv -python $WORKSPACE/tests/jenkins/neuron_direct.py 2>&1 | tee output.txt
+fi
+
+if [[ ! $(grep "Voltage times same: True" output.txt) ]]; then
+    exit 1
+fi
+
+if [[ ! $(grep "Voltage difference less than 1e-10: True" output.txt) ]]; then
+    exit 1
+fi
```
pramodk: Ahh now I see..might be better if that python script returns error (non-zero return code) rather comparing output like this?

tests/jenkins/envutils.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
pramodk: Is this required? I would prefer to use minimal setup here!

tests/jenkins/envutils.sh
```diff
@@ -0,0 +1,53 @@
...
+
+if [[ -z "$ADDITIONAL_ENV_VARS" && -n "$GERRIT_CHANGE_COMMIT_MESSAGE" ]]; then
+    ADDITIONAL_ENV_VARS=$(set +x; echo "$GERRIT_CHANGE_COMMIT_MESSAGE" | sed -n "s/^ENV_VARS://p")
+fi
+log "Checking for override hooks (ADDITIONAL_ENV_VARS or gerrit 'ENV_VARS:')..."
+set -$_setbk
+[ "$ADDITIONAL_ENV_VARS" ] && eval $ADDITIONAL_ENV_VARS
+
+
+bb5_run() (
```
pramodk: this is not used?

tests/jenkins/spack_setup.sh
```diff
@@ -0,0 +1,106 @@
...
+      ${PKG_NAME}@${PKG_VERSION}${PKG_VARIANT}: ${PKG_MODULE}'
+
+    # mpi4py (work w intel stack)
+    echo "$external_pkg_module_tpl" | PKG_NAME=py-mpi4py PKG_VERSION=99 PKG_PATH=. PKG_MODULE=py-mpi4py \
+        envsubst >> $SPACK_ROOT/etc/spack/packages.yaml
+
+    for pkg in numpy h5py lazy-property setuptools; do
+        echo "$external_pkg_tpl" | PKG_NAME=py-$pkg PKG_VERSION=99 PKG_PATH=pydeps envsubst >> $SPACK_ROOT/etc/spack/packages.yaml
+    done
+)
```
pramodk: Is this required? delete / simplify

tests/jenkins/run_neuron_direct.sh
iomaganaris: You are right, I agree. I will change this

tests/jenkins/install_neuron.sh
iomaganaris: `export PATH=$SPACK_ROOT/bin:/usr/bin:\$PATH` is still needed. There are some issues with `git` when running `spack install` only by using `source spack/../setup_env.sh`

tests/jenkins/nrnivmodl.sh
iomaganaris: I tried looking for this but I didn't find anything. Another way I found to solve this is to use `$(spack arch)` instead of `linux-rhel7-x86_64`, similarly to what is done in other jenkins plans

tests/jenkins/envutils.sh
iomaganaris: You are right. It is not needed and will be removed :+1:

tests/jenkins/run_neuron_direct.sh
iomaganaris: Answered in the previous comment

tests/jenkins/envutils.sh
iomaganaris: Yes, this can be removed. I forgot it. Thanks for pointing that out :+1:

tests/jenkins/spack_setup.sh
iomaganaris: This can be removed as well :+1:

PR #188, opened 2019-07-23
Moved deletion of nrnthread_chkpnt to nrn_cleanup 

iomaganaris:
- Avoids segmentation fault when accessing it in write_phase2 of nrn_checkpoint.cpp


iomaganaris:
please retest

PR #192, opened 2019-08-09
Fix bug when spikes vector is empty

jorblancoa:
The bug is reproduced when calling std::min/max_element() when spikevec_time is empty.

The use of those functions with an empty range returns the end iterator. Attempting to dereference this iterator is undefined behavior (segmentation fault in most cases).

### DIFF
--- a/coreneuron/nrniv/output_spikes.cpp
+++ b/coreneuron/nrniv/output_spikes.cpp
@@ -95,8 +95,12 @@ void local_spikevec_sort(std::vector<double>& isvect,
 #if NRNMPI
 
 void sort_spikes(std::vector<double>& spikevec_time, std::vector<int>& spikevec_gid) {
-    double lmin_time = *(std::min_element(spikevec_time.begin(), spikevec_time.end()));
-    double lmax_time = *(std::max_element(spikevec_time.begin(), spikevec_time.end()));
+    double lmin_time = std::numeric_limits<double>::max();
+    double lmax_time = std::numeric_limits<double>::min();
+    if(!spikevec_time.empty()) {
+        lmin_time = *(std::min_element(spikevec_time.begin(), spikevec_time.end()));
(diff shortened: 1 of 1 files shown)


coreneuron/nrniv/output_spikes.cpp
```diff
@@ -95,8 +95,12 @@ void local_spikevec_sort(std::vector<double>& isvect,
 #if NRNMPI
 
 void sort_spikes(std::vector<double>& spikevec_time, std::vector<int>& spikevec_gid) {
-    double lmin_time = *(std::min_element(spikevec_time.begin(), spikevec_time.end()));
```
ohm314: good catch. Maybe it would be good to also guard against the case where there were no spikes at all. maybe with something like
```c++
auto nspikes = nrnmpi_int_allmax(spikevec_time.size());
if (nspikes == 0) {
    return;
}
```

jorblancoa: I totally agree with this conceptually, however everything should work fine even with 0 total spikes.
Do you think it worth the time spent doing the allreduce on the 99.9% of cases where there would be spikes against the time saved when there are 0 spikes?

ohm314: You are right of course, this is overkill and is implicitly handled further down.

coreneuron/nrniv/output_spikes.cpp
```diff
@@ -95,8 +95,12 @@ void local_spikevec_sort(std::vector<double>& isvect,
...
-    double lmax_time = *(std::max_element(spikevec_time.begin(), spikevec_time.end()));
+    double lmin_time = std::numeric_limits<int>::max();
```
ohm314: ok well, I do have one more question. Why do you use `std::numeric_limits<int>` and not `std::numeric_limits<double>` ?

jorblancoa: You are right, I will change this.

PR #198, opened 2019-10-05
CoreNEURON CMake refactoring for NEURON integration (5 years of cleanup!)

pramodk:
Goal of this PR is
  - remove unnecessary CMake options
  - partial cleanup of existing implementation

Todos
~- [ ] Update README~
~- [ ] Update Spack recipes~
~- [ ] Launch CI with the appropriate Spack branch~
- [x] Fix boost linking on BB5


pramodk:
@iomaganaris  : this is not in perfect shape but already significant refactoring of what was there. Could you skim through it and see if anything obvious you can see?

README.md
```diff
@@ -42,12 +42,12 @@ cmake ..
 make -j
 ```
 
-If you don't have MPI, you can disable MPI dependency using the CMake option `-DENABLE_MPI=OFF`:
+If you don't have MPI, you can disable MPI dependency using the CMake option `-DCORENRN_ENABLE_MPI=OFF`:
```
iomaganaris: I was thinking if the `CORENRN_<option>` prefix is needed to be added in the various options. It doesn't seem to me very practical for the user to have to write this prefix multiple times for generic options. But it just might me, so feel free to keep it :smile:

CMake/packages/Findnmodl.cmake
```diff
@@ -35,7 +14,7 @@
 # Using nmodl:
 #
 # ::
-#   set(NMODL_ROOT "" CACHE PATH "Path nmodl source-to-source compiler root")
+#   set(CORENRN_NMODL_ROOT "" CACHE PATH "Path nmodl source-to-source compiler root")
```
iomaganaris: In this case maybe `CORENRN_` prefix makes sense to distinguish the new `NMODL` from the old implementations :+1:

CMake/OpenAccHelper.cmake
```diff
@@ -0,0 +1,28 @@
...
+    set(ACC_FLAGS "-acc -Minline=size:200,levels:10")
+    set(CMAKE_C_FLAGS "${ACC_FLAGS} ${CMAKE_C_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${ACC_FLAGS} ${CMAKE_CXX_FLAGS}")
+  else()
+    message(WARNING "Non-PGI compiler : make sure to add required compiler flags to enable OpenACC")
+  endif()
+
+  find_package(CUDA 5.0 REQUIRED)
+  set(CUDA_SEPARABLE_COMPILATION ON)
+  add_definitions(-DCUDA_PROFILING)
```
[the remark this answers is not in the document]
pramodk: the profiling aspect is enabled for very high level functions so it should be ok in terms of overhead (but this needs to be re-looked later)

README.md
pramodk: the problem is when we use CoreNEURON with NRN as submodule, you get all CMake options mixed. And hence you need a way to distinguish those (and this is commonly done).

CMake/TestScriptUtils.cmake
```diff
@@ -1,29 +1,8 @@
...
-# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
-# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
-# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
-# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
-# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
-# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
-# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
-# THE POSSIBILITY OF SUCH DAMAGE.
-
+# =============================================================================
```
pramodk: I removed licenses in CMake files only at the moment (which I should have done in one go) but just was annoyed of copying full license in every file especially when code is smaller than license text itself.

CMake/ClangFormatHelper.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+  foreach(R123_PATH ${RANDOM123_FILES})
+    list(REMOVE_ITEM SRC_FILES_FOR_CLANG_FORMAT ${R123_PATH})
+  endforeach(R123_PATH)
+
+  add_custom_target(formatsource
+                    COMMAND ${CMAKE_COMMAND} -DSOURCE_FILES:STRING="${SRC_FILES_FOR_CLANG_FORMAT}"
+                            -DCLANG_FORMAT_EXECUTABLE=${CLANG_FORMAT_EXECUTABLE} -P
+                            "${CORENEURON_PROJECT_SOURCE_DIR}/CMake/ClangFormatUtils.cmake")
+endif()
```
pramodk: this will be replaced by hpc-conding convention submodule but I will do that in separate PR.

CMake/portability/BlueGenePortability.cmake
```diff
@@ -1,66 +0,0 @@
-# Copyright (c) 2016, Blue Brain Project
```
ohm314: my heart is broken about the removing of this file :)

coreneuron/nrnoc/multicore.cpp
```diff
@@ -212,26 +212,17 @@ void nrn_thread_table_check() {
 void nrn_multithread_job(void* (*job)(NrnThread*)) {
     int i;
 #if defined(_OPENMP)
-
 /* Todo : Remove schedule clause usage by using OpenMP 3 API.
  *        Need better CMake handling for checking OpenMP 3 support.
  *        Loop below is duplicated and default(none) removed to avoid
  *        Opari2 parsing errors.
  */
 // clang-format off
```
ohm314: that line isn't needed either I suppose

extra/CMakeLists.txt
```diff
@@ -1,63 +1,79 @@
...
-get_directory_property(CORENRN_COMPILE_FLAGS COMPILE_DEFINITIONS)
+# =============================================================================
+# Prepare nrnivmodl-core script with correct compiler flags
+# =============================================================================
 
-if (CORENRN_COMPILE_FLAGS)
-    set(CORENRN_COMPILE_DEFS "-D${CORENRN_COMPILE_FLAGS}")
-    string(REPLACE ";" " -D" CORENRN_COMPILE_DEFS "${CORENRN_COMPILE_DEFS}")
+# extract the COMPILE_DEFINITIONS property from the directory.
+get_directory_property(CORENRN_COMPILE_FLAGS COMPILE_DEFINITIONS)
```
ohm314: why do we need the directory scope here?

extra/CMakeLists.txt
```diff
@@ -1,63 +1,79 @@
...
+if(CORENRN_COMPILE_FLAGS)
+  set(CORENRN_COMPILE_DEFS "-D${CORENRN_COMPILE_FLAGS}")
```
ohm314: I've seen this pattern already above.. I think this should be done differently. First convert to list, then iterate and prepend `-D`. This works, but it's a bit weird.

extra/CMakeLists.txt
```diff
@@ -1,63 +1,79 @@
...
+string(REPLACE ";"
+               " "
+               CXX11_STANDARD_COMPILE_OPTION
+               "${CMAKE_CXX11_STANDARD_COMPILE_OPTION}")
 
 # MPI path might be multiple directories
-if (MPI_C_INCLUDE_PATH)
-    string(REPLACE ";" " -I" CORENRN_MPI_INCLUDES "${MPI_C_INCLUDE_PATH}")
+if(MPI_C_INCLUDE_PATH)
+  string(REPLACE ";"
```
ohm314: Don't you need to here, similarly to above also prepend a `-I` in front of the first path?

```
set(MPI_C_INCLUDE_ATH "-I${MPI_C_INCLUDE_PATH}")
```

CMake/ClangFormatHelper.cmake
```diff
@@ -0,0 +1,45 @@
+# =============================================================================
```
tristan0x: I am very surprised to see this file. I would love to see hpc-coding-standard used here instead.
Is at least the .clang-format file the one specified in the BBP coding standard?

CMake/ClangFormatHelper.cmake
```diff
@@ -0,0 +1,45 @@
...
+  # exclude random123
+  file(GLOB_RECURSE RANDOM123_FILES
+                    ${CORENEURON_PROJECT_SOURCE_DIR}/coreneuron/utils/randoms/Random123/*.cpp
+                    ${CORENEURON_PROJECT_SOURCE_DIR}/coreneuron/utils/randoms/Random123/*.h)
+
+  foreach(R123_PATH ${RANDOM123_FILES})
+    list(REMOVE_ITEM SRC_FILES_FOR_CLANG_FORMAT ${R123_PATH})
+  endforeach(R123_PATH)
+
+  add_custom_target(formatsource
```
tristan0x:
```suggestion
  add_custom_target(clang-format
```

to be consistent with [BBP Coding Conventions](https://github.com/BlueBrain/hpc-coding-conventions/tree/master/cpp#usage-1)

CMake/ClangFormatHelper.cmake
ohm314: Absolutely, @pramodk said he'll defer this to a seperate PR together with the formatting changes applied. Makes sense, I think.

CMake/packages/FindSLURM.cmake
```diff
@@ -1,46 +1,15 @@
...
-# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
-# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
-# THE POSSIBILITY OF SUCH DAMAGE.
-
-
-find_program(SLURM_SBATCH_COMMAND sbatch DOC "Path to the SLURM sbatch executable")
 find_program(SLURM_SRUN_COMMAND srun DOC "Path to the SLURM srun executable")
-find_program(SLURM_SACCTMGR_COMMAND sacctmgr DOC "Path to the SLURM sacctmgr executable")
-mark_as_advanced(SLURM_SRUN_COMMAND SLURM_SBATCH_COMMAND SLURM_SACCTMGR_COMMAND)
+mark_as_advanced(SLURM_SRUN_COMMAND)
```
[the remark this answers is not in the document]
pramodk: I don't think this is required..I will remove it now.

coreneuron/nrnoc/multicore.cpp
pramodk: done

extra/CMakeLists.txt
pramodk: Fernando did this but I assume this is the way to get all compiler flags : https://stackoverflow.com/questions/5403705/cmake-add-definitions-and-compile-definitions-how-to-see-them

extra/CMakeLists.txt
pramodk: yes...I noticed that this is not used anywhere so I will remove it

extra/CMakeLists.txt
pramodk: yeah agree.

CMake/ClangFormatHelper.cmake
pramodk: yeah agree, this should definitely go with use of hpc conding convention and I have already created issue https://github.com/BlueBrain/CoreNeuron/issues/200.

Note that this file new file is just extracted from [existing CMake](https://github.com/BlueBrain/CoreNeuron/blob/master/CMakeLists.txt#L231). I will work on this in a separate PR.

CMake/portability/BlueGenePortability.cmake
pramodk: You can't beat BlueGene ;)

CMakeLists.txt
```diff
@@ -1,335 +1,250 @@
...
-
-#Find all dependencies
-if (ENABLE_MPI)
-    find_package(MPI REQUIRED)
-    add_definitions("-DNRNMPI=1")
-else()
-    message(STATUS "Use of MPI disabled by user-provided option")
-    add_definitions("-DNRNMPI=0")
-    add_definitions("-DNRN_MULTISEND=0")
+check_include_files(malloc.h have_malloc_h)
```
ohm314: This type of checks have to go away

extra/CMakeLists.txt
ohm314: OK, I did not know that

pramodk:
Please retest

PR #199, opened 2019-10-15
Set PYTHONPATH for nrnivmodl-core workflow

pramodk:
- when cmake is run, PYTHONPATH is setup correctly and nmodl
    with sympy can be used.
  - for nrnivmodl-core workflow, PYTHONPATH needs to be set
    in the makefile so that nrnivmodl is self contained.


st4rl3ss:
LGTM!

PR #204, opened 2019-10-22
Make scopmath a depdendency for coreneuron

ohm314:
- Fixed the add_custom_command call
- Added scopmath as a dependency to coreneuron as it is needed for the
  custom command to succeed (it links with scopmath down the line).


ohm314:
File was renamed in this commit already, that should have happened elsewhere, sorry. I'll fix it

PR #209, opened 2019-10-26
mingw/windows build fixes

pramodk:
coreneuron now build as submodule of neuron on windows and runs ring test:

Screenshot 2019-10-27 at 01 24 39

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(VERSION_STRING "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}")
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 # =============================================================================
 # Settings to enable project as submodule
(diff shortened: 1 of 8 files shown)


pramodk:
@nrnhines : everything works if build static library of coreneuron (which is default). With shared library if I build as:

```
cmake ..  -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=`pwd`/install -DCOMPILE_LIBRARY_TYPE=SHARED
```

Then what produced is following and I get link error while running:

Screenshot 2019-10-27 at 12 40 46

If I build as submodule then everything works because I believe Cmake of neuron has some extra settings that influence how libraries/dlls are generated.

Do you remember what we have done in NEURON to properly generate dlls?

nrnhines:
I wonder if I'm using my own wrappers for dlopen, etc., implemented in
src/oc/cygwinprt.c. At least I see
```
$ nm libnrniv.dll.a |grep dlopen
0000000000000000 I __imp_dlopen_noerr
0000000000000000 T dlopen_noerr
0000000000000000 I __imp_dlopen
0000000000000000 T dlopen
```
and
```
$ grep DLFCN nrnconf.h
/* #undef HAVE_DLFCN_H */
```
and it is part of the liboc_la_SOURCES in nrn/src/oc/Makefile.am

nrnhines:
And also look at how dlopen is declared in src/nrniv/nrnpy.cpp with
```
#ifdef MINGW
#define RTLD_NOW 0
#define RTLD_GLOBAL 0
#define RTLD_NOLOAD 0
extern "C" {
extern void* dlopen_noerr(const char* name, int mode);
#define dlopen dlopen_noerr
extern void* dlsym(void* handle, const char* name);
extern int dlclose(void* handle);
extern char* dlerror();
}
#else
//#define _GNU_SOURCE
#include <dlfcn.h>
#endif
```

pramodk:
@nrnhines : note that we are not using dlopen in coreneuron. The issue is coming from the fact that the lib directory only contains `.dll.a` files which are helper files but actual `dll`s are in `build/coreneuron` directory which are not being copied in bin directory:

Screenshot 2019-10-27 at 18 52 35

pramodk:
Ok, got it! It's following CMake property:

```
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
```

PR #211, opened 2019-10-29
Support SONATA reports

jorblancoa:
- Adaptation for SONATA reports
- Send spike arrays


jorblancoa:
please retest

CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ include(CTest)
 option(CORENRN_ENABLE_OPENMP "Build the CORE NEURON with OpenMP implementation" ON)
 option(CORENRN_ENABLE_TIMEOUT "Enable nrn_timeout implementation" ON)
 option(CORENRN_ENABLE_REPORTINGLIB "Enable use of ReportingLib for soma reports" OFF)
+option(CORENRN_ENABLE_SONATA_REPORTS "Reports in SONATA format" OFF)
```
pramodk:
```suggestion
option(CORENRN_ENABLE_SONATA "Enable SONATA output format" OFF)
```

CMakeLists.txt
```diff
@@ -138,8 +139,17 @@ if(NOT CORENRN_ENABLE_TIMEOUT)
   add_definitions("-DDISABLE_TIMEOUT")
 endif()
 
+if ( CORENRN_ENABLE_SONATA_REPORTS AND NOT CORENRN_ENABLE_REPORTINGLIB )
```
pramodk: I see why you enable reportinglib as well but this is confusing! We should either:
- have reportinglib enabled
- or, sonata enabled

In order to make it work,  you can use `add_definitions("-DENABLE_REPORTING")` for sonata as well.

CMakeLists.txt
```diff
@@ -138,8 +139,17 @@ if(NOT CORENRN_ENABLE_TIMEOUT)
...
+  message(FATAL_ERROR "Can't have SONATA reports without enabling reportinglib")
+endif()
+
 if(CORENRN_ENABLE_REPORTINGLIB)
-  find_package(reportinglib REQUIRED)
+  if(CORENRN_ENABLE_SONATA_REPORTS)
+    find_package(reportinglibsonata REQUIRED)
+    add_definitions("-DENABLE_SONATA_REPORTS")
```
pramodk:
```suggestion
    add_definitions("-DENABLE_SONATA_REPORTS -DENABLE_REPORTING")
```
something like this I meant!

coreneuron/utils/reports/nrnreport.cpp
```diff
@@ -35,8 +35,12 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #include "coreneuron/nrnoc/mech_mapping.hpp"
 #include "coreneuron/nrnoc/membfunc.h"
 #ifdef ENABLE_REPORTING
+#ifdef ENABLE_SONATA_REPORTS
+#include "reportinglib/records.h"
```
pramodk: note that we will change naming of headers etc in sonata.

CMake/packages/Findreportinglibsonata.cmake
```diff
@@ -0,0 +1,41 @@
+# =============================================================================
```
pramodk: Didn't we say we will have just libsonata that will contain everything? In that case I would rename this to `Findlibsonata.cmake` (and there might be some already exist?).

CMake/packages/Findreportinglibsonata.cmake
```diff
@@ -0,0 +1,41 @@
...
+find_path(reportinglib_INCLUDE_DIR reportinglib/records.h)
+find_library(reportinglib_LIBRARY reportinglibsonata)
+get_filename_component(reportinglib_LIB_DIR ${reportinglib_LIBRARY} DIRECTORY)
+
+# Checks 'REQUIRED', 'QUIET' and versions.
+include(FindPackageHandleStandardArgs)
+
+find_package_handle_standard_args(reportinglibsonata
+  FOUND_VAR reportinglibsonata_FOUND
+  REQUIRED_VARS reportinglib_INCLUDE_DIR reportinglib_LIBRARY reportinglib_LIB_DIR)
```
pramodk: rename based on above comment : should be just libsonata?  (or, in worst case we need to have both, something like libsonatareport make more sense?)

CMakeLists.txt
jorblancoa: Should I check then that only one could be enabled? Or if both are enabled it means we use SONATA?

CMake/packages/Findreportinglibsonata.cmake
jorblancoa: What do you mean with that it might already exist? I think we should separate them anyway, maybe Findsonatareports.cmake as you mention in the next comment

CMake/packages/Findreportinglibsonata.cmake
jorblancoa: I like the libsonatareports name better

PR #230, opened 2020-01-10
Une new and delete instead of different kinds of malloc

alkino:
TEST BEFORE MERGE!

### DIFF
--- a/coreneuron/nrniv/cellorder.cpp
+++ b/coreneuron/nrniv/cellorder.cpp
@@ -65,13 +65,13 @@ InterleaveInfo& InterleaveInfo::operator=(const InterleaveInfo& info) {
 
 InterleaveInfo::~InterleaveInfo() {
     if (stride) {
-        free_memory(stride);
-        free_memory(firstnode);
-        free_memory(lastnode);
-        free_memory(cellsize);
+        delete[] stride;
+        delete[] firstnode;
+        delete[] lastnode;
+        delete[] cellsize;
     }
     if (stridedispl) {
-        free_memory(stridedispl);
+        delete[] stridedispl;
     }
     if (idle) {
         delete[] nnode;
--- a/coreneuron/nrniv/cellorder1.cpp
+++ b/coreneuron/nrniv/cellorder1.cpp
@@ -530,9 +530,9 @@ static void admin1(int ncell,
     // cellsize is the number of nodes in the cell not counting root.
     // nstride is the maximum cell size (not counting root)
     // stride[i] is the number of cells with an ith node.
-    firstnode = (int*)ecalloc_align(ncell, sizeof(int));
-    lastnode = (int*)ecalloc_align(ncell, sizeof(int));
-    cellsize = (int*)ecalloc_align(ncell, sizeof(int));
+    firstnode = new int[ncell];
+    lastnode = new int[ncell];
+    cellsize = new int[ncell];
 
     nwarp = (ncell % warpsize == 0) ? (ncell / warpsize) : (ncell / warpsize + 1);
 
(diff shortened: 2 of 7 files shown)


alkino:
Before my patch it was a bad mix "alloc vs delete", "new vs free", after it is maybe a mix between "new [] vs delete" and "new vs delete[]"

alkino:
please retest

coreneuron/nrniv/cellorder.cpp
```diff
@@ -65,13 +65,13 @@ InterleaveInfo& InterleaveInfo::operator=(const InterleaveInfo& info) {
 
 InterleaveInfo::~InterleaveInfo() {
```
tristan0x: The `InterleaveInfo` class is a good candidate for `std::unique_ptr` instead of manual new/delete.

coreneuron/nrniv/cellorder1.cpp
```diff
@@ -530,9 +530,9 @@ static void admin1(int ncell,
     // cellsize is the number of nodes in the cell not counting root.
     // nstride is the maximum cell size (not counting root)
     // stride[i] is the number of cells with an ith node.
-    firstnode = (int*)ecalloc_align(ncell, sizeof(int));
-    lastnode = (int*)ecalloc_align(ncell, sizeof(int));
-    cellsize = (int*)ecalloc_align(ncell, sizeof(int));
+    firstnode = new int[ncell];
```
tristan0x: From the function name, I assume `ecalloc_align` initializes the allocated memory to 0. So I would:

```c++
firstnode = new int[ncell]();
lastnode = new int[ncell]();
cellsize = new int[ncell]();
```

coreneuron/nrniv/cellorder1.cpp
```diff
@@ -556,7 +556,7 @@ static void admin1(int ncell,
         }
     }
 
-    stride = (int*)ecalloc_align(nstride + 1, sizeof(int));
+    stride = new int[nstride + 1];
```
tristan0x: Same here:
```suggestion
    stride = new int[nstride + 1]();
```

coreneuron/nrniv/cellorder1.cpp
```diff
@@ -616,11 +616,10 @@ static void admin2(int ncell,
     // ncore is the number of warps * warpsize
     nwarp = nodevec[ncell - 1]->groupindex + 1;
 
-    ncycles = (int*)ecalloc_align(nwarp, sizeof(int));
-    stridedispl =
-        (int*)ecalloc_align(nwarp + 1, sizeof(int));          // running sum of ncycles (start at 0)
-    rootbegin = (int*)ecalloc_align(nwarp + 1, sizeof(int));  // index (+1) of first root in warp.
-    nodebegin = (int*)ecalloc_align(nwarp + 1, sizeof(int));  // index (+1) of first node in warp.
+    ncycles = new int[nwarp];
```
tristan0x: Same here is `ecalloc_align` initializes memory.

coreneuron/nrniv/cellorder1.cpp
```diff
@@ -650,7 +649,7 @@ static void admin2(int ncell,
     }
 
     // strides
-    strides = (int*)ecalloc_align(nstride, sizeof(int));
+    strides = new int[nstride];
```
tristan0x: Same here is `ecalloc_align` initializes memory.

pramodk:
Before reviewing details, here are some general comments:
* we don't want to just overload global new/delete with target specific memory allocation (e.g. CUDA).
* this is because we don't want to allocate every object into device memory (or unified memory) due to memory limitations (e.g. ~16GB on GPUs compared to ~384GB on BB5 CPU)

I think we need policies with allocators. I suggested to look into [kokkos](https://github.com/kokkos/kokkos) / [RAJA](https://github.com/LLNL/RAJA) to see how they implement target specific allocators (e.g. [containers](https://github.com/kokkos/kokkos/tree/master/containers))

PR #231, opened 2020-01-15
fix for checkpoint saving issue in neurodamus-py:

WeinaJi:
NetCon with no target is allowed  in checkpoint
  set to pnttype = 0 and pntindex = -1

### DIFF
--- a/coreneuron/nrniv/nrn_checkpoint.cpp
+++ b/coreneuron/nrniv/nrn_checkpoint.cpp
@@ -428,21 +428,26 @@ static void write_phase2(NrnThread& nt, FileHandlerWrap& fh) {
     for (int i = 0; i < nnetcon; ++i) {
         NetCon& nc = nt.netcons[i];
         Point_process* pnt = nc.target_;
-        assert(pnt);  // nrn_setup.cpp allows type <=0 which generates nullptr target.
-        pnttype[i] = pnt->_type;
+        if (pnt == nullptr) {
+            // nrn_setup.cpp allows type <=0 which generates nullptr target.
+            pnttype[i] = 0;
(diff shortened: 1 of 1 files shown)


coreneuron/nrniv/nrn_checkpoint.cpp
```diff
@@ -428,21 +428,26 @@ static void write_phase2(NrnThread& nt, FileHandlerWrap& fh) {
...
-        int ix = pnt->_i_instance;
-        if (ml_pinv[pnt->_type]) {
-            ix = ml_pinv[pnt->_type][ix];
-        }
+            // todo: this seems most natural, but does not work. Perhaps should look
+            // into how pntindex determined in nrnbbcore_write.cpp and change there.
+            int ix = pnt->_i_instance;
+            if (ml_pinv[pnt->_type]) {
+                ix = ml_pinv[pnt->_type][ix];
+            }
```
ohm314: This code is in an `#if 0` should we not just drop it? If it's something that doesn't work, but still useful to keep maybe just make those 4 lines comments and remove the `#if 0`. @pramodk ?
In general we should rid the code of these ugly `#if 0` preproc branches.

WeinaJi:
Simulation stack CI got stuck in build stage, after rerunning in https://bbpcode.epfl.ch/ci/job/hpc.SimulationStack/1158/ , everything passed.

pramodk:
@WeinaJi @iomaganaris : will the simulation stack  failures will be fixed by #237?

iomaganaris:
They will be fixed by my latest change in the `blueconfigs` branch in `gerrit`.

iomaganaris:
please retest

PR #235, opened 2020-01-17
Small fixes in nrnivmodl regarding MODS_LOCATION

ferdonline:
MODS_LOCATION was always being overwritten from CLI. Now it can be
loaded from the save file.
Check if provided mods directory exists

### DIFF
--- a/extra/nrnivmodl-core.in
+++ b/extra/nrnivmodl-core.in
@@ -11,6 +11,7 @@ ROOTDIR=$(perl -e "use Cwd 'abs_path'; print abs_path('$_PARENT')")
 MAKE_OPTIONS="MECH_NAME NMODL_BINARY DESTDIR INCFLAGS LINKFLAGS MODS_PATH VERBOSE"
 PARALLELISM=4
 SAVE_FILE="@CMAKE_HOST_SYSTEM_PROCESSOR@/nrnivmodl_options.txt"
+params_MODS_PATH="."
 
 while getopts "n:m:v:d:i:l:p:hV" OPT; do
     case "$OPT" in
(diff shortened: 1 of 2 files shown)


pramodk:
> "no mod files found" warning transformed to an error

Is it necessary to change this behaviour? It's true that one can use just nrniv if there are no mod files but I have used `nrnivmodl .` without any mod files in `.` to build special in some workflows.

ferdonline:
> > "no mod files found" warning transformed to an error
> 
> Is it necessary to change this behaviour? It's true that one can use just nrniv/nrniv-core if there are no mod files but I have used `nrnivmodl/nrnivmodl-core .` without any mod files in `.` to build special in some workflows.

I thought about it and I understand the use case, but there were so many times we overlooked the warning and had trouble later... I think that's not an end-user use-case, if we want to allow it, what about the option "--allow-empty" ?

nrnhines:
I would prefer that nrnivmodl work correctly and quietly if there are no mod files. Ie just an empty
nrnmech.dll. I've been meaning to do that for a long time.

ferdonline:
> I've been meaning to do that for a long time.

You mean you'd like an option to silence warnings when building empty libraries?

nrnhines:
I just mean that nrnivmodl should work when there is no mod file just as it does when there are mod file. If I run nrnivmodl of standard NEURON in an empty folder, I presently get
```
$ nrnivmodl
Creating x86_64 directory for .o files.

/home/hines/neuron/ivcmake/temp
ls: cannot access '*.mod': No such file or directory
```
If I then create the mod file
```
echo "NEURON {SUFFIX nothing}" > temp.mod
```
nrnvimodl works fine.

ferdonline:
> @ferdonline : Am I correct that this is now same behaviour as in NEURON? 

It is the same behavior regarding we allow building from an empty mod set. However here we do these checks for the directory existence which Neuron doesnt seem to do.
Disclaimer, it seems that the jenkins pipeline didnt create the path properly as this is not passing... :/

iomaganaris:
> Disclaimer, it seems that the jenkins pipeline didnt create the path properly as this is not passing... :/

In the Jenkins plan we call `nrnivmodl-core mod-ext` with `mod-ext` not existing, so that we build a `special-core` without any additional mod files. I think that now the Jenkins plan should be edited to call `nrnivmodl-core` without pointing to any folder or create just an empty folder named `mod-ext`.

ferdonline:
> Nitpick: just curious why -x has been added (for end users).

`-x` was added to the test script, not end users

ferdonline:
@pramodk I dont have merge permissions

PR #237, opened 2020-01-20
Add branch selection for Neuron and Spack

iomaganaris:


tests/jenkins/Jenkinsfile
```diff
@@ -2,6 +2,14 @@ pipeline {
     agent {
         label 'bb5'
     }
+    parameters {
+        string(name: 'sha1', defaultValue: 'master',
```
ferdonline: Like we did in blueconfigs. We dont need to declare the variables that are passed. We only need to declare the variables that we want the user to edit. If we want to give the user the possibility of choosing a different commit  maybe we could use a more meaningful var name. In the logic we can then check both variables.

tests/jenkins/install_neuron.sh
```diff
@@ -1,15 +1,36 @@
...
 export SPACK_INSTALL_PREFIX="${SPACK_INSTALL_PREFIX:-${WORKSPACE}/INSTALL_HOME}"
-source $WORKSPACE/BUILD_HOME/spack/share/spack/setup-env.sh
+source ${SPACK_ROOT}/share/spack/setup-env.sh
 source /gpfs/bbp.cscs.ch/apps/hpc/jenkins/config/modules.sh
 export PATH=$WORKSPACE/BUILD_HOME/spack/bin:/usr/bin:$PATH
 export MODULEPATH=$SPACK_INSTALL_PREFIX/modules/tcl/$(spack arch):$MODULEPATH
 
 unset $(env|awk -F= '/^(PMI|SLURM)_/ {if ($1 != "SLURM_ACCOUNT") print $1}')
 
+patch_neuron
```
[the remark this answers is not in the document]
iomaganaris: I think that mostly the point is to run `neuron` to generate reference results. I thought though that it might be useful to test different branches of `Neuron` with `CoreNeuron` if we want to check their integration (for example `direct memory transfer`). Also this was useful to check the spack recipe of neuron I am working on with `CoreNeuron`.

pramodk:
> If that is the case, and if we can assume bb5, why not using the deployed neuron@develop?

Note that develop is moving target and hence it won’t work in this case and I assume we are not deploying neuron@develop).

tests/jenkins/Jenkinsfile
iomaganaris: True. I kept `sha1` because this is the variable to select the `CoreNeuron` branch but we can add another one (`CORENEURON_BRANCH`) to expose to the user and then select the branch using this one.

ferdonline:
> and I assume we are not deploying neuron@develop

my bad, I wanted to say `+debug`. We deploy `neuron@<latest>+debug` for the simulation stack tests, where we expect perfect results match. Wouldnt it make sense to do something similar here?

tests/jenkins/Jenkinsfile
ferdonline: > I kept sha1 because this is the variable to select the CoreNeuron branch

I meant we can remove from the `parameters`. It is anyway always passed as an environment variable

iomaganaris: True true. I will handle this parameter like we do in the `SimulationStack`.

iomaganaris:
> > and I assume we are not deploying neuron@develop
> 
> my bad, I wanted to say `+debug`. We deploy `neuron@<latest>+debug` for the simulation stack tests, where we expect perfect results match. Wouldnt it make sense to do something similar here?

You are right. I agree we should also use `neuron+debug@develop` in this CI.

PR #247, opened 2020-01-27
Remove pointtype from global class

alkino:


alkino:
Can someone merge it?

PR #252, opened 2020-02-06
Auto checkpointing

ferdonline:
This is WIP for a possible auto-checkpointing system.
The idea is to from time-to-time automatically create (& overwrite) previous checkpoints, so that in case a simulation fails after a long time not all progress gets lost.

TODO: As an extension to this concept, the coreneuron main could also detect SLURM allocation variables and program a checkpoint when the allocation time is about to expire.


ferdonline:
Closing PR in favor of new ones. Auto-checkpointing might produce too much data. We will consider it maybe at halfway

PR #254, opened 2020-02-10
Fix issue with progress bar not showing in gap junction simulations

iomaganaris:


coreneuron/sim/fadvance_core.cpp
```diff
@@ -108,11 +108,37 @@ void finalize_progress_bar() {
     }
 }
 
-void nrn_fixed_step_group_minimal(int n) {
-    static int step = 0;
+void nrn_fixed_single_steps_minimal(int total_steps, double tstop) {
+    const int progressbar_update_interval = 5;
+    double updated_tstop;
```
pramodk:
```suggestion
```

coreneuron/sim/fadvance_core.cpp
```diff
@@ -108,11 +108,37 @@ void finalize_progress_bar() {
...
+    static int current_steps = 0;
+    initialize_progress_bar(total_steps);
+#if NRNMPI
+    updated_tstop = tstop - dt;
+    nrn_assert(nrn_threads->_t <= tstop);
+    // It may very well be the case that we do not advance at all
+    while (nrn_threads->_t <= updated_tstop) {
+#else
+    updated_tstop = tstop - .5 * dt;
```
pramodk:
```suggestion
    double updated_tstop = tstop - .5 * dt;
```

coreneuron/sim/fadvance_core.cpp
```diff
@@ -108,11 +108,37 @@ void finalize_progress_bar() {
...
 
-void nrn_fixed_step_group_minimal(int n) {
-    static int step = 0;
+void nrn_fixed_single_steps_minimal(int total_steps, double tstop) {
+    const int progressbar_update_interval = 5;
+    double updated_tstop;
+    static int current_steps = 0;
+    initialize_progress_bar(total_steps);
+#if NRNMPI
+    updated_tstop = tstop - dt;
```
pramodk:
```suggestion
    double updated_tstop = tstop - dt;
```

coreneuron/sim/fadvance_core.cpp
```diff
@@ -108,11 +108,37 @@ void finalize_progress_bar() {
     }
 }
 
-void nrn_fixed_step_group_minimal(int n) {
-    static int step = 0;
+void nrn_fixed_single_steps_minimal(int total_steps, double tstop) {
```
pramodk: Just to be sure : `total_steps` here is number of steps in single mindelay / communication interval and not whole simulation?

coreneuron/sim/fadvance_core.cpp
iomaganaris: Good point :+1:

coreneuron/sim/fadvance_core.cpp
iomaganaris: It's the `total_steps` of the whole simulation. `nrn_fixed_single_steps_minimal` and `nrn_fixed_step_group_minimal` are called by `ncs2nrn_integrate` with the total number of steps, calculated based on the `tstop` given as argument of the CLI of `CoreNeuron`. I will rename some parameters to make it clearer. :+1:

coreneuron/network/netcvode.cpp
```diff
@@ -576,26 +576,12 @@ void SelfEvent::pr(const char* s, double tt, NetCvode*) {
 }
 
 void ncs2nrn_integrate(double tstop) {
-    double ts;
-    int n = (int)((tstop - nrn_threads->_t) / dt + 1e-9);
+    int total_sim_steps = (int)((tstop - nrn_threads->_t) / dt + 1e-9);
```
ohm314: sorry for nitpicking, but could you please use C++-style casts instead? as you are changing this line already

coreneuron/sim/fadvance_core.cpp
```diff
@@ -108,11 +108,36 @@ void finalize_progress_bar() {
...
+#else
+    double updated_tstop = tstop - .5 * dt;
+    while (nrn_threads->_t < updated_tstop) {
+#endif
+        nrn_fixed_step_minimal();
+        if (stoprun) {
+            break;
+        }
+        current_steps++;
+        if (current_steps%progressbar_update_interval) {
```
ohm314:
```suggestion
        if (current_steps % progressbar_update_interval == 0) {
```
otherwise this would be true for all `current_steps` except the ones divisible by 5? I suppose you want to update the progressbar only every 5 steps?
Also, if you really prefer:
```suggestion
        if (!(current_steps % progressbar_update_interval)) {
```

coreneuron/network/netcvode.cpp
iomaganaris: Good point :+1: This was an oversight

coreneuron/sim/fadvance_core.cpp
iomaganaris: Good catch! This is what I wanted to do. I just forgot the ` == 0`.. The second suggestion is even better. Thanks a lot :+1:

orena1:
Related to this - https://github.com/BlueBrain/CoreNeuron/pull/134

PR #260, opened 2020-02-14
make nrn_multithread_job accept arguments

alkino:
It let us reduce global variable for keeping state

### DIFF
--- a/coreneuron/sim/multicore.hpp
+++ b/coreneuron/sim/multicore.hpp
@@ -157,7 +157,6 @@ extern int nrn_nthread;
 extern NrnThread* nrn_threads;
 template<typename F, typename... Args>
 void nrn_multithread_job(F&& job, Args&&... args) {
-#if defined(_OPENMP)
     int i;
 // clang-format off
     #pragma omp parallel for private(i) shared(nrn_threads, job, nrn_nthread, \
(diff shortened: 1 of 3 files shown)


coreneuron/sim/multicore.hpp
```diff
@@ -154,7 +155,24 @@ struct NrnThread : public MemoryManaged {
...
+                                           nrnmpi_myid) schedule(static, 1)
+    for (i = 0; i < nrn_nthread; ++i) {
+        job(nrn_threads + i, std::forward<Args>(args)...);
+    }
+// clang-format on
+#else
+    for (int i = 1; i < nrn_nthread; ++i) {
+        job(nrn_threads + i, std::forward<Args>(args)...);
+    }
+    job(nrn_threads + 0, std::forward<Args>(args)...);
```
pramodk: @nrnhines : can you remind me why we were doing `0`th thread last in case of serial execution? Was that for efficiency purpose? Is that still applicable? (otherwise we can simply remove the `else` part.

pramodk: @nrnhines : in case you missed notification, could you check above question?

nrnhines: The code was inherited from nrn and  I believe it was just to follow the same pattern as for pthreads where thread 0 started all the other threads and then executed its own job. I would
think you could eliminate the entire #else clause and the #if defined(_OPENMP) line
assuming the pragma is automatically skipped. Looks like all the old pthreads is entirely gone from this file.

PR #275, opened 2020-03-16
Bugfix : Support for using CoreNEURON in embedded mode with external mechanisms

### LINKED ISSUE #274: Issue while using CoreNEURON in embedded mode with NEURON using external mod files
Currently we run [in-memory test](https://github.com/BlueBrain/CoreNeuron/blob/master/tests/jenkins/neuron_direct.py) under CI (embedded mode) that uses only internal mechanisms like hh.mod. When I try to run other tests like [this](https://github.com/nrnhines/testcorenrn/blob/master/testderiv.hoc), we get:

```
 Memory (MBs) :             After mk_mech : Max 15.6641, Min 15.6641, Avg 15.6641
 num_mpi=1
 num_omp_thread=1
 Memory (MBs) :            After MPI_Init : Max 15.6914, Min 15.6914, Avg 15.6914
 Memory (MBs) :          Before nrn_setup : Max 15.7148, Min 15.7148, Avg 15.7148
error: hhderiv mechanism does not exist
```

Enabling debug log I see that external mechanisms are not registered:
 
```
 num_mpi=1
 num_omp_thread=1

register_mech na_ion 15
register_mech k_ion 16

 Duke, Yale, and the BlueBrain Project -- Copyright 1984-2019
 version id unimplemented

register_mech capacitance 3
register_mech pas 4
register_mech IClamp 7
register_mech hh 17
register_mech ExpSyn 9
register_mech NetStim 18
register_mech Exp2Syn 10
 Memory (MBs) :             After mk_mech : Max 15.6641, Min 15.6641, Avg 15.6641
 num_mpi=1
 num_omp_thread=1

 Memory (MBs) :            After MPI_Init : Max 15.6914, Min 15.6914, Avg 15.6914
 Memory (MBs) :          Before nrn_setup : Max 15.7148, Min 15.7148, Avg 15.7148
```

So the `modl_reg` is not being called in the embedded mode.

pramodk:
- modl_reg was not called when corenrn_embedded_run was used
    for running simulations with in-memory mode transfer
  - move corenrn_embedded_run() into enginemech.cpp to avoid
    link error for calling coreneuron::modl_reg

fixes #274

### DIFF
--- a/coreneuron/mechanism/mech/enginemech.cpp
+++ b/coreneuron/mechanism/mech/enginemech.cpp
@@ -1,21 +1,107 @@
-/// Corenrnmech is a wrapper lib providing a single solve_core function
-/// which initializes the solver, loads the external mechanisms and launches the simulation
+/**
+ * \file
+ * \brief Provides interface function for CoreNEURON mechanism library and NEURON
+ *
+ * libcorenrnmech is a interface library provided to building standalone executable
+ * special-core. Also, it is used by NEURON to run CoreNEURON via dlopen to execute
+ * models via in-memory transfer.
+ */
 
+#include <cstdlib>
 #include <coreneuron/engine.h>
 
-#ifdef ADDITIONAL_MECHS
 namespace coreneuron {
+
+/** Mechanism registration function
+ *
+ * If external mechanisms present then use modl_reg function generated
+ * in mod_func.cpp otherwise use empty one.
+ */
+#ifdef ADDITIONAL_MECHS
 extern void modl_reg();
+#else
+void modl_reg() {
 }
 #endif
 
+/// variables defined in coreneuron library
+extern bool nrn_have_gaps;
+extern bool nrn_use_fast_imem;
+
+}  // namespace coreneuron
+
+/** Initialize mechanisms and run simulation using CoreNEURON
+ *
+ * This is mainly used to build nrniv-core executable
+ */
 int solve_core(int argc, char** argv) {
     mk_mech_init(argc, argv);
-
-#ifdef ADDITIONAL_MECHS
-    /// Initializing additional Neurodamus mechanisms (in mod_func.c, built by mech/mod_func.c.pl)
     coreneuron::modl_reg();
-#endif
-
     return run_solve_core(argc, argv);
+}
+
+extern "C" {
+
+/// global variables from coreneuron library
+extern bool corenrn_embedded;
+extern int corenrn_embedded_nthread;
+
+/// parse arguments from neuron and prepare new one for coreneuron
+char* prepare_args(int& argc, char**& argv, int use_mpi, const char* nrn_arg);
+
+/// initialize standard mechanisms from coreneuron
+void mk_mech_init(int argc, char** argv);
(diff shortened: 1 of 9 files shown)


pramodk:
> That is libnrniv.so is agnostic as to whether external mod files will be included in the simulation. I wonder if the same can be done for CoreNEURON.

@nrnhines : I think we simply had a bug in the sense that `corenrn_embedded_run` was not calling `modl_reg` at all. 

> How about tomorrow at 8am my time?

Yes, fine for me. (I am also available now if you prefer).

coreneuron/mechanism/mech/enginemech.cpp
```diff
@@ -1,21 +1,107 @@
...
+    // initialize extra arguments built into special-core
     coreneuron::modl_reg();
-#endif
 
-    return run_solve_core(argc, argv);
+    // run simulation
+    run_solve_core(argc, argv);
+
+    // free temporary string created from prepare_args
+    free(new_arg);
```
ohm314:
```suggestion
    delete new_arg;
```
let's get rid of `free` :)

pramodk: @ohm314 : that is C pointer from `strdup` and hence should to be `free`d.

tests/jenkins/install_neuron.sh
```diff
@@ -21,7 +21,7 @@ patch_neuron() (
 set -e
 source ${JENKINS_DIR:-.}/_env_setup.sh
 
-set -x
 patch_neuron
 spack install neuron+debug@develop
+source $SPACK_ROOT/share/spack/setup-env.sh
```
iomaganaris: This could be removed, right?

tests/jenkins/_env_setup.sh
```diff
@@ -10,7 +10,10 @@ export SPACK_ROOT="${BUILD_HOME}/spack"
 export SPACK_INSTALL_PREFIX="${SPACK_INSTALL_PREFIX:-${INSTALL_HOME}}"
 export SOFTS_DIR_PATH=$SPACK_INSTALL_PREFIX  # Deprecated, but might still be reqd
 export PATH=$SPACK_ROOT/bin:/usr/bin:$PATH
```
iomaganaris: Also this I believe that it's not needed

tests/jenkins/install_neuron.sh
pramodk: When there is `no` package installed, `source $SPACK_ROOT/share/spack/setup-env.sh` doesn't do anything.
So we need to re-source after very first package installation to have MODULEPATH updated (and visible for `module av`).

tests/jenkins/_env_setup.sh
pramodk: removed

pramodk:
travis CI is taking long time but [previous build](https://travis-ci.org/github/BlueBrain/CoreNeuron/builds/663479825?utm_medium=notification&utm_source=github_status) is almost green. So I will go ahead with the merge.

Issue #279, opened 2020-03-19
Issues with nrnivmodl-core and NMODL


iomaganaris:
After the installation of `CoreNeuron` _nrnivmodl-core_ doesn't include the include directories for `nmodl`, `newton` and `Eigen`. 
The include directories of those should be added to the `nrnivmodl_core_makefile`.
What we added in this file to make `nrnivmodl-core` work with ISPC was:
```
INCLUDES = $(INCFLAGS) -I$(incdir) -I$(incdir)/coreneuron/utils/randoms -I<nmodl-install-dir>/include -I<nmodl-source-dir>/ext/eigen

ISPC_COMPILE = $(ISPC) $(ISPC_FLAGS) -I$(incdir) -I<nmodl-install-dir>/include
```

iomaganaris:
Fixed here https://github.com/BlueBrain/CoreNeuron/pull/280

PR #280, opened 2020-03-19
Fix issues while NMODL is built with sympy and analytical passes

### LINKED ISSUE #278: Issues with NMODL as submodule and use of sympy
When `CoreNeuron` is installed with `NMODL` as submodule it's not setting the `PYTHONPATH` that includes the `NMODL` python module path and its dependencies (for example _sympy_) to be able to run `NMODL`.
Branch `fix/nmodl_submodule` attempts to fix this issue. Should also be fixed if `CoreNeuron` and `NMODL` are installed as submodule of `Neuron`.
Then there is another issue when building `CoreNeuron` with `NMODL` as submodule in parallel. In this case the execution of `nrnivmodl-core` in `CoreNeuron` starts before the generation of the `NMODL` python module, so the execution of `nrnivmodl-core` fails.
Tried to tackle that on the same branch but didn't manage to fix it.

pramodk:
- set PYTHONPATH for nmodl sympy solver passes
  - make sure nmodl python module is built before compiling mod files
  - install nmodl headers to install prefix
  - enable inline pass always (--inline)
  - nrnivmodl-core inherit PYTHONPATH correctly

Fixes #278 and #279

- [x] Make sure https://github.com/BlueBrain/nmodl/pull/278 is merged
- [x] Update submodule to master of NMODL
- [x] Test if submodule under NEURON works
- [x] Test if external nmodl works under NEURON as well as CoreNEURON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,8 @@ if(CORENRN_ENABLE_NMODL)
     include(AddNmodlSubmodule)
     set(CORENRN_NMODL_BINARY ${CMAKE_BINARY_DIR}/bin/nmodl${CMAKE_EXECUTABLE_SUFFIX})
     set(CORENRN_NMODL_INCLUDE ${CMAKE_BINARY_DIR}/include)
+    set(ENV{PYTHONPATH} "$ENV{PYTHONPATH}")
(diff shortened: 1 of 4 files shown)


pramodk:
Merging this with Simulation Stack CI is bad idea but have do that today!

PR #296, opened 2020-04-24
clang-format is now at the version 10, fix regex to match version

alkino:


alkino:
I'm not authorized @pramodk

pramodk:
This doesn’t solve the issue because code itself is formatted with ancient version of clang-format.

@ohm314 : you planned to integrate hpc-coding-convention repo. If your have already started in branch, should I take a look at it? Or, can start from the beginning if it was not done.

alkino:
By the way in the meantime it fixes the current behaviour: "We want a version `> X.Y`"

PR #324, opened 2020-06-28
Use PYTHON_EXECUTABLE instead of python and update mod2c to latest master

pramodk:
* update mod2c to latest master with fixes for olfactory 3d-bulb model
* minor fix to remove hardcoded python binary name


pramodk:
cc: @iomaganaris

PR #333, opened 2020-07-07
NMODL PYTHONPATH fixes

iomaganaris:
- Fixed NMODL PYTHONPATH in another place in CMake
- Added "sympy --analytic" flag in CoreNEURON build in CI to test if the proper PYTHONPATH of NMODL is set on CoreNEURON

### DIFF
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -59,7 +59,7 @@ include(NmodlHelper)
 # Command for MOD to CPP conversion
 # =============================================================================
 set(CORENRN_NMODL_COMMAND env "MODLUNIT=${NMODL_UNITS_FILE}"
-                              "PYTHONPATH=$ENV{PYTHONPATH}:${CMAKE_BINARY_DIR}/lib/python"
+                              "PYTHONPATH=${CORENRN_NMODL_PYTHONPATH}"
                               ${CORENRN_NMODL_BINARY})
 
 if(${CORENRN_ENABLE_ISPC})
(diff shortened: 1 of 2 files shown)


coreneuron/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ include(NmodlHelper)
 # Command for MOD to CPP conversion
 # =============================================================================
 set(CORENRN_NMODL_COMMAND env "MODLUNIT=${NMODL_UNITS_FILE}"
-                              "PYTHONPATH=$ENV{PYTHONPATH}:${CMAKE_BINARY_DIR}/lib/python"
+                              "PYTHONPATH=${CORENRN_NMODL_PYTHONPATH}"
```
pramodk, line 62: I thought `$ENV{PYTHONPATH}:` is required when nrnivmodl-core after installation.

iomaganaris: I checked and the only place where `PYTHONPATH` is set is in
https://github.com/BlueBrain/CoreNeuron/blob/f0daaa460edd3251894ece5a1d8299288cb38ce9/CMakeLists.txt#L204
and 
https://github.com/BlueBrain/CoreNeuron/blob/f0daaa460edd3251894ece5a1d8299288cb38ce9/CMakeLists.txt#L210
Then `CORENRN_NMODL_PYTHONPATH` includes the `PYTHONPATH` set above
https://github.com/BlueBrain/CoreNeuron/blob/f0daaa460edd3251894ece5a1d8299288cb38ce9/CMakeLists.txt#L212
so I thought it's fine.
`mod2c` doesn't need `PYTHONPATH` I believe, so adding an extra `PYTHONPATH` there is not needed, right?

PR #335, opened 2020-07-12
Add CMake option to set shared or static build type

pramodk:


CMakeLists.txt
```diff
@@ -117,6 +118,12 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "PGI")
   add_definitions(-DEIGEN_DONT_VECTORIZE=1)
 endif()
 
+if(CORENRN_ENABLE_SHARED)
+  set(COMPILE_LIBRARY_TYPE "SHARED")
+else()
+  set(COMPILE_LIBRARY_TYPE "STATIC")
```
iomaganaris, line 124: Should `COMPILE_LIBRARY_TYPE` be set to `STATIC` by default for `GPU` build?

pramodk: It was initially done in CMake/OpenAccHelper.cmake. But as you said, better to do it here. Done.

iomaganaris: Sorry. I missed that it was set there and thought that there was an issue with this since the GPU test in the CI failed.
All good now!

PR #349, opened 2020-07-27
Fixes for multi-execution in-memory transfer with NEURON

pramodk:
- global ion maps were deleted at the end of every run
  - in case of in-memory execution mechanisms are registered
    only once and hence ions map is created only one
  - this is why ions shouldnt not be deleted if we are doing
    embedded run with neuron

@cattabiani  : this will also fix the issue where you were seeing segfault with `40%` of mod files from v5 circuit.


pramodk:
I will go ahead with merge now!

PR #354, opened 2020-08-03
Return all spikes vectors to NEURON

alexsavulescu:
In link with: https://github.com/neuronsimulator/nrn/pull/692 

Should we add an extra parameter to the NEURON callback in order to indicate if writing `output.dat` should be skipped? Or is something we'd like to have at a global level, say a CLI option?

### DIFF
--- a/coreneuron/io/output_spikes.cpp
+++ b/coreneuron/io/output_spikes.cpp
@@ -47,10 +47,16 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #include "bbp/sonata/reports.h"
 #endif  // ENABLE_SONATA_REPORTS
 
-static void all_spikes_return(std::vector<double>& spiketvec, std::vector<int>& spikegidvec) {
-    if (nrn2core_all_spike_vectors_return_) {
-        (*nrn2core_all_spike_vectors_return_)(spiketvec, spikegidvec);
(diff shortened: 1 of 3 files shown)


nrnhines:
Could generate the out.dat if the callback is nullptr or if the callback does not copy the spikes (and so returns 0(?))

pramodk:
> Could generate the out.dat if the callback is nullptr or if the callback does not copy the spikes (and so returns 0(?))

I am thinking similar : if coreneuron is running in embedded mode then don't generate the spike output file.

coreneuron/io/output_spikes.cpp
```diff
@@ -226,6 +236,10 @@ void output_spikes_parallel(const char* outpath, const std::string& population_n
 #endif
 
 void output_spikes_serial(const char* outpath) {
+
+    // transfer spikes to NRN
+    all_spikes_return(spikevec_time, spikevec_gid);
+
```
pramodk, line 242: `nitpick` : instead of above two callbacks, I think we can add this function call in the beginning of : 

```
268 void output_spikes(const char* outpath, const std::string& population_name) {
```

alexsavulescu: Indeed, thank you

PR #355, opened 2020-08-04
FOR_NETCON

nrnhines:
Framework in place along with comment describing the data structures.

This interacts with BlueBrain/mod2c#43 and NeuronSimulator/nrn#695 and they need to be built together.


pramodk:
@nrnhines : the CI failures are caused by this line : https://github.com/BlueBrain/CoreNeuron/blob/f2e0f01549e98c23bcc2b6886ab37f53164e01dd/coreneuron/nrnoc/md1redef.h#L62

nrnhines:
This has been tested with a mod file with a printf statement that prints all the external events arriving at the NET_RECEIVE block and processing the stdout with
```
$ cat test/coreneuron/test.sh
#!/bin/bash

nrniv -python test_fornetcon.py | sed -n '/zzz/s/^.*zzz //p'|uniq|sed '1d'> temp
sed -n '1,/INITIAL/p' < temp > temp1
sed '1,/INITIAL/d' < temp > temp2
sortspike temp1 temp
mv temp temp1
sortspike temp2 temp
mv temp temp2
diff temp1 temp2
```
(The ```uniq``` part of the filter converts the 10 lines from the INITIAL block from ```printf("zzz INITIAL\n")``` to just one line in order to conveniently separate the output of NEURON from the output of CoreNEURON).

Anyway, a better test would be for CoreNEURON to send back the weights to NEURON at the end of the run (just like the spikes) and then it will be easy to integrate the test into nrn/test/coreneuron for automatic testing.

Til then, I think it would be best to go ahead and merge this into the master as CoreNEURON is undergoing major refactoring that causes conflict with this branch.

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+    fn = ml->pdata + (instance*sz + fnslot);
+  }else if (layout == 0) { /* SoA */
+    int padded_cnt = nrn_soa_padded_size(ml->nodecount, layout);
+    fn = ml->pdata + (fnslot*padded_cnt + instance);
+  }
+  return fn;
+}
+
+#if 0 // for debugging
+static void print_fornetcon_slot(const char* s, std::map<int, int> type_to_slot, NrnThread& nt) {
```
alexsavulescu, line 94: Looks `print_fornetcon_slot` is not referenced. 

`std::map<int, int> type_to_slot` should be passed as `const std::map<int, int>& type_to_slot` to avoid copy

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+    int nodecount = nt._ml_list[mtype]->nodecount;
+    std::cout << "nodecount " << nodecount << std::endl;
+    for (int i=0; i < nodecount; ++i) {
+      int* fn = fornetcon_slot(mtype, i, fnslot, nt);
+      std::cout << "i=" << i << " *fn=" << *fn << std::endl;
+    }
+  }
+}
+
+static void print_vector(const char* s, std::vector<size_t>& v) {
```
alexsavulescu, line 109: Looks like `print_vector`  is not referenced

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  // Mechanism types in use that have FOR_NETCONS statements
+  // Nice to have the dparam fornetcon slot as well so use map
+  // instead of set
+  std::map<int, int> type_to_slot;
+  for (int i = 0; i < nrn_fornetcon_cnt_; ++i) {
+    int type = nrn_fornetcon_type_[i];
+    Memb_list* ml = nt._ml_list[type];
+    if (ml && ml->nodecount) {
+      type_to_slot.insert(std::pair<int, int>(type, nrn_fornetcon_index_[i]));
```
alexsavulescu, line 129: I'd use operator[] over insert here (in this case fore readability): 
   type_to_slot[type] = nrn_fornetcon_index_[i];

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+    int padded_cnt = nrn_soa_padded_size(ml->nodecount, layout);
+    fn = ml->pdata + (fnslot*padded_cnt + instance);
+  }
+  return fn;
+}
+
+#if 0 // for debugging
+static void print_fornetcon_slot(const char* s, std::map<int, int> type_to_slot, NrnThread& nt) {
+  std::cout << s << std::endl;
+  for (auto it = type_to_slot.begin(); it != type_to_slot.end(); ++it) {
```
alexsavulescu, line 96: nitpick: 
  for (const auto& kv : type_to_slot) {
  }

then use : `kv.first` and `kv.second`

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return;
+  }
+
+  // How many NetCons (weight groups) are involved.
+  // Also count how many weight groups for each target instance.
+  // For the latter we can count in the dparam fornetcon slot.
+
+  // zero the dparam fornetcon slot for counting and count number of slots.
+  size_t n_perm_indices = 0;
+  for (auto it = type_to_slot.begin(); it != type_to_slot.end(); ++it) {
```
alexsavulescu, line 142: nitpick: 
  for (const auto& kv : type_to_slot) {
  }

then use : `kv.first` and `kv.second`

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+  // at n-1 is x[n] - x[n-1] and number for first is x[0] = 0.
+  nt._fornetcon_perm_indices.resize(n_perm_indices + 1);
+  nt._fornetcon_weight_perm.resize(n_weight_perm);
+
+  // From dparam fornetcon slots, compute displacement vector, and
+  // set the dparam fornetcon slot to the index of the displacement vector
+  // to allow later filling the _fornetcon_weight_perm.
+  size_t i_perm_indices = 0;
+  nt._fornetcon_perm_indices[0] = 0;
+  for (auto it = type_to_slot.begin(); it != type_to_slot.end(); ++it) {
```
alexsavulescu, line 177: nitpick: 
  for (const auto& kv : type_to_slot) {
  }

then use : `kv.first` and `kv.second`

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+      int* fn = fornetcon_slot(mtype, i_instance, search->second, nt);
+      size_t nc_w_index = size_t(nc.u.weight_index_);
+      nt._fornetcon_weight_perm[size_t(*fn)] = nc_w_index;
+      *fn += 1; // next item conceptually adjacent
+    }
+  }
+
+  // Put back the proper values into the dparam fornetcon slot
+  i_perm_indices = 0;
+  for (auto it = type_to_slot.begin(); it != type_to_slot.end(); ++it) {
```
alexsavulescu, line 208: nitpick: 
  for (const auto& kv : type_to_slot) {
  }

then use : `kv.first` and `kv.second`

coreneuron/io/setup_fornetcon.cpp
```diff
@@ -0,0 +1,220 @@
...
+   NrnThread._fornetcon_perm_indices is an index vector into the
+   NrnThread._fornetcon_weight_perm to the first of the list of NetCon weights
+   that target the instance. The index of _fornetcon_perm_indices
+   containing this first in the list is stored in the mechanism instances
+   dparam at the dparam's semantic fornetcon slot. (Note that the next index
+   points to the first index of the next target instance.)
+
+**/
+
+static int* fornetcon_slot(int mtype, int instance, int fnslot, NrnThread& nt) {
```
alexsavulescu, line 79: consider adding `const` to function parameters

CMakeLists.txt
```diff
@@ -315,7 +315,7 @@ endif()
 # =============================================================================
 if(DOXYGEN_FOUND AND SPHINX_FOUND)
   add_custom_target(
-          docs
+          corenrn_docs
```
alexsavulescu, line 318: Need to update this in `.travis.yml`, `docs/README.md` and the following line in `CMakeLists.txt`: 
```
message(STATUS "make docs           | Build full docs. Calls targets: doxygen, sphinx")
```

coreneuron/apps/main1.cpp
```diff
@@ -601,6 +601,16 @@ extern "C" int run_solve_core(int argc, char** argv) {
         output_spikes(output_dir.c_str(), spikes_population_name);
     }
 
+    // copy weights back to NEURON NetCon
+    if (nrn2core_all_weights_return_) {
+        std::vector<double*> weights;
```
alexsavulescu, line 606: Simple optimization:
```cpp
std::vector<double*> weights(nrn_nthread, NULL);
for (int i=0; i < nrn_nthread; ++i) {
          weights[i]=nrn_threads[i].weights;
}
```

coreneuron/io/setup_fornetcon.cpp
nrnhines: Nice idiom. Thanks!

CMakeLists.txt
nrnhines: Thanks.

coreneuron/apps/main1.cpp
nrnhines: I agree. Thanks.  I have been explicitly casting int to std::size_t for indexing into std::vector. Is that uselessly verbose?

.travis.yml
```diff
@@ -183,7 +183,7 @@ after_success:
   - if [ "$BUILD_DOCUMENTATION" = "ON" ] && [ "$TRAVIS_OS_NAME" == "linux" ]; then
       echo "------- Build Doxygen Documentation -------";
       pushd $TRAVIS_BUILD_DIR/build;
-      make docs;
+      make corenrn_docs;
```
pramodk, line 186: @alexsavulescu @nrnhines : should this be changed to avoid conflict? `docs` is much easier to remember / standard for any project.

nrnhines: I greatly prefer leaving it as 'docs' and reverting this. The reason for it is that cmake raised an error when it was a submodule of nrn. Because nrn already has a target called 'docs'?

alexsavulescu: https://cmake.org/cmake/help/v3.0/policy/CMP0002.html

Since `Logical target names must be globally unique.` I guess it would make sense to have it named differently.
There is a setting that would allow duplicates:
```
Custom targets must simply have globally unique names (unless one uses the global property ALLOW_DUPLICATE_CUSTOM_TARGETS with a Makefiles generator).
```
, but I'm unsure of the implications; maybe it would be best to have a clear distinction nevertheless.

nrnhines: I wonder if https://cmake.org/cmake/help/latest/module/ExternalProject.html
may work around this in the nrn cmake.

pramodk: @alexsavulescu : in some external projects, which are typically used as submodule in other project, what I have seen is that the generic targets like `docs` are disabled if project is being built as submodule i.e.

```
if (NOT IS_USED_AS_SUBMODULE)
    add_custom_target(docs ...)
endif()
```

The reason is that, if project X is being used as submodule under Y then forcing project X to use different targets than project Y is annoying IMO. In this case we can do it because both are managed by us but not the case for other projects.

> I wonder if https://cmake.org/cmake/help/latest/module/ExternalProject.html

External project is much more pain and we are using it for MOD2c : https://github.com/BlueBrain/CoreNeuron/blob/b1ea553f19673ae0c107ed4c01c23c4484ef00c0/CMake/AddMod2cSubmodule.cmake#L37. I am going to remove that soon! :)

pramodk: Note that we are using this pattern already:

https://github.com/BlueBrain/nmodl/blob/f90f671e03fd537540de359e52e21852a335f2b1/CMakeLists.txt#L166
https://github.com/BlueBrain/CoreNeuron/blob/b1ea553f19673ae0c107ed4c01c23c4484ef00c0/CMakeLists.txt#L101

nrnhines: Hmm... I wonder.... If I install sphinx and doxygen on this Mac, will I have the same issues when coreneuron is a submodule.  I expect so.

pramodk: I didn't understand. Do you mean for `docs` target if NEURON as well as CoreNEURON have the same target?
What I saw saying is that if coreneuron is submodule under NEURON then `docs` target of CoreNEURON won't be added in CMake.

alexsavulescu: @pramodk `docs` is calling two other targets (`sphinx` and `doxygen`) that are created only if we have the respective tools installed. Those would clash as well since we have the same setup in `nrn` and `corenrn`

alexsavulescu: Ok added the extra check. @nrnhines it should go fine with `nrn` now. Thanks @pramodk

alexsavulescu:
@pramodk we should see about stopping running jenkins jobs when PR gets updated (save some time)

.travis.yml
nrnhines: I'm a little confused about coreneuron as submodule docs. Wouldn't it be the case that the super module would want to install those docs. And similarly for neuron. If nrn is a submodule then docs could change to nrn_docs

pramodk: We can discuss how would like to see documentation of different projects : 1) whether all should be integrated into single tree or 2) there should be link to different projects doscs.

Obviously 1) has advantage of being able to search under single tree. But today this might be too early to do so? 

@alexsavulescu : another way to achieve previous use case would be:

```
if (CORENEURON_AS_SUBPROJECT)
  set(CORENEURON_TARGET_PREFIX "corenrn")
endif()

if(DOXYGEN_FOUND AND SPHINX_FOUND)
  add_custom_target(
    ${CORENEURON_TARGET_PREFIX}docs
   ...
   )
```
@nrnhines  : we can merge this PR as it is. But if we still thing we should re-enable docs with different target names then we can use above approach.

nrnhines: I'm satisfied

alexsavulescu: Option 1 is ideal and definitely the long term goal.

At this point I'd avoid doing more work here because we can end up with clashes in the `build/docs` folder as well, not just targets. My vision is as soon as we get production-ready tools to incorporate doxygen and jupyter notebooks directly into sphinx, we will have a singular target called `docs` in neuron.

pramodk: Agree ! all good then!

pramodk:
> @pramodk we should see about stopping running jenkins jobs when PR gets updated (save some time)

thats true! Seems like there are [some solutions](https://stackoverflow.com/questions/40760716/jenkins-abort-running-build-if-new-one-is-started).

PR #357, opened 2020-08-08
Update voltage vector from GPU to CPU if trajectories are recorded

pramodk:
- NEURON can request various trajectories from CoreNEURON
  - In this implementation we are just updating voltage vector
    and satisfy the use case of olfactory 3d-bulb model
  - We will need more fine grain approach (for future PR)

### DIFF
--- a/coreneuron/sim/fadvance_core.cpp
+++ b/coreneuron/sim/fadvance_core.cpp
@@ -286,6 +286,11 @@ void nrncore2nrn_send_values(NrnThread* nth) {
 
     TrajectoryRequests* tr = nth->trajec_requests;
     if (tr) {
+        // \todo Check if user has requested voltages for this NrnThread object.
+        //       Currently we are updating voltages if there is any trajectory
+        //       requested by NEURON.
+        update_voltage_from_gpu(nth); 
+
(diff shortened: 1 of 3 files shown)


coreneuron/sim/fadvance_core.cpp
```diff
@@ -286,6 +286,11 @@ void nrncore2nrn_send_values(NrnThread* nth) {
 
     TrajectoryRequests* tr = nth->trajec_requests;
     if (tr) {
+        // \todo Check if user has requested voltages for this NrnThread object.
+        //       Currently we are updating voltages if there is any trajectory
+        //       requested by NEURON.
+        update_voltage_from_gpu(nth);
```
nrnhines, line 292: Although it is true that most requests are for voltage. The present non-gpu implementation allows any RANGE variable and t.  Is the entire data array too much (and isn't that already done at the end of a run anyway?)

pramodk: yes, updating data from gpu to cpu could be too heavy. For efficiency reason we might need to accumulate buffers directly on GPU (based on memory requirements). This advanced implementation need to be done separately.

With this PR, I am finishing the functionality required for bulb-3d model.

PR #360, opened 2020-08-10
Remove extracon and multiple, fix #291

### LINKED ISSUE #291: Remove code related to model duplication factor  (-z,--multiple, -x,--extracon command line argument)
In the old days we wanted to duplicate input model by certain factor to perform scaling studies. Now models are quite bigger. We have ability to generate arbitrary size models to perform scaling studies.

In order to reduce the code complexity, I would suggest to remove the code related to model duplication from coreneuron.

@nrnhines : would that be ok? I haven't used this for few years already and don't see any use cases now days.

alkino:


pramodk:
Ahh..`Phase2::handle_extracon` contain logic of setting up weight vectors and that was causing an issue. Ok!

coreneuron/nrniv/nrniv_decl.h
```diff
@@ -65,7 +65,6 @@ extern void nrn_setup(const char* filesdat,
 extern double* stdindex2ptr(int mtype, int index, NrnThread&);
 extern void delete_trajectory_requests(NrnThread&);
 extern int nrn_setup_multiple;
```
alkino, line 67: remove

pramodk:
Didn't;t re

PR #365, opened 2020-08-11
Applying consistent source formatting with hpc-coding-convention

pramodk:
fix order of nrnmpi.h include to build issue
Additional changes with clang-format v9


pramodk:
@alexsavulescu @ohm314 @alkino  : could you skim through this formatting changes and see the result of the code style applied? We have room for **minor** adjustments if we see specific style is not suitable for coreneuron.

You can locally try this branch as:

```
git checkout pramodk/hpc-clang-format

# make sure clang-format v9 (or v8) is $PATH
 cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install -DCORENRN_CLANG_FORMAT=ON

# below will be created, you can modify it 
cat ../.clang-format

# first build
make -j

# update clang-format
make clang-format

# update config
vim ../.clang-format

# re-apply style
make clang-format
```

In this PR I haven't done refactoring but just applied code formatting changes + some minor fixes.

- [ ] One non-deterministic behaviour I have seen is that sometime below diff appears and some time not i.e. comment / pragma is not consistently formatted : 

```
$ make clang-format
→ git diff
diff --git a/CMake/hpc-coding-conventions b/CMake/hpc-coding-conventions
--- a/CMake/hpc-coding-conventions
+++ b/CMake/hpc-coding-conventions
@@ -1 +1 @@
-Subproject commit de25d00e911ce73f7db1ede349562efb37263125
+Subproject commit de25d00e911ce73f7db1ede349562efb37263125-dirty
diff --git a/coreneuron/apps/main1.cpp b/coreneuron/apps/main1.cpp
index 372c469c..57bfe45d 100644
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -508,7 +508,7 @@ extern "C" int run_solve_core(int argc, char** argv) {
     bool compute_gpu = corenrn_param.gpu;
     bool skip_mpi_finalize = corenrn_param.skip_mpi_finalize;

-    // clang-format off
+// clang-format off
     #pragma acc data copyin(celsius, secondorder) if (compute_gpu)
     // clang-format on
     {
diff --git a/coreneuron/gpu/nrn_acc_manager.cpp b/coreneuron/gpu/nrn_acc_manager.cpp
index 5545992a..44032a2e 100644
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -770,9 +770,9 @@ void update_voltage_from_gpu(NrnThread* nt) {
 void update_matrix_from_gpu(NrnThread* _nt) {
 #ifdef _OPENACC
     if (_nt->compute_gpu && (_nt->end > 0)) {
-        /* before copying, make sure all computations in the stream are completed */
+/* before copying, make sure all computations in the stream are completed */
....
```

coreneuron/io/global_vars.cpp
```diff
@@ -100,7 +101,7 @@ void set_globals(const char* path, bool cli_global_seed, int cli_global_seed_val
                 }
                 it = n2v->find(name);
                 if (it != n2v->end()) {
-                    nrn_assert(it->second.first == (size_t)n);
+                    nrn_assert(it->second.first == (size_t) n);
```
alexsavulescu, line 104: I kind of like to have casts glued to what we are casting from, but this is a matter of personal taste.

pramodk: I am ok with such change. Is there such option in clang-format? If so we can update it.

alexsavulescu: `SpaceAfterCStyleCast: false`

ohm314: Let's not discuss style here - since the formatting comes from hpc coding conventions this discussion could be discussed there. We certainly wouldn't want to divert from the format file there, right?

pramodk:
superseded by #456

Issue #369, opened 2020-08-13
Multiple invocations of psolve when using direct memory interface


iraikov:
Hello,
Thanks again for your work on direct memory integration with NEURON.  One of my use cases involves repeated invocations of psolve as part of an optimization loop. This works in standard NEURON, but when attempting to do the same in CoreNEURON via the  direct memory interface with NEURON, either psolve returns immediately and does not conduct the simulation, or causes a segmentation fault. For example, adding a second call to psolve in the test_spikes.py script causes a segmentation fault. Should it be expected the repeated calls to psolve in the same context is supported in CoreNEURON?  And if so, is there a specific initialization that has to be performed in between calls to psolve? My current code only invokes finitialize before the next psolve invocation. Thanks a lot for your help.

pramodk:
@iraikov : we have [here very simple test](https://github.com/BlueBrain/CoreNeuron/blob/65dffb235d4090b715546d04b1504d87dd10aacf/tests/jenkins/neuron_direct.py) to run multiple invocations but it may not be capturing all model aspects and there might be a bug.

If it's possible to provide a test to reproduce the issue then it would help us to debug and fix it faster. 

@nrnhines : Did you encounter similar issue recently?

iraikov:
Hi @pramodk thanks, this test script causes a segmentation fault on my computer as well. Perhaps there is something peculiar to my setup (MPICH+gcc on Debian). Let me try on Frontera with intel to see if the same issue occurs.

pramodk:
Interesting! I can tomorrow try it on Ubuntu/Debian system. Could you tell me your cmake  arguments used for neuron /  coreneuron build?

iraikov:
Hi @pramodk the test script worked with Intel MPI on Frontera. Another difference was that I built without Interviews. The compilation options for MPICH+gcc were:

`cmake  ../nrn -DCMAKE_INSTALL_PREFIX=$HOME/bin/nrnpython3  -DNRN_ENABLE_CORENEURON=ON -DPYTHON_EXECUTABLE=`which python3`  -DNRN_ENABLE_MPI=ON -DCMAKE_CXX_FLAGS="-D MPICH_SKIP_MPICXX=1" -DNRN_ENABLE_RX3D=ON`

nrnhines:
Fixed a segmentation fault on ubuntu 18.4, gcc7.5.0 with "here very simple test". The #377 will be merged soon.
[...]

PR #370, opened 2020-08-14
Use coreneuron module API from NEURON instead of pc.nrncore_run()

pramodk:


pramodk:
This is to also test the issue mentioned in #369.

PR #372, opened 2020-08-15
Avoid linking to C++ MPI libraries

pramodk:
- MPI C++ bindings are removed from MPI
  - avoid linking mpi c++ library for simplicity


coreneuron/CMakeLists.txt
```diff
@@ -207,8 +207,7 @@ add_dependencies(coreneuron scopmath)
 include_directories(${CORENEURON_PROJECT_SOURCE_DIR})
 add_executable(nrniv-core "apps/coreneuron.cpp")
 
-target_link_libraries(nrniv-core corenrnmech ${reportinglib_LIBRARY} ${sonatareport_LIBRARY}
-                      ${MPI_C_LIBRARIES} ${MPI_CXX_LIBRARIES})
+target_link_libraries(nrniv-core corenrnmech ${reportinglib_LIBRARY} ${sonatareport_LIBRARY})
```
nrnhines, line 210: I understand about eliminating the ```{MPI_CXX_LIBRARIES}``` but I don't understand why ```${MPI_C_LIBRARIES}``` was removed

CMakeLists.txt
```diff
@@ -159,6 +159,10 @@ endif()
 if(CORENRN_ENABLE_MPI)
   find_package(MPI REQUIRED)
   add_definitions("-DNRNMPI=1")
+  # avoid linking to C++ bindings
+  add_definitions("-DMPI_NO_CPPBIND=1")
```
nrnhines, line 163: Just for my information, who uses these? ```grep -r MPI_NO_CPPBIND *``` turns up nothing in my nrn repository that has coreneuron as a submodule.

coreneuron/CMakeLists.txt
pramodk: It's ok to keep but `libcoreneuron` target already links to `MPI_C_LIBRARIES` and hence it was not needed.

CMakeLists.txt
pramodk: Actually it's in MPI library header. For example, HPE -MPI on BB5 has:

```
# /gpfs/bbp.cscs.ch/ssd/apps/hpc/jenkins/deploy/tools/2020-02-01/linux-rhel7-x86_64/gcc-8.3.0/hpe-mpi-2.21-7pbszh6v5u/include/mpi.h
 * mpio.h must be included after mpi.h
 */
#include "mpio.h"
#include "mpi_t.h"

#if defined(__cplusplus)
}
#if !defined(MPI_NO_CPPBIND)
#include "mpi++.h"
#endif
#endif

```

and that we want to avoid.

ohm314:
please retest

PR #373, opened 2020-08-15
GPU build improvements for NEURON integration

pramodk:
*   Improvemenets for GPU build integration with NEURON
      - create property CORENEURON_LIB_LINK_FLAGS with necessary LDFLAGS
        to link coreneuron with neuron via nrnivmodl
      - suppress additional PGI warnings which are not necessary to emit
        with old mod2c
      - minor change in nrn_setup.hpp : functions declarated as static
        in the header result into warning as definitions are in .cpp file


pramodk:
Also include #372, should be merged after that.

PR #377, opened 2020-08-16
Unitialized pointer. Later segmentation fault after freeing.

### LINKED ISSUE #369: Multiple invocations of psolve when using direct memory interface

nrnhines:
Fixes #369.

There are several other issues in this context that  I would like to discuss.

### DIFF
--- a/coreneuron/permute/cellorder.cpp
+++ b/coreneuron/permute/cellorder.cpp
@@ -280,7 +280,8 @@ int* interleave_order(int ith, int ncell, int nnode, int* parent) {
         }
     }
 
-    int nwarp, nstride, *stride, *firstnode, *lastnode, *cellsize, *stridedispl;
+    int nwarp = 0, nstride=0, *stride=nullptr, *firstnode=nullptr;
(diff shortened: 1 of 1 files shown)


pramodk:
Ahh, was that uninitialised pointée was being treated as non-null and then was freed?

nrnhines:
Yes.  And fixing this eliminated all valgrind errors. However there are 3 related issues I'd like to discuss some time.

PR #378, opened 2020-08-16
Minor fixes

nrnhines:
I separated these into small independent commits. Especially solicit your opinion on not implementing assignment and copy constructors. Also, if those are useful to have, is it a good idea to specifically have ```delete``` correspond to ```new`` and ```free``` corespond to ```malloc```.

### DIFF
--- a/coreneuron/permute/cellorder.hpp
+++ b/coreneuron/permute/cellorder.hpp
@@ -13,8 +13,8 @@ extern void solve_interleaved(int ith);
 class InterleaveInfo {
   public:
     InterleaveInfo() = default;
-    InterleaveInfo(const InterleaveInfo&) = delete;
-    InterleaveInfo& operator=(const InterleaveInfo&) = delete;
+    InterleaveInfo(const InterleaveInfo&);
+    InterleaveInfo& operator=(const InterleaveInfo&);
     virtual ~InterleaveInfo();
     int nwarp = 0;  // used only by interleave2
     int nstride = 0;
@@ -30,6 +30,9 @@ class InterleaveInfo {
     size_t* idle = nullptr;
     size_t* cache_access = nullptr;
     size_t* child_race = nullptr;
+
+  private:
+    void swap(InterleaveInfo& info);
 };
 
 // interleaved from cellorder2.cpp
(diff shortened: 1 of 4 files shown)


pramodk:
@nrnhines : assignment is being used below:

```
/jenkins/07/workspace/oss.coreneuron.pipeline/tests/unit/interleave_info/check_constructors.cpp(70): error: function "coreneuron::InterleaveInfo::operator=(const coreneuron::InterleaveInfo &)" (declared at line 17 of "/jenkins/07/workspace/oss.coreneuron.pipeline/coreneuron/permute/cellorder.hpp") cannot be referenced -- it is a deleted function
      info3 = info1;
```

alexsavulescu:
> Especially solicit your opinion on not implementing assignment and copy  constructors.

It depends on usage. What would be the case to copy or assign `InterleaveInfo` data? If `InterleaveInfo` is doing only resource management, then we wouldn't normally need them. 

> Also, if those are useful to have, is it a good idea to specifically have delete correspond to new`` and freecorespond tomalloc```.
`new/malloc` and  `delete/free` can be interleaved, but the deallocation shall always match the allocation otherwise things can get problematic.

coreneuron/permute/cellorder.hpp
```diff
@@ -12,8 +13,8 @@ extern void solve_interleaved(int ith);
 class InterleaveInfo {
   public:
     InterleaveInfo() = default;
-    InterleaveInfo(const InterleaveInfo&);
-    InterleaveInfo& operator=(const InterleaveInfo&);
+    InterleaveInfo(const InterleaveInfo&) = delete;
+    InterleaveInfo& operator=(const InterleaveInfo&) = delete;
     virtual ~InterleaveInfo();
```
alexsavulescu, line 18: As far as I can tell there are no derived classes here. No need to have `virtual` here (which creates the extra `vtable`)

nrnhines:
> assignment is being used

Ok. So the choice is to revert 7dc3acc or to change the test. I don't have much of an opinion either way. Which would you prefer?
> No need to have virtual 

I don't expect to subclass in the future...  Just a habit. Will get rid of it.

pramodk:
> So the choice is to revert 7dc3acc or to change the test. I don't have much of an opinion either way. Which would you prefer?

yeah, reverting is fine!

nrnhines:
Put back the assignment and copy constructors. Is it still worthwhile to make the destructor non-virtual?

pramodk:
Yeah, virtual can be removed here.

PR #380, opened 2020-08-18
Clear spike vectors after writing or transferring.

nrnhines:
Fixes Ivan's last problem mentioned in #369.

### DIFF
--- a/coreneuron/io/output_spikes.cpp
+++ b/coreneuron/io/output_spikes.cpp
@@ -203,7 +203,7 @@ void output_spikes_parallel(const char* outpath, const std::string& population_n
     char spike_entry[SPIKE_RECORD_LEN];
     unsigned spike_data_offset = 0;
     for (unsigned i = 0; i < num_spikes; i++) {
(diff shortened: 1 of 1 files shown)


nrnhines:
Oops. I inadvertently committed as well my private change to the spike output format.

nrnhines:
The errors on the default pipeline are
```
[ERROR] No output files. Test failed!
```
Not clear to me why this change caused that.

pramodk:
> Not clear to me why this change caused that.

I believe `{ }` were missing to the if statement. Fixes that now.

coreneuron/io/output_spikes.cpp
```diff
@@ -268,8 +268,10 @@ void output_spikes_serial(const char* outpath) {
 
 void output_spikes(const char* outpath, const std::string& population_name) {
     // try to transfer spikes to NEURON. If successfull, don't write out.dat
-    if (all_spikes_return(spikevec_time, spikevec_gid))
+    if (all_spikes_return(spikevec_time, spikevec_gid)) {
+        clear_spike_vectors();
         return;
+    }
```
nrnhines, line 274: Never fails to surprise me that something can be right in front of my nose and I can't see it til someone says, "see!"

PR #382, opened 2020-08-18
Copies to NEURON the voltage, i_membrane_, and mechanism data.

nrnhines:
Framework in place.

Interacts with neuronsimulator/nrn#717

### DIFF
--- a/coreneuron/io/core2nrn_data_return.cpp
+++ b/coreneuron/io/core2nrn_data_return.cpp
@@ -37,6 +37,9 @@ size_t (*nrn2core_type_return_)(int type, int tid,
 
 namespace coreneuron {
 
+static void inverse_permute_copy(size_t n, double* permuted_src, double* dest,
+                                 int* permute);
+
 void core2nrn_data_return() {
   if (!nrn2core_type_return_) {
     return;
@@ -46,10 +49,23 @@ void core2nrn_data_return() {
     double* data = nullptr;
     double** mdata = nullptr;
     NrnThread& nt = nrn_threads[tid];
-    if (nt.end) {
+    if (nt.end) { // transfer voltage and possibly i_membrane_
       n = (*nrn2core_type_return_)(voltage, tid, data, mdata);
-      assert(n == size_t(nt.end));
+      assert(n == size_t(nt.end) && data);
+      inverse_permute_copy(n, nt._actual_v, data, nt._permute);
+
+      if (nt.nrn_fast_imem) {
+        n = (*nrn2core_type_return_)(i_membrane_, tid, data, mdata);
+        assert(n == size_t(nt.end) && data);
+        inverse_permute_copy(n, nt.nrn_fast_imem->nrn_sav_rhs, data, nt._permute);
+      }
     }
+  }
+}
+
+void inverse_permute_copy(size_t n, double* permuted_src, double* dest, int* permute) {
+  for (size_t i = 0; i < n; ++i) {
+    dest[permute[i]] = permuted_src[i];
   }
 }
 
--- a/coreneuron/io/core2nrn_data_return.cpp
+++ b/coreneuron/io/core2nrn_data_return.cpp
@@ -38,7 +38,42 @@ size_t (*nrn2core_type_return_)(int type, int tid,
 namespace coreneuron {
 
 static void inverse_permute_copy(size_t n, double* permuted_src, double* dest,
-                                 int* permute);
+                                 int* permute)
+{
+  if (permute) {
+    for (size_t i = 0; i < n; ++i) {
+      dest[permute[i]] = permuted_src[i];
+    }
+  }else{
+    std::copy(permuted_src, permuted_src + n, dest);
+  }
+}
+
+static void soa2aos_inverse_permute_copy(size_t n, int sz, int stride,
+  double* src, double** dest, int* permute)
+{
+  // src is soa and permuted. dest is n pointers to sz doubles (aos).
+  for (size_t instance = 0; instance < n; ++instance) {
+    double* d = dest[permute[instance]];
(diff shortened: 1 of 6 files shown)


coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -0,0 +1,174 @@
...
+size_t (*nrn2core_type_return_)(int type, int tid,
+  double*& data, double**& mdata);
+
+namespace coreneuron {
+
+/** @brief permuted array copied to unpermuted array
+ *  If permute is NULL then just a copy
+ */
+static void inverse_permute_copy(size_t n, double* permuted_src, double* dest,
+                                 int* permute)
```
pramodk, line 51: Just a remark for the overall code : we typically use `const` for  pointer parameters for the reasons [discussed here](https://stackoverflow.com/questions/7715371/whats-the-point-of-const-pointers).

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -0,0 +1,174 @@
...
+      inverse_permute_copy(n, nt._actual_v, data, nt._permute);
+
+      if (nt.nrn_fast_imem) {
+        n = (*nrn2core_type_return_)(i_membrane_, tid, data, mdata);
+        assert(n == size_t(nt.end) && data);
+        inverse_permute_copy(n, nt.nrn_fast_imem->nrn_sav_rhs, data, nt._permute);
+      }
+    }
+
+    for (NrnThreadMembList* tml = nt.tml; tml; tml = tml->next) {
```
pramodk, line 143: just a remark : we can start using `auto` more : `for (auto& tml = nt.tml;`

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -0,0 +1,174 @@
...
+      int mtype = tml->index;
+      Memb_list* ml = tml->ml;
+      n = (*nrn2core_type_return_)(mtype, tid, data, mdata);
```
pramodk, line 146: I should look at neuron side but curious how `(*nrn2core_type_return_)(....);` works when called multiple times.

Is that first argument `mtype` decides what type of data is it?

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -0,0 +1,174 @@
...
+  if (!nrn2core_type_return_) {
+    return;
+  }
+  for (int tid = 0; tid < nrn_nthread; ++tid) {
+    size_t n = 0;
+    double* data = nullptr;
+    double** mdata = nullptr;
+    NrnThread& nt = nrn_threads[tid];
+
+    n = (*nrn2core_type_return_)(0, tid, data, mdata); // 0 means time
```
pramodk, line 126: neuron doesn't have mechanism type `0` and that is the reason we are using first parameter`0` to nrn2core_type_return_?

I wonder if we should change this to some different value (e.g. `< 0`) in case above assumption change?

pramodk (approved):
@nrnhines : overall looks good for me. Just left few remarks but already approving it.

coreneuron/io/core2nrn_data_return.cpp
nrnhines: Would this really work here without changing NrnThreadMembList into std::vector or giving it some kind of explicit iterator operator?

coreneuron/io/core2nrn_data_return.cpp
nrnhines: tid is the thread index, mtype is the mechanism type. A lot of this is in fact thread safe on the neuron side and I hope that the refactoring project will take advantage of it so that all the loops over nrn_threads can be done in parallel.

coreneuron/io/core2nrn_data_return.cpp
nrnhines: > change this to some different value
Good idea Should add to the enum (on both NEURON and CoreNEURON side).

coreneuron/io/core2nrn_data_return.cpp
nrnhines: Thanks. Does the fact that permuted_src and permute elements do not change mean that I should
```
static void inverse_permute_copy(size_t n, const double* const permuted_src, double* const dest,
                                 const int* const permute)
```
I have to admit I never thought it was obvious which const meant what.

pramodk: > Thanks. Does the fact that permuted_src and permute elements do not change mean that I should

> static void inverse_permute_copy(size_t n, const double* const permuted_src, double* const dest,
>                                 const int* const permute)

In this case,

* `permuted_src` and `permute` are read-only and remain constant. So it's important to declare them with `double* const ` and `int* const`
* additionally, when we say `const int* permute`, this means we can't change permute itself with `permute = new_ptr`. This is fine as well. But just want to mention that as pointer is passed by value (and not reference),  even if pointer is changed, it doesn't affect the caller. 


> static void soa2aos_inverse_permute_copy(size_t n, int sz, int stride,
  const double* const src, double* const * const dest, const int* const permute)

In this case we are changing values inside dest[i][j] and hence it can't be `* const dest` ?

coreneuron/io/core2nrn_data_return.cpp
pramodk: auto just deduce the type from assignment. If `nt.tml` becomes container like std::vector then we have to use `for(auto& tml : nt.tml)`.

alexsavulescu (approved):

coreneuron/io/core2nrn_data_return.cpp
alexsavulescu: To take care of with refactoring

PR #384, opened 2020-08-20
Remove extra include path

ohm314:
There is no need to point into the random123 path as we are already
including the top-level path and all generated code includes paths based
on the coreneuron include directory.


ohm314:
not true! actually the additional `-I` is useful because modfiles typically include random123.h differently in the VERBATIM block...

PR #385, opened 2020-08-20
Cleaning netpar.cpp

alkino:


pramodk:
@nrnhines : do you agree that removing `#if NRNSTAT` code is fine? I haven't used that for years and I think we are almost never (?) looked at those stats?

nrnhines:
I have no objection to removing them. They are just counts and as such are available from NEURON.

coreneuron/utils/nrnmutdec.h
```diff
@@ -132,6 +132,32 @@ class OMP_Mutex {
...
+        OMP_Mutex(const OMP_Mutex&&) = delete;
+        OMP_Mutex& operator=(const OMP_Mutex&&) = delete;
+
+        // Basic Lockable
+        void lock() {}
+
+        void unlock() {}
+
+        // Lockable
+        bool try_lock() {}
```
pramodk, line 159: @alkino : missed here : 

```
utils/nrnmutdec.h:159:26: warning: control reaches end of non-void function
      [-Wreturn-type]
        bool try_lock() {}
```

PR #390, opened 2020-09-08
If a netcon_srcgid is negative, need to determine the thread.

nrnhines:
Only for direct transfer mode is it allowed that a negative srcgid is not
in the same thread as the NetCon target.

To enable thread determination an std::vector<int> involving the name
netcon_negsrcgid_tid is associated with netcon_srcgid in that when a negative
gid appears in netcon_srcgid, the tid is the value of the element in
netcon_negsrcgid_tid.

This pr is assciated with neuronsimulator/nrn#735

### DIFF
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -186,11 +186,11 @@ std::map<int, InputPreSyn*> gid2in;
 std::vector<NetCon*> netcon_in_presyn_order_;
 
 /// Only for setup vector of netcon source gids
-std::vector<int*> netcon_srcgid;
+std::vector<int*> nrnthreads_netcon_srcgid;
 
-/// If a netcon_srcgid is negative, need to determine the thread when
+/// If a nrnthreads_netcon_srcgid is negative, need to determine the thread when
 /// in order to use the correct neg_gid2out[tid] map
-std::vector<std::vector<int> > netcon_negsrcgid_tid;
+std::vector<std::vector<int> > nrnthreads_netcon_negsrcgid_tid;
 
 /* read files.dat file and distribute cellgroups to all mpi ranks */
 void nrn_read_filesdat(int& ngrp, int*& grp, const char* filesdat) {
@@ -288,10 +288,10 @@ void determine_inputpresyn() {
         // associate gid with InputPreSyn and increase PreSyn and InputPreSyn count
         nt.n_input_presyn = 0;
         std::vector<int> dummy;
-        std::vector<int>& negsrcgid_tid = corenrn_embedded ? netcon_negsrcgid_tid[ith] : dummy;
+        std::vector<int>& negsrcgid_tid = corenrn_embedded ? nrnthreads_netcon_negsrcgid_tid[ith] : dummy;
         size_t i_tid = 0;
         for (int i = 0; i < nt.n_netcon; ++i) {
-            int gid = netcon_srcgid[ith][i];
+            int gid = nrnthreads_netcon_srcgid[ith][i];
             if (gid >= 0) {
                 /// If PreSyn or InputPreSyn is already in the map
                 auto gid2out_it = gid2out.find(gid);
@@ -380,16 +380,16 @@ void determine_inputpresyn() {
 
     // fill the netcon_in_presyn_order and recompute nc_cnt_
     // note that not all netcon_in_presyn will be filled if there are netcon
-    // with no presyn (ie. netcon_srcgid[nt.id][i] = -1) but that is ok since they are
+    // with no presyn (ie. nrnthreads_netcon_srcgid[nt.id][i] = -1) but that is ok since they are
     // only used via ps.nc_index_ and ps.nc_cnt_;
     for (int ith = 0; ith < nrn_nthread; ++ith) {
         NrnThread& nt = nrn_threads[ith];
         std::vector<int> dummy;
-        std::vector<int>& negsrcgid_tid = corenrn_embedded ? netcon_negsrcgid_tid[ith] : dummy;
+        std::vector<int>& negsrcgid_tid = corenrn_embedded ? nrnthreads_netcon_negsrcgid_tid[ith] : dummy;
         size_t i_tid = 0;
         for (int i = 0; i < nt.n_netcon; ++i) {
             NetCon* nc = nt.netcons + i;
-            int gid = netcon_srcgid[ith][i];
+            int gid = nrnthreads_netcon_srcgid[ith][i];
             int tid = ith;
             if (!negsrcgid_tid.empty() && gid < -1) {
               tid = negsrcgid_tid[i_tid++];
@@ -416,11 +416,11 @@ void determine_inputpresyn() {
 /// Clean up
 void nrn_setup_cleanup() {
     for (int ith = 0; ith < nrn_nthread; ++ith) {
-        if (netcon_srcgid[ith])
-            delete[] netcon_srcgid[ith];
+        if (nrnthreads_netcon_srcgid[ith])
+            delete[] nrnthreads_netcon_srcgid[ith];
     }
-    netcon_srcgid.clear();
-    netcon_negsrcgid_tid.clear();
+    nrnthreads_netcon_srcgid.clear();
+    nrnthreads_netcon_negsrcgid_tid.clear();
     neg_gid2out.clear();
 }
 
@@ -476,9 +476,9 @@ void nrn_setup(const char* filesdat,
     /// std::map<int, PreSyn*> gid2out;
     gid2out.clear();
 
-    netcon_srcgid.resize(nrn_nthread);
+    nrnthreads_netcon_srcgid.resize(nrn_nthread);
     for (int i = 0; i < nrn_nthread; ++i)
-        netcon_srcgid[i] = nullptr;
+        nrnthreads_netcon_srcgid[i] = nullptr;
 
     // gap junctions
     if (nrn_have_gaps) {
@@ -500,7 +500,7 @@ void nrn_setup(const char* filesdat,
     if (!corenrn_embedded) {
         coreneuron::phase_wrapper<coreneuron::phase::one>(userParams);
     } else {
-        netcon_negsrcgid_tid.resize(nrn_nthread);
+        nrnthreads_netcon_negsrcgid_tid.resize(nrn_nthread);
         nrn_multithread_job([](NrnThread* n) {
             Phase1 p1; 
             p1.read_direct(n->id);
@@ -509,7 +509,7 @@ void nrn_setup(const char* filesdat,
         });
     }
 
-    // from the gid2out map and the netcon_srcgid array,
+    // from the gid2out map and the nrnthreads_netcon_srcgid array,
     // fill the gid2in, and from the number of entries,
     // allocate the process wide InputPreSyn array
     determine_inputpresyn();
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -497,10 +497,10 @@ void nrn_setup(const char* filesdat,
         nrn_partrans::gap_mpi_setup(userParams.ngroup);
     }
 
+    netcon_negsrcgid_tid.resize(nrn_nthread);
     if (!corenrn_embedded) {
         coreneuron::phase_wrapper<coreneuron::phase::one>(userParams);
     } else {
-        netcon_negsrcgid_tid.resize(nrn_nthread);
         nrn_multithread_job([](NrnThread* n) {
             Phase1 p1; 
             p1.read_direct(n->id);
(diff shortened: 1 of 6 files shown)


coreneuron/io/nrn_setup.cpp
```diff
@@ -394,6 +420,7 @@ void nrn_setup_cleanup() {
             delete[] netcon_srcgid[ith];
     }
     netcon_srcgid.clear();
+    netcon_negsrcgid_tid.clear();
```
alexsavulescu, line 423: not sure if this has an impact on its usage in `netpar.cpp` ?

coreneuron/io/nrn_setup.cpp
```diff
@@ -180,6 +188,10 @@ std::vector<NetCon*> netcon_in_presyn_order_;
 /// Only for setup vector of netcon source gids
 std::vector<int*> netcon_srcgid;
 
+/// If a netcon_srcgid is negative, need to determine the thread when
+/// in order to use the correct neg_gid2out[tid] map
+std::vector<std::vector<int> > netcon_negsrcgid_tid;
```
alexsavulescu, line 193: I'd rename this to `std::vector<std::vector<int>> netcon_negsrcgid_tids;`, it's a bit ambiguous further down the line wrt  `std::vector<int> netcon_negsrcgid_tid; // entries only for negative srcgids` + callback

coreneuron/io/phase1.cpp
```diff
@@ -20,10 +21,24 @@ void Phase1::read_file(FileHandler& F) {
 
     this->output_gids = F.read_vector<int>(n_presyn);
     this->netcon_srcgids = F.read_vector<int>(n_netcon);
+    // For file mode transfer, it is not allowed that negative gids exist
+    // in different threads. So this->netcon_tids remains clear.
 
     F.close();
 }
 
+static void prsrctid(int thread_id, int n, int* srcgid, std::vector<int>& srctid) {
```
alexsavulescu, line 30: function not referenced.

coreneuron/io/nrn_setup.cpp
nrnhines: ```netcon_srcgid``` and ```netcon_negsrcgid_tid``` should stay together. Of course, that is not a direct answer to your question as maybe ```netcon_srcgid.clear()``` is not in its correct place either. In this case, I think they are both ultimately in the wrong place in a future refactoring where structure and data are separated to allow greater interactivity between NEURON and CoreNEURON. At present, however, the netpar.cpp use of these is in ```set_mindelay``` which is called
from within ```io/nrn_setup.cpp```.  As ```netcon_srcgid``` is in the class of largest arrays, it would be a good idea to review whether its existence is really necessary. But that is beyond the scope of this bugfix.

alexsavulescu:
A stack trace from the BBP pipeline:
```
MPT: (gdb) #0  0x00002aaaab44f12c in waitpid () from /lib64/libpthread.so.0
MPT: #1  0x00002aaaabdc6776 in mpi_sgi_system (
MPT: #2  MPI_SGI_stacktraceback (
MPT:     header=header@entry=0x7fffffffa3c0 "MPT ERROR: Rank 0(g:0) received signal SIGSEGV(11).\n\tProcess ID: 164164, Host: r2i0n25, Program: /jenkins/07/workspace/oss.coreneuron.pipeline/build_AoS/bin/x86_64/special-core\n\tMPT Version: HPE MPT 2"...) at sig.c:340
MPT: #3  0x00002aaaabdc6972 in first_arriver_handler (signo=signo@entry=11, 
MPT:     stack_trace_sem=stack_trace_sem@entry=0x2aaab0da0080) at sig.c:489
MPT: #4  0x00002aaaabdc6d0b in slave_sig_handler (signo=11, 
MPT:     siginfo=<optimized out>, extra=<optimized out>) at sig.c:565
MPT: #5  <signal handler called>
MPT: #6  0x00002aaaab0e9408 in std::vector<int, std::allocator<int> >::begin (
MPT:     this=0x0)
MPT:     at /gpfs/bbp.cscs.ch/ssd/apps/hpc/jenkins/deploy/compilers/2020-02-01/linux-rhel7-x86_64/gcc-4.8.5/gcc-8.3.0-4hd4baobq2/include/c++/8.3.0/bits/stl_vector.h:708
MPT: #7  0x00002aaaab0e9598 in std::vector<int, std::allocator<int> >::empty (
MPT:     this=0x0)
MPT:     at /gpfs/bbp.cscs.ch/ssd/apps/hpc/jenkins/deploy/compilers/2020-02-01/linux-rhel7-x86_64/gcc-4.8.5/gcc-8.3.0-4hd4baobq2/include/c++/8.3.0/bits/stl_vector.h:895
MPT: #8  0x00002aaaab0e574e in coreneuron::determine_inputpresyn ()
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/coreneuron/io/nrn_setup.cpp:318
MPT: #9  0x00002aaaab0e662d in coreneuron::nrn_setup (
MPT:     filesdat=0x1759f00 "/jenkins/07/workspace/oss.coreneuron.pipeline/tests/integration/ring/files.dat", is_mapping_needed=false, run_setup_cleanup=true, 
MPT:     datpath=0x6fd360 "/jenkins/07/workspace/oss.coreneuron.pipeline/tests/integration/ring", restore_path=0x7fffffffba50 "", 
MPT:     mindelay=0x2aaaab43a0f0 <coreneuron::corenrn_param+112>)
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/coreneuron/io/nrn_setup.cpp:514
MPT: #10 0x00002aaaab0cb477 in coreneuron::nrn_init_and_load_data (argc=11, 
MPT:     argv=0x7fffffffbe48, is_mapping_needed=false, run_setup_cleanup=true)
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/coreneuron/apps/main1.cpp:252
MPT: #11 0x00002aaaab0c8d52 in run_solve_core (argc=11, argv=0x7fffffffbe48)
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/coreneuron/apps/main1.cpp:488
MPT: #12 0x00002aaaaacd0bfe in solve_core (argc=11, argv=0x7fffffffbe48)
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/build_AoS/share/coreneuron/enginemech.cpp:42
MPT: #13 0x000000000040179f in main (argc=11, argv=0x7fffffffbe48)
MPT:     at /jenkins/07/workspace/oss.coreneuron.pipeline/build_AoS/share/coreneuron/coreneuron.cpp:34
```

coreneuron/io/nrn_setup.cpp
nrnhines: I agree the name is not the best. Neither is ```netcon_srcgid```. A verbose version of ```netcon_srcgid``` is  ```nrnthread_netcon_srcgid``` as the size of the vector is nthread and each element is a an ```n_netcon`` length vector of int (which should be changed to use an std vector of vectors. Similarly I could accept ```nrnthread_netcon_negsrcgid_tid```.
Let me know if you think this is worthwhile for this pull request.

coreneuron/io/phase1.cpp
nrnhines: Forgot that debugging helper was there. Will remove.

nrnhines:
```
coreneuron::determine_inputpresyn () ... coreneuron/io/nrn_setup.cpp:318
                if (!negsrcgid_tid.empty()) {
```
It puzzles me how this can be NULL since line 291
```
        std::vector<int>& negsrcgid_tid = netcon_negsrcgid_tid[ith];
```
and ```netcon_negsrcgid_tid``` was fully allocated in line 503
```
        netcon_negsrcgid_tid.resize(nrn_nthread);
```
prior to Phase1 processing.  Shouldn't that have made a nrn_nthread size vector of empty vectors?

coreneuron/io/nrn_setup.cpp
alexsavulescu: `nrnthread_netcon_negsrcgid_tid` sounds good to me, also fine to leave it as it is.

alexsavulescu:
> prior to Phase1 processing. Shouldn't that have made a nrn_nthread size vector of empty vectors?

Absolutely. My guess is that it doesn't get called. I can have a quick look, I have CoreNeuron set up.

alexsavulescu:
`determine_inputpresyn()` was using `netcon_negsrcgid_tid` as it if were resized regardless if we run `embedded` or not (and it was resized only for the `embedded` control path). The logic must be adjusted, I will propose something.

alexsavulescu:
@nrnhines  what do you think about c870bf3?

nrnhines:
c870bf3 should work but I would prefer avoiding the additional clutter by just creating an empty vector of vectors for ```netcon_negsrcgid_tid``` for the file mode transfer. ie. in line 503 of io/nrn_setup.cpp just move
```
$ git diff
diff --git a/coreneuron/io/nrn_setup.cpp b/coreneuron/io/nrn_setup.cpp
index 7135274..c232326 100644
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -497,10 +497,10 @@ void nrn_setup(const char* filesdat,
         nrn_partrans::gap_mpi_setup(userParams.ngroup);
     }
 
+    netcon_negsrcgid_tid.resize(nrn_nthread);
     if (!corenrn_embedded) {
         coreneuron::phase_wrapper<coreneuron::phase::one>(userParams);
     } else {
-        netcon_negsrcgid_tid.resize(nrn_nthread);
         nrn_multithread_job([](NrnThread* n) {
             Phase1 p1; 
             p1.read_direct(n->id);
```

nrnhines:
I did not notice that 7af3e93 was the same as my last suggestion.  What went wrong with that that kept you going with c870bf3

alexsavulescu:
> I did not notice that [7af3e93](https://github.com/BlueBrain/CoreNeuron/commit/7af3e9369ae6b69a4db17604cf6f1f90ced84855) was the same as my last suggestion. What went wrong with that that kept you going with [c870bf3](https://github.com/BlueBrain/CoreNeuron/commit/c870bf32b03e936654394789a48e9a0bdb749a13)

I know it's more verbose, just wanted to be closer to your original logic of resizing only or direct mode. Then I realized it makes things more obvious if we linked `netcon_negsrcgid_tid` to `corenrn_embedded` instead of relying on it being empty or not. 

We can revert back to 7af3e93 if you prefer, however in that case we would need to keep the following (because `empty()` will not be a valid check, since `netcon_negsrcgid_tid` is always resized) :
```diff
diff --git a/coreneuron/io/phase1.cpp b/coreneuron/io/phase1.cpp
index fd5d3287..eff0dc4c 100644
--- a/coreneuron/io/phase1.cpp
+++ b/coreneuron/io/phase1.cpp
@@ -58,6 +58,10 @@ void Phase1::read_direct(int thread_id) {
     delete[] netcon_srcgid;
 }

+extern "C" {
+extern bool corenrn_embedded;
+}
+
 void Phase1::populate(NrnThread& nt, OMP_Mutex& mut) {
     nt.n_presyn = this->output_gids.size();
     nt.n_netcon = this->netcon_srcgids.size();
@@ -66,7 +70,7 @@ void Phase1::populate(NrnThread& nt, OMP_Mutex& mut) {
     std::copy(this->netcon_srcgids.begin(), this->netcon_srcgids.end(),
               netcon_srcgid[nt.id]);

-    if (!coreneuron::netcon_negsrcgid_tid.empty()) { // multiple threads and direct mode.
+    if (corenrn_embedded) { // multiple threads and direct mode.
         coreneuron::netcon_negsrcgid_tid[nt.id] = this->netcon_negsrcgid_tid;
     }
```

alexsavulescu (approved):

pramodk:
@alexsavulescu  @nrnhines : changes look good to me. But I stumble on this:

```cpp

        std::vector<int> dummy;
        std::vector<int>& negsrcgid_tid = corenrn_embedded ? nrnthreads_netcon_negsrcgid_tid[ith] : dummy;

...
        if (!negsrcgid_tid.empty()) {
            tid = negsrcgid_tid[i_tid++];
         }
```

if anyway we have to protect access to `negsrcgid_tid` with `if(!negsrcgid_tid.empty())`, then may be more clear to remove usage of dummy and use : 

```cpp
        if (corenrn_embedded) {
            tid = nrnthreads_netcon_negsrcgid_tid[ith][i_tid++];
         }
```

which was this comment https://github.com/BlueBrain/CoreNeuron/pull/390#issuecomment-691496551 I think. (reference to empty `dummy` vector and usage pattern in the code looked bit confusing)

nrnhines:
I like that suggestion. Good balance between @alexsavulescu point about "closer to your original logic of resizing only for direct mode." and just creatiing a vector of empty vectors.

alexsavulescu:
Given that `nrnthreads_netcon_negsrc_gid_tid` is not populated for single thread or file transfer,  7af3e93 is definitely a win.

PR #394, opened 2020-09-21
Bug fix for PR #385 : add missing `const` for virtual method

pramodk:
- changing base class method prototype result into bug for
    virtual method in PlayRecordEvent

@alkino : culprit was [this commit](https://github.com/BlueBrain/CoreNeuron/pull/385/files#diff-c79cec806ef8ef3de58d9ed516974213) : 

image


ohm314 (approved):
lgtm

PR #395, opened 2020-10-02
Allow NetSendBuffer_t to grow

ohm314:
Until now NetSendBuffer_t was allocated at a fixed size that could not
be changed. When using buffered net_send and invoking multiple
net_send's in INITIAL blocks in NMODL it may happen that we run out of
buffer. Currently this case is not handled and causes a buffer overflow.

As a first step NetSendBuffer_t has been given a constructor and
destructor as well as a grow function that doubles all internal buffer
sizes and copies data to the new buffers.

### DIFF
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -733,12 +733,7 @@ void nrn_cleanup() {
             NetSendBuffer_t* nsb = ml->_net_send_buffer;
             if (nsb) {
                 if (nsb->_size) {
-                    free_memory(nsb->_sendtype);
-                    free_memory(nsb->_vdata_index);
-                    free_memory(nsb->_pnt_index);
-                    free_memory(nsb->_weight_index);
-                    free_memory(nsb->_nsb_t);
-                    free_memory(nsb->_nsb_flag);
+                    free(nsb);
                 }
                 free_memory(nsb);
             }
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -732,10 +732,7 @@ void nrn_cleanup() {
 
             NetSendBuffer_t* nsb = ml->_net_send_buffer;
             if (nsb) {
-                if (nsb->_size) {
-                    free(nsb);
-                }
-                free_memory(nsb);
+                delete nsb;
             }
 
             if (tml->dependencies)
--- a/coreneuron/io/phase2.cpp
+++ b/coreneuron/io/phase2.cpp
@@ -465,7 +465,6 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
         Memb_list* ml = ml_list[type];
         if (ml) {  // needs a NetSendBuffer
             // begin with a size equal to twice number of instances
-            // at present there is no provision for dynamically increasing this.
             NetSendBuffer_t* nsb = new NetSendBuffer_t(ml->nodecount * 2);
             ml->_net_send_buffer = nsb;
         }
--- a/coreneuron/io/phase2.cpp
+++ b/coreneuron/io/phase2.cpp
@@ -464,23 +464,10 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
         // Does this thread have this type.
         Memb_list* ml = ml_list[type];
         if (ml) {  // needs a NetSendBuffer
-            NetSendBuffer_t* nsb = (NetSendBuffer_t*)ecalloc_align(1, sizeof(NetSendBuffer_t));
-            ml->_net_send_buffer = nsb;
-
             // begin with a size equal to twice number of instances
             // at present there is no provision for dynamically increasing this.
-            nsb->_size = ml->nodecount * 2;
-            nsb->_cnt = 0;
-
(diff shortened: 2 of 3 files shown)


coreneuron/io/phase2.cpp
```diff
@@ -464,23 +464,10 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
         // Does this thread have this type.
         Memb_list* ml = ml_list[type];
         if (ml) {  // needs a NetSendBuffer
-            NetSendBuffer_t* nsb = (NetSendBuffer_t*)ecalloc_align(1, sizeof(NetSendBuffer_t));
-            ml->_net_send_buffer = nsb;
-
             // begin with a size equal to twice number of instances
             // at present there is no provision for dynamically increasing this.
```
nrnhines, line 468: now an incorrect comment

coreneuron/io/nrn_setup.cpp
```diff
@@ -733,12 +733,7 @@ void nrn_cleanup() {
             NetSendBuffer_t* nsb = ml->_net_send_buffer;
             if (nsb) {
                 if (nsb->_size) {
-                    free_memory(nsb->_sendtype);
-                    free_memory(nsb->_vdata_index);
-                    free_memory(nsb->_pnt_index);
-                    free_memory(nsb->_weight_index);
-                    free_memory(nsb->_nsb_t);
-                    free_memory(nsb->_nsb_flag);
+                    free(nsb);
```
nrnhines, line 736: Line 738 below uses free_memory(nsb). Shouldn't this be eliminated and  738 changed to delete nsb?

coreneuron/io/phase2.cpp
```diff
@@ -464,23 +464,10 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
...
-            nsb->_sendtype = (int*)ecalloc_align(nsb->_size, sizeof(int));
-            nsb->_vdata_index = (int*)ecalloc_align(nsb->_size, sizeof(int));
-            nsb->_pnt_index = (int*)ecalloc_align(nsb->_size, sizeof(int));
-            nsb->_weight_index = (int*)ecalloc_align(nsb->_size, sizeof(int));
-            // when == 1, NetReceiveBuffer_t is newly allocated (i.e. we need to free previous copy
-            // and recopy new data
-            nsb->reallocated = 1;
-            nsb->_nsb_t = (double*)ecalloc_align(nsb->_size, sizeof(double));
-            nsb->_nsb_flag = (double*)ecalloc_align(nsb->_size, sizeof(double));
+            NetSendBuffer_t* nsb = new NetSendBuffer_t(ml->nodecount * 2);
```
pramodk, line 469: In case of GPU (with unified memory), need to check how this is allocated on GPUs.

coreneuron/mechanism/mechanism.hpp
```diff
@@ -76,6 +79,50 @@ struct NetSendBuffer_t {
...
+        grow_buf(&_vdata_index, _size, new_size);
+        grow_buf(&_pnt_index, _size, new_size);
+        grow_buf(&_weight_index, _size, new_size);
+        grow_buf(&_nsb_t, _size, new_size);
+        grow_buf(&_nsb_flag, _size, new_size);
+    }
+
+    private:
+        template <typename T>
+        void grow_buf(T** buf, int size, int new_size) {
```
pramodk, line 118: Would `realloc` be sufficient here? (I haven't checked but do we have `realloc_align` or something already?)

coreneuron/io/phase2.cpp
ohm314: ah indeed!

coreneuron/io/phase2.cpp
ohm314: I use `ecalloc_align` in the constructor, that wraps the cuda malloc too, no? (here: https://github.com/BlueBrain/CoreNeuron/pull/395/files#diff-59c3a6f34686c1d23c8a82ac944d22cc)

coreneuron/mechanism/mechanism.hpp
ohm314: No, we don't have it. I could of course implement that. there is no `_aligned_realloc` in stdlib, I checked that. Hence I had to do the allocation, and `memcpy` myself. But on the other hand, we still have the plan to replace `memory.h` by a newer implementation that uses proper C++ allocators. Not sure how concrete this plan is. If yes, then this is good enough, if not I can add the generic implementation of aligned realloc.

coreneuron/io/nrn_setup.cpp
ohm314: Fixed

nrnhines (commented):
LGTM. But is it a no-no to approve WIP

ohm314:
please retest

coreneuron/io/phase2.cpp
pramodk: Everything should be fine after extending this via `MemoryManaged`.

coreneuron/mechanism/mechanism.hpp
pramodk: That's true. And I think that all aligned* should go away now (not really required nowadays).

pramodk (approved):
LGTM

Issue #402, opened 2020-10-13
multiple definition of `at_time' while compiling on GPU


pramodk:
```
/global/homes/r/roybens/install/share/nrn/nrnmain.cpp:
/usr/bin/ld: /global/homes/r/roybens/install/lib/libcoreneuron.a(cvodestb.cpp.o): in function `at_time':
/project/projectdirs/m2043/roybens/neuron8/corneuron/nrn/external/coreneuron/coreneuron/network/cvodestb.cpp:111: multiple definition of `at_time'; /global/homes/r/roybens/install/lib/libnrniv.a(cvodestb.cpp.o):/project/projectdirs/m2043/roybens/neuron8/corneuron/nrn/src/nrniv/../nrncvode/cvodestb.cpp:131: first defined here
pgacclnk: child process exit status 1: /usr/bin/ld
make: *** [/global/homes/r/roybens/install/bin/nrnmech_makefile:89: special] Error 2
roybens@cgpu02:~/NeuroGPU/Benchmarking/BBP_TTPC_clean> cd /project/projectdirs/m2043/roybens/
```

iomaganaris:
I tried building `CoreNEURON` as `NEURON` submodule with `pgi`, with `CORENRN_ENABLE_GPU=ON` and with all the libraries `static` but I couldn't reproduce the issue. Could you provide the `cmake` command or any more info on how `NEURON/CoreNEURON` were build?

pramodk:
This was on NERSC Cori (Cray) with:

```
cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/install \
	-DNRN_ENABLE_INTERVIEWS=OFF \
	-DNRN_ENABLE_RX3D=OFF \
	-DNRN_ENABLE_CORENEURON=ON \
	-DPYTHON_EXECUTABLE=`which python` \
	-DCORENRN_ENABLE_MPI=ON \
	-DCORENRN_ENABLE_OPENMP=OFF \
	-DCORENRN_ENABLE_GPU=ON \
	-DNRN_ENABLE_BINARY_SPECIAL=ON \
	-DNRN_ENABLE_SHARED=OFF \
	-DNRN_ENABLE_INTERNAL_READLINE=ON
```

If we are not able to reproduce, I think the issue is "real" i.e. we have now moved `at_time` in global scope and that will conflict with NEURON?

iomaganaris:
I was thinking the same thing however I saw this:
```
[...]

PR #403, opened 2020-10-14
Move at_time inside coreneuron namespace and redefine at_time in ispc

iomaganaris:
- Fixes https://github.com/BlueBrain/CoreNeuron/issues/402


pramodk (approved):
LGTM

PR #407, opened 2020-10-22
No more check parameters from argc / argv but from the parser. Fix #184

### LINKED ISSUE #184: -mpi still needed with --read-config
Even if `-mpi` flag is included in the configure file used with `--read-config`, it is still needed to be added in the command executing `CoreNeuron`.
This is due to the fact that `-mpi` flag is checked if exists in the arguments of the executable.
If it is not provided, there simulation fails with `Segmentation Faults`

alkino:
We have to decide what to do with other arguments just like `-p4` that is not valid with the new parser.

Does someone know what is `nrn_music`?

### DIFF
--- a/coreneuron/mpi/nrnmpi.cpp
+++ b/coreneuron/mpi/nrnmpi.cpp
@@ -44,8 +44,6 @@ THE POSSIBILITY OF SUCH DAMAGE.
 namespace coreneuron {
 
 #if NRNMPI
-int nrnmusic;
-
 MPI_Comm nrnmpi_world_comm;
 MPI_Comm nrnmpi_comm;
 MPI_Comm nrn_bbs_comm;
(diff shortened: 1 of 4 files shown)


coreneuron/mpi/nrnmpi.cpp
```diff
@@ -73,7 +74,7 @@ void nrnmpi_init(int nrnmpi_under_nrncontrol, int* pargc, char*** pargv) {
                 b = true;
                 break;
             }
-            if (strcmp("--mpi", (*pargv)[i]) == 0) {
+            if (corenrn_param.mpi_enable) {
```
alexsavulescu, line 77: Can the enclosing loop be modified to smtg like: 

```cpp 
if (corenrn_param.mpi_enable || corenrn_param.p4)
{
 b = true;
}
```

alexsavulescu:
> Does someone know what is `nrn_music`?

I've seen this in NEURON, however I couldn't find any reference for its use. I believe it's a sort of NET events relayer. @nrnhines @pramodk would you know more about it?

coreneuron/mpi/nrnmpi.cpp
alkino: The problem with **p4** is that it does not exists anymore (coreneuron side at least), `-p` (for `--pattern`) take a file path.

alkino:
Maybe I need to add `|| arg == "--mpi"` because there is maybe someone / somewhere giving it args!

coreneuron/mpi/nrnmpi.cpp
alexsavulescu: At the very least `corenrn_param.mpi_enable` can be checked outside the loop

alkino: ahahaha of course, I'm so disappointed by this piece of code, that I miss that xD

pramodk:
> Does someone know what is nrn_music?

Currently it's not but I wouldn't remove ( it could come back).

> Maybe I need to add || arg == "--mpi" because there is maybe someone / somewhere giving it args!

@alkino : wanted to ask same thing because NEURON (and neurodamus) pass arguments via CLI.

alkino:
Endly remove all of this, as we think that `main1.cpp` is the only user of `nrnmpi_init`.

alkino:
Pr. @nrnhines can we have a really quick and fast review of this?

Regarding callers of `nrnmpi_init` (only us?), and `nrn_music` (useless?).

Regards

pramodk:
CoreNEURON will be either called by NEURON as library (where MPI is already initialised) or CoreNEURON will be launched independently. Hence extra options we had (e.g. always initialise MPI) are not necessary in my opinion.

Thanks for cleaning that up!

pramodk (approved):

alkino:
`The base branch restricts merging to authorized users.` I leave authorized users do.

coreneuron/mpi/nrnmpi.cpp
```diff
@@ -30,6 +30,7 @@ THE POSSIBILITY OF SUCH DAMAGE.
 #include <sys/time.h>
 
 #include "coreneuron/nrnconf.h"
+#include "coreneuron/apps/corenrn_parameters.hpp"
```
iomaganaris, line 33: I think that this is not needed anymore?

iomaganaris (commented):
LGTM. Just a small nitpick

coreneuron/mpi/nrnmpi.cpp
pramodk: yes! thanks Ioannis!

pramodk (approved):

coreneuron/mpi/nrnmpi.cpp
alkino: of course, sorry

iomaganaris (approved):

Issue #405, opened 2020-10-21
CMake option to enable/disable legacy units


pramodk:
NEURON is changing units values (CODATA 2018 constants), see : https://github.com/BlueBrain/mod2c/pull/47
As discussed in https://github.com/BlueBrain/mod2c/pull/47#issuecomment-711156915, we need to add `-DCORENEURON_ENABLE_LEGACY_UNIT=ON` option to CMake to enable/disable use of legacy units.

What needs to be done:
- [ ] add CMake switch CORENEURON_ENABLE_LEGACY_UNIT
- [x] `share/nrnunits.lib` in MOD2C should become template file `share/nrnunits.lib.in`
- [x] Depending on value of `CORENEURON_ENABLE_LEGACY_UNIT` option and using CMake's [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) functionality, `share/nrnunits.lib.in` should be converted to `share/nrnunits.lib` at build time with old or new values of different constants. 

The same mechanism was used in earlier NEURON versions  (see usage of `NRN_ENABLE_LEGACY_FR` in [cmake/ConfigFileSetting.cmake](https://github.com/neuronsimulator/nrn/blob/6aff783a9e8c031a6071f7b917dca753ba7cfc19/cmake/ConfigFileSetting.cmake#L108)).

@iomaganaris  can provide additional context.

alkino:
Why is this ticket in CoreNeuron and not in mod2c?
What about nmodl?

iomaganaris:
`mod2c` has its `nrnunits.lib` in https://github.com/BlueBrain/mod2c/blob/master/share/nrnunits.lib and `NMODL` in https://github.com/BlueBrain/nmodl/blob/master/share/nrnunits.lib. They should both be edited to be able to select the legacy or new units.

nrnhines:
the coreneuron eion.c file also uses this option.

alkino:
Should I copy the file from neuronsimulator/nrn? They differ about available values sometimes.

nrnhines:
You need to copy ```nrn/share/lib/nrnunits.lib.in``` . That uses ```@LegacyY@``` and ```@LegacyN@``` that get configured into 
```#``` and ```<empty>``` or vice versa depending on ```CORENRN_ENABLE_LEGACY_UNITS``` to create the ```nrnunits.lib``` file.

alkino:
[...]

PR #417, opened 2020-11-11
Add propagation of legacy units for nmodl

alkino:


alkino:
This is the last piece of #405

pramodk (approved):

CMakeLists.txt
```diff
@@ -246,7 +246,9 @@ if(CORENRN_ENABLE_LEGACY_UNITS)
 else()
   set(CORENRN_USE_LEGACY_UNITS 0)
 endif()
+# Propagate Legacy Units flag to backends.
 set(MOD2C_ENABLE_LEGACY_UNITS ${CORENRN_ENABLE_LEGACY_UNITS} CACHE BOOL "" FORCE)
+set(NMODL_ENABLE_LEGACY_UNITS ${CORENRN_ENABLE_LEGACY_UNITS} CACHE BOOL "" FORCE)
```
alkino, line 252: We can use option instead of `set` because there are option in sub project.
Any wish @pramodk ?

pramodk: if there will be `options` twice, would that work same way? I don't have any preference.

PR #424, opened 2020-11-13
Fix cmake to find ispc compiler

alkino:
NEURON_BRANCH=master

### DIFF
--- a/CMake/CMakeTestISPCCompiler.cmake
+++ b/CMake/CMakeTestISPCCompiler.cmake
@@ -22,7 +22,7 @@ unset(CMAKE_ISPC_COMPILER_WORKS CACHE)
 # set and cmake stops processing commands and will not generate any makefiles or projects.
 # ~~~
 if(NOT CMAKE_ISPC_COMPILER_WORKS)
-  PrintTestCompilerStatus("ISPC")
+  printtestcompilerstatus("ISPC" "")
   file(WRITE ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/kernel.ispc
        "export uniform int kernel(uniform int){return 0;}\n")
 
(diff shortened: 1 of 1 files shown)


CMake/CMakeTestISPCCompiler.cmake
```diff
@@ -1,3 +1,4 @@
+include(CMakeTestCompilerCommon)
```
pramodk, line 1: could you confirm is this works with Cmake 3.8/3.9?

alkino: of course not :'(

pramodk:
Retest this please

alkino:
Please retest

alkino:
Please retest

cattabiani (approved):

CMake/CMakeTestISPCCompiler.cmake
pramodk: Does this fail with CMake v3.8? (If so, it shouldn't have merged?)

According to below, we say CMake 3.7 compatible: https://github.com/BlueBrain/CoreNeuron/blob/c03400580f6eeefdd4de994ade8aba609ac32042/CMakeLists.txt#L7 

We should implement this in CoreNEURON CI : https://github.com/neuronsimulator/nrn/pull/812

alkino:
No it doesn't

CMake/CMakeTestISPCCompiler.cmake
alkino: no it doesn't fail

alkino: what was failing was to use the new form of functions

pramodk: ok, thanks for clarification!

PR #425, opened 2020-11-17
Try to stop using FILE*

alkino:


alkino:
Really hard to make it works

PR #433, opened 2020-11-28
CoreNEURON CI via Github Actions

alexsavulescu:
CI_BRANCHES:NEURON_BRANCH=master

### DIFF
--- a/.github/workflows/coreneuron-ci.yml
+++ b/.github/workflows/coreneuron-ci.yml
@@ -1,0 +1,162 @@
+name: CoreNEURON CI
+
+on:
+  push:
+    branches: 
+      - master
+      - live-debug*
+  pull_request:
+    branches:
+      - master
+
+env:
+  BUILD_TYPE: Release
+  DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
+  DEFAULT_PY_VERSION: 3.8
+
+jobs:
+  ci:
+    runs-on: ${{ matrix.os }}
+
+    name: ${{ matrix.os }} - ${{ toJson(matrix.config) }})
+
+    env:
+      INSTALL_DIR: install
+      SDK_ROOT: $(xcrun --sdk macosx --show-sdk-path)
+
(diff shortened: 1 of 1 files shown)


.github/workflows/coreneuron-ci.yml
```diff
@@ -0,0 +1,162 @@
...
+        shell: bash
+        working-directory: ${{runner.workspace}}/CoreNeuron
+        run:  |
+          if [ -n "$GCC_VERSION" ]; then
+              export CXX="g++-${GCC_VERSION}" CC="gcc-${GCC_VERSION}";
+          fi
+          
+          if [[ "${{ startsWith(matrix.os, 'macOS') }}" = "true" ]]; then
+              git submodule update --init -- external/Random123
+              export PATH=${{runner.workspace}}/CoreNeuron/external/Random123:/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH;
```
pramodk, line 110: Curious why random123 is added to PATH?

pramodk (approved):

.github/workflows/coreneuron-ci.yml
alexsavulescu: Some strange Random123 is found in the image. Had to put that first

pramodk: Ok..but PATH is for binaries/executables.Random123 is header only library. I wonder how PATH affects that. Do you have build where it was failing?

alexsavulescu: https://github.com/alexsavulescu/CoreNeuron/runs/1464884089?check_suite_focus=true

alexsavulescu: CMake was finding this : `Found Random123: /Users/runner/hostedtoolcache/stack/2.5.1/x64`, and found that to be in the PATH.

pramodk: Quite surprising !

PR #442, opened 2020-12-18
Create a new script nrnivmodl-core

alkino:
**The old bash script is limited so do a new one with python >= 2.7**

The idea behind the rewriting is having several options in the same time.
Right now if you want a backend with mod2c or nmodl you need to compile twice CoreNeuron, let it be an option of nrnivmodl-core.

Supported features
- [x] ISPC
- [ ] Openacc
- [ ] omp
- [ ] gpu
- [ ] Installed nmodl
- [ ] nmodl compiled
- [ ] mod2c compiled


alkino:
This it too hard

pramodk:
keeping this as draft so that we don't accidentally delete this branch. we have some useful work here that will be needed eventually.

PR #443, opened 2020-12-20
Update nmodl submodule to latest master and avoid recursive cloning

pramodk:
- nmodl initialize necessary submodules. so we do not need to
    do recursive initialize that adds extra submodules for no reason
  - update nmodl to latest master

Todo:

- [x] update nmodl from branch to master once PR is merged

### DIFF
--- a/CMake/AddCLI11Submodule.cmake
+++ b/CMake/AddCLI11Submodule.cmake
@@ -17,12 +17,12 @@ find_package_handle_standard_args(CLI11 REQUIRED_VARS CLI11_PROJ)
 if(NOT CLI11_FOUND)
   find_package(Git 1.8.3 QUIET)
   if(NOT ${GIT_FOUND})
-    message(FATAL_ERROR "git not found, clone repository with --recursive")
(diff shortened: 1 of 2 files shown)


CMake/AddCLI11Submodule.cmake
```diff
@@ -17,12 +17,12 @@ find_package_handle_standard_args(CLI11 REQUIRED_VARS CLI11_PROJ)
 if(NOT CLI11_FOUND)
   find_package(Git 1.8.3 QUIET)
   if(NOT ${GIT_FOUND})
-    message(FATAL_ERROR "git not found, clone repository with --recursive")
+    message(FATAL_ERROR "git not found, clone repository with")
```
nrnhines, line 20: Don't understand why the message ends with 'with'. Do you mean "git not found. Please install git."

nrnhines (approved):

CMake/AddCLI11Submodule.cmake
pramodk: Ah yes, thats an error!

PR #445, opened 2020-12-22
Printing mechanism count and mech stats to log

### LINKED ISSUE #444: Optionally count mechanisms in the simulation
## The Problem

Especially when running benchmark it is useful to know how many instances of each mechanism are overall run in the current simulation. This together with an analysis of the number of state variables in each mechanism type allows us to calculate for instance the total number of state variables being simulated.

## The Solution

@pramodk has in fact already implemented this feature in 055368d.

The commit should be cherry-picked and augmented with proper runtime option handling to make this a runtime option that can be enabled whenever we're interested in calculating the number of mechanisms.

ohm314:
Access feature through `--count_mechs` runtime option.

**Description**

Cherry picked 055368d from @pramodk and added a runtime option to enable/disable this step.

Fixes #444

**How to test this?**

```bash
cmake ..
make -j8
nrnivmodl mod
./bin/nrnivmodl-core mod
./x86_64/special script.py
./x86_64/special-core --count_mechs --tstop=10 --datpath=coredat
```

### DIFF
--- a/coreneuron/io/mech_report.cpp
+++ b/coreneuron/io/mech_report.cpp
@@ -1,0 +1,51 @@
+#include <iostream>
+#include <vector>
+
+#include "coreneuron/coreneuron.hpp"
+#include "coreneuron/mpi/nrnmpi.h"
+#include "coreneuron/mpi/nrnmpi_impl.h"
+
+namespace coreneuron {
+
+/** display global mechanism count */
+void write_mech_report() {
+#if NRNMPI
+    /// mechanim count across all gids, local to rank
+    auto n_memb_func = corenrn.get_memb_funcs().size();
+    std::vector<unsigned long long> local_mech_count(n_memb_func, 0);
+
+    /// each gid record goes on separate row, only check non-empty threads
+    for (size_t i = 0; i < nrn_nthread; i++) {
+        const auto& nt = nrn_threads[i];
+        std::vector<int> mech_count(n_memb_func, 0);
+        for (auto* tml = nt.tml; tml; tml = tml->next) {
+            int type = tml->index;
+            const auto& ml = tml->ml;
+            local_mech_count[type] += ml->nodecount;
+        }
+    }
+
+    /// get global sum of all mechanism instances
+    std::vector<unsigned long long> total_mech_count(n_memb_func);
+    MPI_Allreduce(&local_mech_count[0],
+                  &total_mech_count[0],
+                  local_mech_count.size(),
+                  MPI_UNSIGNED_LONG_LONG,
+                  MPI_SUM,
+                  MPI_COMM_WORLD);
+
+    /// print global stats to stdout
+    if (nrnmpi_myid == 0) {
+        printf("\n================ MECHANISMS COUNT BY TYPE ==================\n");
+        printf("%4s %20s %10s\n", "Id", "Name", "Count");
+        for (int i = 0; i < total_mech_count.size(); i++) {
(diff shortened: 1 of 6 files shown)


coreneuron/io/mech_report.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "coreneuron/mpi/nrnmpi.h"
+#include "coreneuron/mpi/nrnmpi_impl.h"
+
+namespace coreneuron {
+
+/** display global mechanism count */
+void write_mech_report() {
+#if NRNMPI
+    /// mechanim count across all gids, local to rank
+    auto n_memb_func = corenrn.get_memb_funcs().size();
```
alexsavulescu, line 14: Can be const:

```suggestion
    const auto n_memb_func = corenrn.get_memb_funcs().size();
```

coreneuron/io/mech_report.cpp
```diff
@@ -0,0 +1,51 @@
...
+    std::vector<unsigned long long> local_mech_count(n_memb_func, 0);
+
+    /// each gid record goes on separate row, only check non-empty threads
+    for (size_t i = 0; i < nrn_nthread; i++) {
+        const auto& nt = nrn_threads[i];
+        std::vector<int> mech_count(n_memb_func, 0);
```
alexsavulescu, line 20: Not used

coreneuron/io/mech_report.cpp
```diff
@@ -0,0 +1,51 @@
...
+    if (nrnmpi_myid == 0) {
+        printf("\n================ MECHANISMS COUNT BY TYPE ==================\n");
+        printf("%4s %20s %10s\n", "Id", "Name", "Count");
+        for (int i = 0; i < total_mech_count.size(); i++) {
+            printf("%4d %20s %10lld\n", i, nrn_get_mechname(i), total_mech_count[i]);
+        }
+        printf("=============================================================\n");
+    }
+#else
+    std::cout << "Build and run with MPI to get mechanism stats \n";
```
alexsavulescu, line 47: Was wondering why this feature should be MPI-only?

coreneuron/io/mech_report.cpp
```diff
@@ -0,0 +1,55 @@
...
+        for (auto* tml = nt.tml; tml; tml = tml->next) {
+            const int type = tml->index;
+            const auto& ml = tml->ml;
+            local_mech_count[type] += ml->nodecount;
+        }
+    }
+
+#if NRNMPI
+    /// get global sum of all mechanism instances
+    std::vector<unsigned long long> total_mech_count(n_memb_func);
```
pramodk, line 28: nitpick / suggestion : initialise `total_mech_count = local_mech_count` outside `#if NRNMPI` and then have below printing code common. (`if (nrnmpi_myid == 0)` can be also common because the `nrnmpi_myid` variable exist with 0).

ohm314: ah, the latter I was wondering about - in this case I can do that

coreneuron/io/mech_report.cpp
ohm314: That's because i had simply cherry-picked Pramod's commit without putting too much thought into it, he didn't need non-mpi for his quick patch and in his initial implementation mpi was actually necessary. but you're right that it should be done properly before merging (and it is trivial to fix now). done.

alexsavulescu (approved):

alexsavulescu:
Ah, one last thing. Maybe we could add `--count_mechs` to an existing test.

ohm314:
thanks for the suggestion, Alex 😄 added it into the ring test.

PR #447, opened 2020-12-25
Setup pnttype2presyn in main thread to avoid race.

nrnhines:
**Phase2 setup of pnttype2presyn causes race.**

Move the setup to main thread prior to phase2.

Fixes neuronsimulator/nrn#884

**How to test this?**

See test-mpithread.py at neuronsimulator/nrn#884 and follow build instructions. Without this change the error reliability is high with:
```
mpiexec -n 4 nrniv -mpi -python test-mpithread.py
```

### DIFF
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -448,6 +448,16 @@ void nrn_setup(const char* filesdat,
     // Note that rank with 0 dataset/cellgroup works fine
     nrn_threads_create(userParams.ngroup <= 1 ? 2 : userParams.ngroup);
 
+    // from nrn_has_net_event create pnttype2presyn for use in phase2.
+    auto& memb_func = corenrn.get_memb_funcs();
+    auto& pnttype2presyn = corenrn.get_pnttype2presyn();
+    auto& nrn_has_net_event_ = corenrn.get_has_net_event();
+    pnttype2presyn.clear();
+    pnttype2presyn.resize(memb_func.size(), -1);
+    for (size_t i = 0; i < nrn_has_net_event_.size(); ++i) {
+        pnttype2presyn[nrn_has_net_event_[i]] = i;
+    }
+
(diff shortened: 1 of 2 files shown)


pramodk:
Retest this please

coreneuron/io/nrn_setup.cpp
```diff
@@ -448,6 +448,16 @@ void nrn_setup(const char* filesdat,
     // Note that rank with 0 dataset/cellgroup works fine
     nrn_threads_create(userParams.ngroup <= 1 ? 2 : userParams.ngroup);
 
+    // from nrn_has_net_event create pnttype2presyn for use in phase2.
+    auto& memb_func = corenrn.get_memb_funcs();
+    auto& pnttype2presyn = corenrn.get_pnttype2presyn();
+    auto& nrn_has_net_event_ = corenrn.get_has_net_event();
+    pnttype2presyn.clear();
+    pnttype2presyn.resize(memb_func.size(), -1);
```
pramodk, line 456: Just to clarify - we clear this because neuron might call this multiple times via direct memory mode right?

pramodk (approved):

coreneuron/io/nrn_setup.cpp
nrnhines: The clear is not strictly necessary as the vector starts out empty and is cleared at the end of nrn_cleanup(). Furthermore, it does not really matter if it is cleared or not as the loop will not change previous values (if previously set) (assuming the manifest of mechanisms is still the same).  Lastly, this more or less is just a minimally modified refactor of what was in phase2 and there is not a clear reason why, instead of a ```std::vector<int>``` it should not be a ```std::map<int, int>``` unless maximum performance in usage is desirable and extra memory use (total number of mechanisms vs number of mechanisms that call net_event.) doesn't matter.

pramodk: Ok. 👍

PR #458, opened 2021-01-08
Print memory use from model_size

ohm314:
**Description**

Small patch cleaning up `model_size` and using it to print memory requirements at the end of `nrn_setup` the print is only done when coreneuron is _not_ run in quiet mode.

**How to test this?**

Just a normal run should show the print

```bash
cmake ..
make -j8
nrnivmodl mod
./bin/nrnivmodl-core mod
./x86_64/special script.py
./x86_64/special-core --tstop=10 --datpath=coredat
```


coreneuron/io/nrn_setup.cpp
```diff
@@ -1026,47 +1034,53 @@ size_t model_size(void) {
...
-        printf("n_input_presyn = %d sz=%ld nbyte=%ld\n", nt.n_input_presyn, sz_psi,
-               nt.n_input_presyn * sz_psi);
-        printf("n_pntproc=%d sz=%ld nbyte=%ld\n", nt.n_pntproc, sz_pp, nt.n_pntproc * sz_pp);
-        printf("n_netcon=%d sz=%ld nbyte=%ld\n", nt.n_netcon, sz_nc, nt.n_netcon * sz_nc);
+        printf("n_presyn = %d sz=%ld nbyte=%ld\n", nt.n_presyn, sz_presyn, nt.n_presyn * sz_presyn);
+        printf("n_input_presyn = %d sz=%ld nbyte=%ld\n", nt.n_input_presyn, sz_input_presyn,
+               nt.n_input_presyn * sz_inputPreSyn);
+        printf("n_pntproc=%d sz=%ld nbyte=%ld\n", nt.n_pntproc, sz_pntproc, nt.n_pntproc * sz_pntproc);
+        printf("n_netcon=%d sz=%ld nbyte=%ld\n", nt.n_netcon, sz_netcon, nt.n_netcon * sz_netcon);
         printf("n_weight = %d\n", nt.n_weight);
```
pramodk, line 1055: @ohm314 : just wondering : last week you added options for `--mech-stats`. What if we change that to `--model-stats` and also move this print memory information like this there? (we can print min / max / avg across ranks for different components)

ohm314: Yeah, i think it's not a bad idea.. I was actually consider the same (which is why i just pushed this as draft for now 😄 )

ohm314: Currently we have a lot of numbers printed in debug. Do we want to move all of those into model-stats? It seems maybe some are a bit redundant and overkill?

ohm314:
That's how the output now looks like:

```
Memory size information for all NrnThreads per rank
------------------------------------------------------------------
                 field          min          max          avg
                 ncell          200          202          201.27
                   end        49547        50801        50216.98
                 nmech           24           24           24.00
                _ndata     15289032     15602344     15476950.00
               _nidata            0            0            0.00
               _nvdata      2487717      2568556      2535105.75
              n_presyn       226321       233670       230629.30
      n_presyn (bytes)     14484544     14954880     14760275.00
        n_input_presyn        31081        32748        31992.47
n_input_presyn (bytes)       745944       785952       767819.38
             n_pntproc       452645       467343       461261.59
     n_pntproc (bytes)      3621160      3738744      3690092.75
              n_netcon       452238       466936       460854.06
      n_netcon (bytes)     18089520     18677440     18434162.00
              n_weight      2261190      2334680      2304270.25
     NrnThread (bytes)    214961108    220213916    218117936.00
    model size (bytes)    243375344    249559432    247077296.00
 Setup Done   : 1.31 seconds
 Model size   : 9.20 GB
```

coreneuron/io/nrn_setup.cpp
```diff
@@ -1026,47 +1043,125 @@ size_t model_size(void) {
...
+                      MPI_COMM_WORLD);
+        for (int i = 0; i < 13; i++) {
+            global_size_data_avg[i] = global_size_data_sum[i] / float(nrnmpi_numprocs);
+        }
+#else
+        global_size_data_max = size_data;
+        global_size_data_min = size_data;
+        for (int i = 0; i < 13; i++) {
+            global_size_data_avg[i] = size_data[i];
+        }
```
pramodk, line 1128: just vector assignment suffice without loop?

pramodk (changes requested):

coreneuron/io/nrn_setup.cpp
```diff
@@ -1026,47 +1043,125 @@ size_t model_size(void) {
...
+        if (nrnmpi_myid == 0) {
+          printf("Memory size information for all NrnThreads per rank\n");
+          printf("------------------------------------------------------------------\n");
+          printf("%22s %12s %12s %12s\n","field", "min", "max", "avg");
+          printf("%22s %12ld %12ld %15.2f\n","ncell",            global_size_data_min[0], global_size_data_max[0], global_size_data_avg[0]);
+          printf("%22s %12ld %12ld %15.2f\n","end",              global_size_data_min[1], global_size_data_max[1], global_size_data_avg[1]);
+          printf("%22s %12ld %12ld %15.2f\n","nmech",            global_size_data_min[2], global_size_data_max[2], global_size_data_avg[2]);
+          printf("%22s %12ld %12ld %15.2f\n","_ndata",           global_size_data_min[3], global_size_data_max[3], global_size_data_avg[3]);
+          printf("%22s %12ld %12ld %15.2f\n","_nidata",          global_size_data_min[4], global_size_data_max[4], global_size_data_avg[4]);
+          printf("%22s %12ld %12ld %15.2f\n","_nvdata",          global_size_data_min[5], global_size_data_max[5], global_size_data_avg[5]);
```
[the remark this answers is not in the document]
nrnhines: I kind of like them to be separate since _nvdata indicates use of arbitrary size Vector data (and maybe some space for Random123). If I remember correctly, the _nidata are mostly integer indices into _ndata (and sometimes just an integer with special semantics).

coreneuron/io/nrn_setup.cpp
ohm314: left over from previous implementation 😄

coreneuron/io/nrn_setup.cpp
```diff
@@ -1026,47 +1043,125 @@ size_t model_size(void) {
...
+          printf("%22s %12ld %12ld %15.2f\n","n_pntproc",         global_size_data_min[8], global_size_data_max[8], global_size_data_avg[8]);
+          printf("%22s %12ld %12ld %15.2f\n","n_pntproc (bytes)", global_size_data_min[8]*sz_pntproc, global_size_data_max[8]*sz_pntproc, global_size_data_avg[8]*sz_pntproc);
+          printf("%22s %12ld %12ld %15.2f\n","n_netcon",          global_size_data_min[9], global_size_data_max[9], global_size_data_avg[9]);
+          printf("%22s %12ld %12ld %15.2f\n","n_netcon (bytes)",  global_size_data_min[9]*sz_netcon, global_size_data_max[9]*sz_netcon, global_size_data_avg[9]*sz_netcon);
+          printf("%22s %12ld %12ld %15.2f\n","n_weight",          global_size_data_min[10], global_size_data_max[10], global_size_data_avg[10]);
+          printf("%22s %12ld %12ld %15.2f\n","NrnThread (bytes)", global_size_data_min[11], global_size_data_max[11], global_size_data_avg[11]);
+          printf("%22s %12ld %12ld %15.2f\n","model size (bytes)",global_size_data_min[12], global_size_data_max[12], global_size_data_avg[12]);
```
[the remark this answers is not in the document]
ohm314: How about for detailed report i leave it in bytes, so one copy this out and have precise figures, and in the one-line output I use larger units when appropriate

coreneuron/io/nrn_setup.cpp
```diff
@@ -559,15 +560,31 @@ void nrn_setup(const char* filesdat,
...
+
+        if (model_size_bytes < 1024) {
+            printf(" Model size   : %ld bytes\n", model_size_bytes);
+        } else if (model_size_bytes  < 1024*1024) {
+            printf(" Model size   : %.2lf kB\n", model_size_bytes/1024.);
+        } else if (model_size_bytes  < 1024*1024*1024) {
+            printf(" Model size   : %.2lf MB\n", model_size_bytes/(1024.*1024.));
+        } else {
+            printf(" Model size   : %.2lf GB\n", model_size_bytes/(1024.*1024.*1024.));
+        }
```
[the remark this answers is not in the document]
ohm314: I think it wont hurt o just leave it like that, it's nice - since it's already done 😄

coreneuron/io/nrn_setup.cpp
ohm314: actually no - one is a float and one is an int vector.. that's why i did not assign.

pramodk: Ahh didn't notice that (by the way, `global_size_data_avg.assign`  would do the same)

pramodk (commented):
with all changes in clang-format, this needs to be rebased now. Might be some pain but one approach would be to just clang-format files changed and then rebase/merge it.

coreneuron/io/nrn_setup.cpp
```diff
@@ -1023,35 +1044,169 @@ size_t model_size(void) {
...
+                      13,
+                      MPI_UNSIGNED_LONG_LONG,
+                      MPI_MAX,
+                      MPI_COMM_WORLD);
+        MPI_Allreduce(&size_data[0],
+                      &global_size_data_sum[0],
+                      13,
+                      MPI_UNSIGNED_LONG_LONG,
+                      MPI_SUM,
+                      MPI_COMM_WORLD);
```
pramodk, line 1101: one more change that I couldn't think last time: could you use existing mpi wrappers from `coreneuron/mpi/mpispike.cpp` like `nrnmpi_long_allreduce_vec`? specially because they use duplicated comm `nrnmpi_comm` instead of hardcoding `MPI_COMM_WORLD`. (in the future they could be different when run coreneuron under NEURON)

coreneuron/io/nrn_setup.cpp
```diff
@@ -1023,35 +1044,169 @@ size_t model_size(void) {
...
+                   global_size_data_max[12],
+                   global_size_data_avg[12]);
+        }
+    }
 
-    nbyte += nrnran123_instance_count() * nrnran123_state_size();
+#if NRNMPI
+    size_t global_nbyte = 0;
+    MPI_Allreduce(&nbyte, &global_nbyte, 1, MPI_UNSIGNED_LONG_LONG, MPI_SUM, MPI_COMM_WORLD);
+    nbyte = global_nbyte;
```
pramodk, line 1208: same here

pramodk (approved):
LGTM

PR #470, opened 2021-01-20
Fix issue with synapse reports

jorblancoa:
**Description**

Fixed issue when passing synapse id to the reporting library 

### DIFF
--- a/coreneuron/io/reports/report_handler.cpp
+++ b/coreneuron/io/reports/report_handler.cpp
@@ -284,10 +284,10 @@ VarsToReport ReportHandler::get_custom_vars_to_report(const NrnThread& nt,
             if ((nodes_to_gids[ml->nodeindices[j]] == gid) && report_variable) {
                 double* var_value =
                     get_var_location_from_var_name(report.mech_id, report.var_name.data(), ml, j);
-                const auto synapse_id = static_cast<int>(
-                    *get_var_location_from_var_name(report.mech_id, SYNAPSE_ID_MOD_NAME, ml, j));
+                double* synapse_id =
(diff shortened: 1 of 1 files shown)


coreneuron/io/reports/report_handler.cpp
```diff
@@ -284,10 +284,10 @@ VarsToReport ReportHandler::get_custom_vars_to_report(const NrnThread& nt,
             if ((nodes_to_gids[ml->nodeindices[j]] == gid) && report_variable) {
                 double* var_value =
                     get_var_location_from_var_name(report.mech_id, report.var_name.data(), ml, j);
-                const auto synapse_id = static_cast<int>(
-                    *get_var_location_from_var_name(report.mech_id, SYNAPSE_ID_MOD_NAME, ml, j));
+                double* synapse_id =
+                    get_var_location_from_var_name(report.mech_id, SYNAPSE_ID_MOD_NAME, ml, j);
                 nrn_assert(synapse_id && var_value);
-                to_report.emplace_back(synapse_id, var_value);
+                to_report.emplace_back(static_cast<int>(*synapse_id), var_value);
```
pramodk, line 290: Unless I am reading this incorrectly, aren't we doing same as previous code here? `double*` was dereferenced and cast to int. So what fixed the bug?

pramodk: thanks for clarifying ! synapse_id can be zero and hence can't be asserted on value!

pramodk (approved):

jorblancoa:
retest this please

pramodk (approved):

PR #473, opened 2021-01-21
Fix issue when passing synapse_id to the reporting

jorblancoa:
**Description**

Assert was failing when synapse_id == 0

- [x] Check if the pointer is not null first and then convert it to int


pramodk (approved):

jorblancoa:
retest this please

PR #475, opened 2021-01-21
Use enum Layout everywhere

alkino:
An enum exists to know that 0 mean SoA and 1 AoS, use it.

I have to move it, so that enum is available everywhere.


pramodk (approved):
LGTM

PR #488, opened 2021-02-02
[Trivial] Fix compilation if git signatures shown by default

olupton:
**Description**

Without this change then the presence of:
```
[log]
  showSignature = True
```
in the local `~/.gitconfig` file breaks compilation of CoreNEURON because the signature information is not escaped properly for inclusion in the generated C++ string literal.

(see also: https://github.com/neuronsimulator/nrn/pull/955)


pramodk (approved):
LGTM

Issue #490, opened 2021-02-05
Update MPI wrappers to be more flexible and handle C++/MPI types correctly


pramodk:
Currently MPI wrappers implemented in [coreneuron/mpi/mpispike.cpp](https://github.com/BlueBrain/CoreNeuron/blob/5a93d9b2b256a12acbaf83085044aba2cc56746f/coreneuron/mpi/mpispike.cpp#L485) have quite some limitations:
* they were derived from old C style code
* for different argument types there are different functions i.e. code duplication (e.g. nrnmpi_dbl_allreduce_vec for double and long argument type)
* doesn't handle type like size_t correctly  (see [this](https://stackoverflow.com/questions/40807833/sending-size-t-type-data-with-mpi))
* MPI Op type is passed as int 1, 2, 3  (quite unintuitive API)
* When #ranks < 2, not all functions avoid calling MPI functions

We should rewrite them to avoid repeated code and address above limitations. I assume @alkino will be interested to re-write this. @olupton  and @ohm314 can provide input if these should be refactored/implemented in certain way.

P.S. the reason for avoiding calling MPI routines for #ranks < 2 is that MPI compiled library can be used in serial or parallel execution. In serial execution NEURON doesn't initialise MPI and libraries like HPE-MPT end-up with floating point exceptions.

ohm314:
Also, keep in mind that currently the passed operation for reductions is not the MPI_Opt code, but another user-defined opreation code that gets translated.

nrnhines:
I don't know if this is relevant to CoreNEURON but one of the reasons for the c api to all this in NEURON is to allow dynamic loading of libmpi (if it exists). This is useful for binary distributions installed on machines where mpi may not be installed, and also be somewhat agnostic about which mpi is installed.

pramodk:
> but one of the reasons for the c api to all this in NEURON is to allow dynamic loading of libmpi (if it exists).

@nrnhines : but even if we implement MPI wrappers in C++ with template parameters (to be type agnostic), we can still compile libnrnmpi.so and that can be dynamically loaded?

nrnhines:
I still have a lot to learn about modern c++.  If that is the case (it wasn't true in the 1990's, and we need to replace all the elaborate macros that were needed because templates didn't work) then go for it.

PR #491, opened 2021-02-05
(Minimum) Changes to avoid floating-point exceptions

pramodk:
* If coreneuron is compiled with MPI but run without --mpi
        flag then coreneuron crashes. This doesn't allow NEURON
        to run existing tests (e.g. using python API)
* Instead of using MPI function, using wrappers from mpispike.cpp
* Updated all reduce wrapper to accept size_t / unsigned long int than just size_t
* Update cast and summary printf

Further cleanup will be done in separate PR, see #490

**How to test this?**

On BB5

```bash
module load unstable hpe-mpi cmake python-dev flex bison cmake gcc
cmake ..
make -j8
make test
```
Specifically look at:

```
ctest -R ring_serial -VV
```

**Test System**
 - Specifically HPE-MPI aka MPT

### DIFF
--- a/coreneuron/io/mech_report.cpp
+++ b/coreneuron/io/mech_report.cpp
@@ -48,7 +48,7 @@ void write_mech_report() {
         printf("\n================ MECHANISMS COUNT BY TYPE ==================\n");
         printf("%4s %20s %10s\n", "Id", "Name", "Count");
         for (size_t i = 0; i < total_mech_count.size(); i++) {
-            printf("%4lu %20s %10lld\n", i, nrn_get_mechname(i), total_mech_count[i]);
+            printf("%4lu %20s %10lu\n", i, nrn_get_mechname(i), total_mech_count[i]);
         }
         printf("=============================================================\n");
     }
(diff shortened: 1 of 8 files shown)


ohm314 (approved):
LGTM

coreneuron/io/mech_report.cpp
```diff
@@ -19,7 +19,7 @@ namespace coreneuron {
 void write_mech_report() {
     /// mechanim count across all gids, local to rank
     const auto n_memb_func = corenrn.get_memb_funcs().size();
-    std::vector<unsigned long long> local_mech_count(n_memb_func, 0);
+    std::vector<size_t> local_mech_count(n_memb_func, 0);
```
olupton, line 22: It seems a bit messy to use `size_t` here and then make `nrnmpi_long_allreduce_vec` expect `unsigned long` when `nrnmpi_long_allreduce` expects `long`.

Might it be easier/more consistent to just use `long` and sidestep issues like whether `size_t*` -> `unsigned long*` is OK, and whether `MPI_UNSIGNED_LONG_LONG` is the correct value for `size_t`?

pramodk: I did somewhat sloppy job here. I started changing stuff and quickly I realised that types are used inconsistently at multiple places. So I end-up creating #491.

> Might it be easier/more consistent to just use long and sidestep issues like whether

Considering these routines are mostly used for statistics printing and `long` should be sufficiently large (on 64bit OS), I will replace size_t with long as you suggested.

olupton (approved):
LGTM!

pramodk:
retest this please

PR #492, opened 2021-02-05
LFP Calculator

bdelmarm:
New LFP calculator class.

Allows for LineSource and PointSource approximations.

A unit test, testing the component, is available in /tests/unit/lfp.

Documentation below:
[LFP.pdf](https://github.com/BlueBrain/CoreNeuron/files/5933094/LFP.pdf)


coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+
+#include "coreneuron/utils/nrn_assert.h"
+#include "coreneuron/mpi/nrnmpidec.h"
+
+namespace coreneuron {
+
+template <typename T>
+using Array3 = std::array<T, 3>;
+
+template <typename Point3D, typename F>
```
castigli, line 19: I am not sure here the template is doing what is supposed to do.
First it seems to me that the template in this method (and the followings) is not necessary (if we can assume that the input needs to be a `Point3D` object and consequently the output is always a `double`).
Instead if you want to keep the template implementation, then I think it should be enabled only for `std::array<F,3>` and for `std::is_floating_point<F>::value == true`.
[Example](https://godbolt.org/z/MzMjbc)

castigli (changes requested):
@bdelmarm , overall it looks good, I have some questions about some of the templates though.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+    return std::sqrt(dot<Point3D, F>(p1, p1));
+}
+
+template <typename Point3D, typename F>
+Array3<F> barycenter(const Point3D& p1, const Point3D& p2) {
+    return {0.5 * (p1[0] + p2[0]), 0.5 * (p1[1] + p2[1]), 0.5 * (p1[2] + p2[2])};
+}
+
+template <typename Point3D, typename F>
+Array3<F> axp(const Point3D& p1, const F alpha, const Point3D& p2) {
```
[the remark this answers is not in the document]
bdelmarm: What about Point3D being a naked pointer?

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+                         const Point3D& seg_0,
+                         const Point3D& seg_1,
+                         const F radius,
+                         const F f) {
+    nrn_assert(radius >= 0.0);
+    Array3<F> dx = axp(seg_1, -1.0, seg_0);
+    Array3<F> de = axp(e_pos, -1.0, seg_0);
+    F dx2(dot<Array3<F>, F>(dx, dx));
+    F dxn(std::sqrt(dx2));
+    if (dxn < 1.0e-20) {
```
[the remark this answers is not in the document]
bdelmarm: Ok.

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
bdelmarm: Ok. I designed the class at the time without knowing the type of classes I would encounter. It would make sense now to remove the template.

coreneuron/io/lfp.hpp
cattabiani: In addition to the previous comment (with which I agree) there is also the fact that we implement everything for points in 3D. With just a few changes we could implement all these functions for points in n dimensions

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
cattabiani: in addition, this function could be split into:

- multiplication by scalar
- point addition

if we implement in the point class these 2 operators we could write the operations in a much nicer way so we do not need to remember, for example, that this function is p1 + a*p2 instead of a*p1 + p2

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+        if (std::abs(q2) < std::numeric_limits<F>::epsilon()) {
+            if (a * b <= 0) {
+                throw std::invalid_argument("Log integral: invalid arguments " + std::to_string(b) +
+                                            " " + std::to_string(a) +
+                                            ". Likely electrode exactly on the segment and "
+                                            "no flooring is present.");
+            }
+            return std::abs(std::log(b / a)) / dxn;
+        } else {
+            return std::log((b + std::sqrt(b * b + q2)) / (a + std::sqrt(a * a + q2))) / dxn;
```
cattabiani, line 98: I will not check this since I already checked last time... except if you changed something here

tests/unit/lfp/lfp.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <iostream>
+
+#include <boost/test/unit_test.hpp>
+#include <boost/test/test_case_template.hpp>
+
+#include "coreneuron/io/lfp.hpp"
+#include "coreneuron/mpi/nrnmpi.h"
+
+template <typename F>
+double integral(F f, double a, double b, int n) {
```
cattabiani, line 13: ~instead of implementing the numerical integral here we could use boost. For example:~
~https://www.boost.org/doc/libs/1_66_0/libs/math/doc/html/math_toolkit/trapezoidal.html~

EDIT: ok, they told me that we do not really have boost in coreneuron. It is used only for testing

tests/unit/lfp/lfp.cpp
```diff
@@ -0,0 +1,104 @@
...
+        area += f(a + (i + 0.5) * step) * step;  // sum up each small rectangle
+    }
+    return area;
+}
+
+std::array<double, 3> operator*(const double x, const std::array<double, 3>& y) {
+    return {x * y[0], x * y[1], x * y[2]};
+}
+
+std::array<double, 3> operator+(const std::array<double, 3>& x, const std::array<double, 3>& y) {
```
cattabiani, line 26: following the previous comment. If you implement the multiplication by scalar the axp function can be eliminated

tests/unit/lfp/lfp.cpp
```diff
@@ -0,0 +1,104 @@
...
+        std::clog << "ANALYTIC line source " << analytic_circling_lfp
+                  << " vs NUMERIC line source LFP " << numeric_circling_lfp << "\n";
+        BOOST_REQUIRE_CLOSE(analytic_circling_lfp, numeric_circling_lfp, 1.0e-6);
+        // TEST of LFP Flooring
+        BOOST_REQUIRE((approaching_elec[1] < 0.866e-6) ? analytic_approaching_lfp == 1.0e6 : true);
+        vals[k] = analytic_circling_lfp;
+    }
+    // TEST of SYMMETRY of LFP FORMULA
+    for (size_t k = 0; k < 5; k++) {
+        BOOST_REQUIRE(std::abs((vals[k] - vals[k + 5]) * 0.5 / (vals[k] + vals[k + 5])) < 1.0e-12);
```
cattabiani, line 79: this is not a norm and we need to take care of division by 0. I would suggest to write:

```suggestion
        BOOST_REQUIRE(std::abs(vals[k] - vals[k + 5]) < 2.0e-12 * max(abs(vals[k], abs(vals[k+5])));
```

or something similar

tests/unit/lfp/lfp.cpp
```diff
@@ -0,0 +1,104 @@
...
+#if NRNMPI
+    nrnmpi_init(nullptr, nullptr);
+#endif
+    double segment_length{1.0e-6};
+    double segment_start_val{1.0e-6};
+    std::array<double, 3> segment_start = std::array<double, 3>{0.0, 0.0, segment_start_val};
+    std::array<double, 3> segment_end = segment_start +
+                                        std::array<double, 3>{0.0, 0.0, segment_length};
+    double floor{1.0e-6};
+    double pi{3.141592653589};
```
cattabiani, line 41: `PI` is passed as global variable by neuron. Probably you can use it

cattabiani (changes requested):
Good job! Maybe we can generalize a little the code to take care of also non-3D points (just in case one day someone comes around and needs points in 2D)

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+                  const Point3Ds& electrodes,
+                  double extra_cellular_conductivity)
+        : segment_ids_(segment_ids) {
+        if (seg_start.size() != seg_end.size()) {
+            throw std::invalid_argument("Different number of segment starts and ends.");
+        }
+        if (seg_start.size() != radius.size()) {
+            throw std::invalid_argument("Different number of segments and radii.");
+        }
+        double f(1.0 / (extra_cellular_conductivity * 4.0 * boost::math::constants::pi<double>()));
```
iomaganaris, line 159: As @cattabiani mentioned below  to be exact and use the `pi` value everywhere, `pi` should take the value which is read from `NEURON` instead of the constant value provided either by `boost` or `math.h`. I can work on that.

iomaganaris (changes requested):

coreneuron/io/lfp.hpp
iomaganaris: I created related PR #494

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
castigli: I think the idea of having one function combined is for speed, possibly premature optimization, but is it a very common operation and to me it makes sense to combine it. The name of the function tells you the order of the operations `paxpy` p(type) a(alpha) x(first input) p(plus) y(second input), basically follows blas naming convention.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+F norm(const Point3D& p1) {
+    return std::sqrt(dot(p1, p1));
+}
+
+Point3D barycenter(const Point3D& p1, const Point3D& p2) {
+    return {0.5 * (p1[0] + p2[0]), 0.5 * (p1[1] + p2[1]), 0.5 * (p1[2] + p2[2])};
+}
+
+Point3D paxp(const Point3D& p1, const F alpha, const Point3D& p2) {
```
castigli, line 31: `paxpy`?

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
bdelmarm: I followed @castigli advice here.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,226 @@
...
+
+#include <boost/math/constants/constants.hpp>
+
+#include "coreneuron/utils/nrn_assert.h"
+#include "coreneuron/mpi/nrnmpidec.h"
+
+namespace coreneuron {
+
+template <typename T>
+using Array3 = std::array<T, 3>;
```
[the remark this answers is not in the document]
bdelmarm: I removed the template altogether.

coreneuron/io/lfp.hpp
bdelmarm: I will leave other dimensions for another PR if that is ok.

coreneuron/io/lfp.hpp
bdelmarm: There has not been material changes since your review I believe.

tests/unit/lfp/lfp.cpp
bdelmarm: Ok.

tests/unit/lfp/lfp.cpp
bdelmarm: I removed the function entirely, relying instead, for ease, on paxpy.

tests/unit/lfp/lfp.cpp
bdelmarm: Done.

tests/unit/lfp/lfp.cpp
bdelmarm: Ok. Thanks.

coreneuron/io/lfp.hpp
bdelmarm: @iomaganaris , following my latest commit, pi is apparently equal to 0. Probably a merge issue on my side. Can you take a look? I restored a hardcoded value.

coreneuron/io/lfp.hpp
bdelmarm: Done.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+    Point3D de_star(paxpy(de, -mu, dx));
+    F de_star2(dot(de_star, de_star));
+    F q2(de_star2 / dx2);
+
+    F delta(mu * mu - (de2 - radius * radius) / dx2);
+    F one_m_mu(1.0 - mu);
+    auto log_integral = [&q2, &dxn](F a, F b) {
+        if (q2 < std::numeric_limits<F>::epsilon()) {
+            if (a * b <= 0) {
+                throw std::invalid_argument("Log integral: invalid arguments " + std::to_string(b) +
```
tristan0x, line 85: You should use `std::ostringstream` instead of using string concatenation.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
+#ifndef LFP_H
+#define LFP_H
+
+#include <cmath>
+#include <array>
+#include <iostream>
```
tristan0x, line 6: Including `iostream` in a header is usually not a good sign.
Either it is not used, or the code that needs it should be in a compilation unit, not a header.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+                                        const Point3D& seg_0,
+                                        const Point3D& seg_1,
+                                        const F radius,
+                                        const F f) const {
+    return point_source_lfp_factor(e_pos, barycenter(seg_0, seg_1), radius, f);
+}
+
+};  // namespace coreneuron
+
+#endif  // AREA_LFP_H
```
tristan0x, line 217: comment doesn't match the macro name in prelude. You should use `#pragma once` anyway.

tristan0x (commented):
Nice addition!
Sorry don't have time to look into details. Will do this afternoon. Here is a preliminary feedback about the general form on the change.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
+#ifndef LFP_H
```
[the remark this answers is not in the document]
bdelmarm: Done.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "coreneuron/nrnconf.h"
+#include "coreneuron/utils/nrn_assert.h"
+#include "coreneuron/mpi/nrnmpidec.h"
+
+namespace coreneuron {
+
+using F = double;
+using Point3D = std::array<F, 3>;
+using Point3Ds = std::vector<Point3D>;
```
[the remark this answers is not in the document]
bdelmarm: Added an lfputils namespace.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+F dot(const Point3D& p1, const Point3D& p2) {
```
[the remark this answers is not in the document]
bdelmarm: Done.

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * \tparam Point3D Type of 3D point
+ * \tparam F Floating point type
+ * \param e_pos electrode position
+ * \param seg_pos segment position
+ * \param radius segment radius
+ * \param f conductivity factor 1/([4 pi] * [conductivity])
+ * \return Resistance of the medium from the segment to the electrode.
+ */
+F line_source_lfp_factor(const Point3D& e_pos,
```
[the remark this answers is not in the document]
bdelmarm: Done.

coreneuron/io/lfp.hpp
bdelmarm: Done.

coreneuron/io/lfp.hpp
bdelmarm: Done.

coreneuron/io/lfp.hpp
bdelmarm: Done.

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
castigli: I know it made sense in the previous implementation, but I would remove ` using F = double`  and `using Vec = double*`, it seems just a bit confounding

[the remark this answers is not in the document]
bdelmarm: Ok. I'll keep the typedef double*, perhaps renaming Vec, DoublePtr, cause I get a strange compilation otherwise.

castigli (commented):
It looks good do me, I am not approving just because there are a couple of points that others (@tristan0x @cattabiani ) should confirm that they are satisfied with.

coreneuron/io/lfp.cpp
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+    std::vector<double> res_reduced(res.size());
+#if NRNMPI
+    int mpi_sum{1};
+    nrnmpi_dbl_allreduce_vec(res.data(), res_reduced.data(), res.size(), mpi_sum);
+#else
+    res_reduced = res;
+#endif
+    lfp_values = res_reduced;
```
tristan0x, line 118: This chunk of code makes a lot of unnecessary vector allocations and copies. I do not understand why variable `res_reduced` is necessary, and we should use `std::swap` when not using MPI instead of doing a vector copy, something like below should work:

```diff
diff --git a/coreneuron/io/lfp.cpp b/coreneuron/io/lfp.cpp
index 00716fd..f4812c1 100644
--- a/coreneuron/io/lfp.cpp
+++ b/coreneuron/io/lfp.cpp
@@ -108,14 +108,14 @@ inline void LFPCalculator<Type, SegmentIdTy>::lfp(const Vector& membrane_current
             res[k] += ms[l] * membrane_current[segment_ids_[l]];
         }
     }
-    std::vector<double> res_reduced(res.size());
+
 #if NRNMPI
     int mpi_sum{1};
-    nrnmpi_dbl_allreduce_vec(res.data(), res_reduced.data(), res.size(), mpi_sum);
+    lfp_values.resize(res.size());
+    nrnmpi_dbl_allreduce_vec(res.data(), lfp_values.data(), res.size(), mpi_sum);
 #else
-    res_reduced = res;
+    std::swap(res, lfp_values);
 #endif
-    lfp_values = res_reduced;
 }
```

Also the `res` variable is not necessary. `lfp_values` should be used directly instead. (assuming that `nrnmpi_dbl_allreduce_vec` supports in-place operation`. @pramodk ?


In term of design, it seems to be the LFPCalculator class should not have a `lfp_values` member variable. IMO the `lfp` member method should be const and returns the result directly. (The class can also be a functor by renaming the `lfp` member function by `std::vector<double> operator()(const Vector& membrane_current) const` function.

tristan0x (changes requested):
All symbols inside in `coreneuron::lfputils` are called either in lfp.cpp or in `LFPCalculator::getFactor` which is a private member function only called in the compilation unit lfp.cpp. So the user is not concerned by it and there is not benefits in term of performance to have them in the header. This is why I believe they should all be moved to the compilation unit.

I also made a comment regarding performance improvement in LFPCalculator::lfp member method.

tests/unit/alignment/alignment.cpp
```diff
@@ -14,7 +14,6 @@
 #include <stdint.h>
 
 #include <boost/test/unit_test.hpp>
-#include <boost/test/test_case_template.hpp>
 #include <boost/mpl/list.hpp>
```
cattabiani, line 17: maybe lexicographic sort?

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+F norm(const Point3D& p1) {
+    return std::sqrt(dot(p1, p1));
+}
+
+Point3D barycenter(const Point3D& p1, const Point3D& p2) {
+    return {0.5 * (p1[0] + p2[0]), 0.5 * (p1[1] + p2[1]), 0.5 * (p1[2] + p2[2])};
+}
+
+Point3D paxpy(const Point3D& p1, const F alpha, const Point3D& p2) {
```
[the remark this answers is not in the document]
bdelmarm: A 

> All symbols inside in `coreneuron::lfputils` are called either in lfp.cpp or in `LFPCalculator::getFactor` which is a private member function only called in the compilation unit lfp.cpp. So the user is not concerned by it and there is not benefits in term of performance to have them in the header. This is why I believe they should all be moved to the compilation unit.
> 
> I also made a comment regarding performance improvement in LFPCalculator::lfp member method.

The LFP test in Coreneuron tests functions in lfputils. I would rather keep things as they are.

tests/unit/lfp/lfp.cpp
```diff
@@ -0,0 +1,103 @@
...
+#if NRNMPI
+    nrnmpi_init(nullptr, nullptr);
+#endif
+    double segment_length{1.0e-6};
+    double segment_start_val{1.0e-6};
+    std::array<double, 3> segment_start = std::array<double, 3>{0.0, 0.0, segment_start_val};
+    std::array<double, 3> segment_end =
+        paxpy(segment_start, 1.0, std::array<double, 3>{0.0, 0.0, segment_length});
+    double floor{1.0e-6};
+    double pi = 3.141592653589;
```
cattabiani, line 35: Following the discussion between @iomaganaris and @bdelmarm : how is it going? Is it still `pi == 0`?

tests/unit/queueing/test_queueing.cpp
```diff
@@ -10,7 +10,6 @@
 #define TYPE              T::cont
 
 #include <boost/test/unit_test.hpp>
-#include <boost/test/test_case_template.hpp>
 #include <boost/filesystem.hpp>
 
 #include <cstdlib>
```
cattabiani, line 15: naming ordering?

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,141 @@
...
+                  const lfputils::Point3Ds& seg_end,
+                  const std::vector<double>& radius,
+                  const std::vector<SegmentIdTy>& segment_ids,
+                  const lfputils::Point3Ds& electrodes,
+                  double extra_cellular_conductivity);
+
+    template <typename Vector>
+    void lfp(const Vector& membrane_current);
+
+    std::vector<double> lfp_values;
```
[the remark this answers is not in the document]
bdelmarm: Ok.

cattabiani (commented):
LGTM! Just some ordering and the value of pi

coreneuron/io/lfp.cpp
bdelmarm: Ok. res variable is removed. 

lfp values need to be cached somehow within the simulation and read subsequently. 
It could be done elsewhere. I would think that for now it is ok to store them within the simulator.

tests/unit/lfp/lfp.cpp
bdelmarm: As far as I could test, pi was indeed 0. @iomaganaris can you take a look again?

coreneuron/io/lfp.hpp
```diff
@@ -91,15 +91,17 @@ struct LFPCalculator {
     template <typename Vector>
     void lfp(const Vector& membrane_current);
 
-    std::vector<double> lfp_values;
+    const std::vector<double> lfp_values() const noexcept {
```
tristan0x, line 94: You should avoid copying the vector.
```suggestion
    const std::vector<double>& lfp_values() const noexcept {
```

tristan0x (changes requested):

coreneuron/io/lfp.hpp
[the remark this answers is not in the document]
tristan0x: > The LFP test in Coreneuron tests functions in lfputils. I would rather keep things as they are.

That is indeed reasonable enough to expose these symbols.

tests/unit/alignment/alignment.cpp
bdelmarm: Done.

tests/unit/lfp/lfp.cpp
iomaganaris: I think I fixed this. Can you have look and see if you can reproduce the previous issue?

tristan0x:
> All symbols inside in `coreneuron::lfputils` are called either in lfp.cpp or in `LFPCalculator::getFactor` which is a private member function only called in the compilation unit lfp.cpp. So the user is not concerned by it and there is not benefits in term of performance to have them in the header. This is why I believe they should all be moved to the compilation unit.

Following discussion with @bdelmarm , these symbols need to be exposed to be unit-tested.

tristan0x:
> I also made a comment regarding performance improvement in LFPCalculator::lfp member method.

adressed in f80818a53320a190bcac4a0e625d33995832d876

tristan0x (approved):

tests/unit/lfp/lfp.cpp
bdelmarm: Looks fine now. CI is passing.

tests/unit/lfp/lfp.cpp
```diff
@@ -32,7 +32,7 @@ BOOST_AUTO_TEST_CASE(LFP_PointSource_LineSource) {
     std::array<double, 3> segment_end =
         paxpy(segment_start, 1.0, std::array<double, 3>{0.0, 0.0, segment_length});
     double floor{1.0e-6};
-    double pi = 3.141592653589;
+    pi = 3.141592653589;
```
cattabiani, line 35: I do not understand: why is there still this hard-coded value?

cattabiani (changes requested):
I am lost about pi. It is still hard-coded. I thought we were going to get it from neuron...

castigli (approved):

tests/unit/lfp/lfp.cpp
iomaganaris: This is still hardcoded in the unit test because for this unit test we don't run `NEURON` to get it from there

cattabiani (approved):
LGTM

pramodk (approved):

coreneuron/io/lfp.hpp
```diff
@@ -0,0 +1,143 @@
...
+enum LFPCalculatorType { LineSource, PointSource };
+
+/**
+ * \brief LFPCalculator allows calculation of LFP given membrane currents.
+ */
+template <LFPCalculatorType Ty, typename SegmentIdTy = int>
+struct LFPCalculator {
+    /**
+     * LFP Calculator constructor
+     * \param comm MPI communicator
```
iomaganaris, line 75:
```suggestion
```
Since there is no `comm` parameter anymore, this can be removed?

iomaganaris (approved):
LGTM! One last small thing

PR #494, opened 2021-02-10
Pass PI value from NEURON to CoreNEURON

iomaganaris:
**Description**

In the `lfp` PR during the calculation of `lfp` there is a variable `PI` used. The value of this variable should be same as `NEURON` and the variable used by the mod files.

- [x] Created new pi global parameter which is read by NEURON `globals.dat` or directly NEURON in online mode
- [x] Use this pi in lfp calculation and test
- [x] Removed a deprecated boost test header file

### DIFF
--- a/coreneuron/io/global_vars.cpp
+++ b/coreneuron/io/global_vars.cpp
@@ -48,6 +48,7 @@ void set_globals(const char* path, bool cli_global_seed, int cli_global_seed_val
     (*n2v)["celsius"] = PSD(0, &celsius);
     (*n2v)["dt"] = PSD(0, &dt);
     (*n2v)["t"] = PSD(0, &t);
+    (*n2v)["PI"] = PSD(0, &pi);
 
     if (corenrn_embedded) {  // CoreNEURON embedded, get info direct from NEURON
 
--- a/tests/unit/lfp/lfp.cpp
+++ b/tests/unit/lfp/lfp.cpp
@@ -4,10 +4,11 @@
 #include <iostream>
 
 #include <boost/test/unit_test.hpp>
-#include <boost/test/test_case_template.hpp>
 
 #include "coreneuron/io/lfp.hpp"
 #include "coreneuron/mpi/nrnmpi.h"
+
+using namespace coreneuron;
 
 template <typename F>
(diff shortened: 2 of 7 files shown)


coreneuron/io/global_vars.cpp
```diff
@@ -48,6 +48,7 @@ void set_globals(const char* path, bool cli_global_seed, int cli_global_seed_val
     (*n2v)["celsius"] = PSD(0, &celsius);
     (*n2v)["dt"] = PSD(0, &dt);
     (*n2v)["t"] = PSD(0, &t);
+    (*n2v)["PI"] = PSD(0, &pi);
```
pramodk, line 51: @iomaganaris : without looking into details, in the file mode everything work fine? i.e. PI is passed?

iomaganaris: Yes, I tried printing `pi` with both on and off line mode and it worked fine

tests/unit/lfp/lfp.cpp
```diff
@@ -38,7 +38,7 @@ BOOST_AUTO_TEST_CASE(LFP_PointSource_LineSource) {
     std::array<double, 3> segment_end = segment_start +
                                         std::array<double, 3>{0.0, 0.0, segment_length};
     double floor{1.0e-6};
-    double pi{3.141592653589};
+    pi = 3.141592653589;
```
cattabiani, line 41: why don't we get pi from neuron here?

iomaganaris: `CoreNEURON` reads its value from either `NEURON` in on line mode or `CoreNEURON` dataset which is generated by `NEURON` (`globals.dat` file). Since in the unit tests we don't run `NEURON` before, we can't get this value from `NEURON`. `pi` is a global variable in the `coreneuron` namespace, so I thought it would be enough to set it manually here.

pramodk (approved):

pramodk:
@bdelmarm : feel free to merge this PR (it will be merged in your PR).

bdelmarm (approved):
Great @iomaganaris . Thank you!

PR #496, opened 2021-02-19
Prevent nrnivmodl running on every build

olupton:
**Description**

This changes how dependencies are declared in CMake to ensure that `nrnivmodl` is not called to recompile `halfgap.mod` ever time the project is built.

**How to test this?**

I tested with CoreNEURON built as a submodule of NEURON. After running
```bash
cmake ..
make -j
make -j # this should barely do anything, but without it this PR it was rerunning nrnivmodl.
```

### DIFF
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -157,16 +157,16 @@ set_target_properties(
 # =============================================================================
 set(modfile_directory "${CORENEURON_PROJECT_SOURCE_DIR}/tests/integration/ring_gap/mod")
 file(GLOB modfiles "${modfile_directory}/*.mod")
-set(special_name "${CMAKE_BINARY_DIR}/bin/${CMAKE_SYSTEM_PROCESSOR}/special")
-set(output_binaries "${special}" "${special_name}-core")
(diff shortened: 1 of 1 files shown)


pramodk (approved):
LGTM

I checked that following rebuilds the target as expected:

```
$ touch ../tests/integration/ring_gap/mod/halfgap.mod
$ make
```

But if I changed `share/modfile/hh.mod` (e.g. add empty line), I don't see make redoing anything. I think this is existing behaviour (bug) but we could add dependency to mod files under `build-dir/share/modfile/*.mod`. (in a separate PR is fine)

olupton:
Good spot! I agree the implicit dependency on the builtin modfiles was not captured.
I pushed two fixes, unfortunately in two repositories:
- in this PR, I added a rule expressing that `{build}/share/modfile/foo.mod` depends on `{source}/coreneuron/mechanism/mech/modfile/foo.mod`, and I now export a cache variable `CORENEURON_BUILTIN_MODFILES` containing all the `{build}/share/modfile/*.mod` paths. The explicit invocation of `nrnivmodl-core` in CoreNEURON now depends on these files.
- in https://github.com/neuronsimulator/nrn/pull/1011 the `nrnivmodl` rules generated by `nrn_add_test` now depend on CoreNEURON's builtin modfiles if `-coreneuron` is passed to `nrnivmodl`

pramodk (approved):

PR #498, opened 2021-02-24
Fix GPU compilation issue when nrn_ghk is used in the MOD files

pramodk:
**Description**

  - nrn_ghk was recently added but `acc routine seq` was missing
  - nrn_ghk uses celsius which is an extern variable. All extern
    variables require `acc declare create/copyin` caluse.
  - As global variables can be initialized after device attachment
    (celsius in this case), we need to make sure celsius is updated
    on the device.
  - Replacing `data` clause with `update device` is sufficient. In
    this case `data` clause doesn't update value on device when
    checked in eion.cpp.

**How to test this?**

```bash
cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install -DCORENRN_ENABLE_GPU=ON
make -j8
make install

# in neuron repo
cd nrn/test/pynrn/
nrnivmodl-core .
```

**Test System**
 - BB5
 - Compiler: nvhpc/20.9 and CUDA 11
 - Backend: [e.g. GPU]

### DIFF
--- a/coreneuron/mechanism/eion.cpp
+++ b/coreneuron/mechanism/eion.cpp
@@ -189,6 +189,10 @@ the USEION statement of any model using this ion\n",
 #define FARADAY     _faraday_codata2018
 #define gasconstant _gasconstant_codata2018
 #endif
+
+// extern variables require acc declare
+#pragma acc declare create(celsius)
+
 #define ktf (1000. * gasconstant * (celsius + 273.15) / FARADAY)
 
 double nrn_nernst(double ci, double co, double z, double celsius) {
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -488,7 +488,8 @@ extern "C" int run_solve_core(int argc, char** argv) {
     bool compute_gpu = corenrn_param.gpu;
     bool skip_mpi_finalize = corenrn_param.skip_mpi_finalize;
 
-// clang-format off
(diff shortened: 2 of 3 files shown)


coreneuron/mechanism/eion.cpp
```diff
@@ -189,6 +189,10 @@ the USEION statement of any model using this ion\n",
 #define FARADAY     _faraday_codata2018
 #define gasconstant _gasconstant_codata2018
 #endif
+
+// extern variables require acc declare
+#pragma acc declare create(celsius)
```
iomaganaris, line 194: I'm wondering if this should be moved in `register_mech.cpp` and do the same for `secondorder` as well? https://github.com/BlueBrain/CoreNeuron/blob/14ad572b5df4a67df419840261e4aa3893dee54b/coreneuron/mechanism/register_mech.cpp#L25  Or there is some issue with moving that there?

iomaganaris: Replied offline:
* at the definition site, every variable needs to have acc declare (copyin or create depending on initial value)
* at any other compilation unit, if variable is extern then it should also have acc declare
https://docs.nvidia.com/hpc-sdk/pgi-compilers/18.10/x86/openacc-gs/index.htm#openacc-globals

iomaganaris (approved):

olupton (approved):
LGTM!

PR #500, opened 2021-02-25
Add GitLab CI pipeline.

olupton:
This introduces a GitLab CI pipeline that runs on BBP compute resources. 

This initial version:
- builds CoreNEURON.
- runs the CoreNEURON CTest suite.
- builds NEURON, using the build of CoreNEURON as an external dependency.
- runs the NEURON CTest suite.

Where everything is build/run on CPU using the Intel compiler.

To make this work, some fixes were needed:
- more variables describing the CoreNEURON build configuration are now exported to the installed `coreneuron-config.cmake` file.
- the number of processors required by some CoreNEURON tests is now declared.


.gitlab-ci.yml
```diff
@@ -0,0 +1,70 @@
...
+  variables:
+    SPACK_COMPILER: intel
+.spack_coreneuron:
+  variables:
+    SPACK_PACKAGE: coreneuron
+    SPACK_SPEC: +debug+tests~legacy-unit
+.spack_neuron:
+  variables:
+    SPACK_PACKAGE: neuron
+    # TODO: need a way of configuring this, maybe from the PR description?
```
pramodk, line 38: We have use this approach already for Jenkins so same can be used?

pramodk (commented):
Apart from todos, this looks good to me.

.gitlab-ci.yml
olupton: I think we can re-use some commands from Jenkins, but it will also need some new plumbing.

This can be addressed later. In the short term we can just modify `.gitlab-ci.yml` in open PRs that are waiting for changes to be merged into NEURON, as was done here.

olupton: I opened #505 to track this.

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
+include:
+  - project: hpc/gitlab-pipelines
+    file: spack-build-components.gitlab-ci.yml
```
pramodk, line 3: q : when some some changes these common template file, are dependent CI projects are triggered so that there is no regression?

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+# Configure when the pipeline actually runs.
+workflow:
+  rules:
+    # Don't run on push events for now to try and save resources
+    - if: '$CI_PIPELINE_SOURCE == "push"'
+      when: never
+    # Don't run on MR events because we also get "external_pull_request_event" from GitHub.
+    - if: '$CI_PIPELINE_SOURCE == "merge_request_event"'
+      when: never
+    - when: always
```
pramodk, line 14: just curious : this would be the case for all GitHub projects I assume? if so, could this be added somewhere else and inherited from there?

pramodk (approved):
LGTM

.gitlab-ci.yml
olupton: Good question. It doesn't happen automatically, but @matz-e already set up the CI of `hpc/gitlab-pipelines` to trigger the CI of `hpc/touchdetector` as a regression test. Once this PR is merged I think we should add CoreNEURON's CI alongside TouchDetector's to improve the test coverage.

.gitlab-ci.yml
olupton: Good suggestion, done in c1fd1dd9fd41e4588088265bf0da9e41578f34e2 -- unfortunately this introduces another TODO; wait for an external MR to be merged.

iomaganaris (approved):

PR #514, opened 2021-03-29
Summation reports

jorblancoa:
**Description**

Add support for summation reports.

### DIFF
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -852,10 +852,6 @@ void nrn_cleanup() {
             delete ((NrnThreadMappingInfo*) nt->mapping);
         }
 
-        if (nt->alu_) {
-            delete ((ALUMapping*) nt->alu_);
-        }
-
         free_memory(nt->_ml_list);
 
         if (nt->nrn_fast_imem) {
@@ -924,7 +920,6 @@ void read_phase3(NrnThread& nt, UserParams& userParams) {
 
     /** mapping information for all neurons in single NrnThread */
     NrnThreadMappingInfo* ntmapping = new NrnThreadMappingInfo();
-    ALUMapping* alu = new ALUMapping();
 
     int count = 0;
 
@@ -957,7 +952,7 @@ void read_phase3(NrnThread& nt, UserParams& userParams) {
 
     // set pointer in NrnThread
     nt.mapping = (void*) ntmapping;
-    nt.alu_ = (void*) alu;
+    nt.summation_report_handler_ = std::make_unique<SummationReportMapping>();
 }
 
 static size_t memb_list_size(NrnThreadMembList* tml) {
--- a/coreneuron/io/nrn_setup.cpp
+++ b/coreneuron/io/nrn_setup.cpp
@@ -36,6 +36,7 @@
 #include "coreneuron/io/mech_report.h"
 #include "coreneuron/apps/corenrn_parameters.hpp"
 #include "coreneuron/io/nrn_setup.hpp"
+#include "coreneuron/io/reports/nrnreport.hpp"
 
 // callbacks into nrn/src/nrniv/nrnbbcore_write.cpp
 #include "coreneuron/sim/fast_imem.hpp"
@@ -851,6 +852,10 @@ void nrn_cleanup() {
             delete ((NrnThreadMappingInfo*) nt->mapping);
         }
 
+        if (nt->alu_) {
+            delete ((ALUMapping*) nt->alu_);
+        }
+
         free_memory(nt->_ml_list);
 
         if (nt->nrn_fast_imem) {
@@ -919,6 +924,7 @@ void read_phase3(NrnThread& nt, UserParams& userParams) {
 
     /** mapping information for all neurons in single NrnThread */
     NrnThreadMappingInfo* ntmapping = new NrnThreadMappingInfo();
+    ALUMapping* alu = new ALUMapping();
 
     int count = 0;
 
@@ -951,6 +957,7 @@ void read_phase3(NrnThread& nt, UserParams& userParams) {
 
     // set pointer in NrnThread
     nt.mapping = (void*) ntmapping;
+    nt.alu_ = (void*) alu;
 }
 
 static size_t memb_list_size(NrnThreadMembList* tml) {
--- a/coreneuron/io/reports/report_configuration_parser.cpp
+++ b/coreneuron/io/reports/report_configuration_parser.cpp
@@ -40,7 +40,7 @@ enum class TargetType {
 };
 
 /*
- * Split filter string ("mech.var_name") into mech_id and var_name
+ * Split filter comma separated strings ("mech.var_name") into mech_name and var_name
  */
 void parse_filter_string(const std::string& filter, ReportConfiguration& config) {
     std::istringstream in(filter);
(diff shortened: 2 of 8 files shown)


olupton:
I think the GitLab CI is failing because this branch is behind `master`.

pramodk:
@jorblancoa : most likely we will tag 1.0b with latest master. So all of your changes can go on top of master (in case you are working on 1.0a)

jorblancoa:
@pramodk Perfect! will do that

coreneuron/io/reports/report_event.cpp
```diff
@@ -22,25 +23,44 @@ namespace coreneuron {
...
     }
     std::sort(gids_to_report.begin(), gids_to_report.end());
 }
 
+void ReportEvent::summation_ALU(NrnThread* nt) {
+    if (static_cast<int>(step) % reporting_period == 0) {
+        auto* alu_mapping = static_cast<ALUMapping*>(nt->alu_);
+        auto& alu = alu_mapping->report_ALU_[report_path];
+        double sum = 0.0;
+        for (const auto& kv: alu.currents_) {
```
pramodk, line 48: nitpick : add here comment about why we need to do this / implementation details
(Q. do we also sum values from all compartments in a section into a single value?)

coreneuron/io/reports/report_handler.cpp
```diff
@@ -253,6 +263,86 @@ VarsToReport ReportHandler::get_section_vars_to_report(const NrnThread& nt,
...
+                  << nt.ncell << '\n';
+        nrn_abort(1);
+    }
+    for (int i = 0; i < nt.ncell; i++) {
+        int gid = nt.presyns[i].gid_;
+        if (report.target.find(gid) == report.target.end()) {
+            continue;
+        }
+        bool has_imembrane = false;
+        int scalar = 1;
```
pramodk, line 286: nitpick : what the scalar mean here?

pramodk (changes requested):
overall looks ok but take a look at the comments/possible improvements.

coreneuron/sim/multicore.hpp
```diff
@@ -129,8 +129,9 @@ struct NrnThread: public MemoryManaged {
     int _net_send_buffer_cnt = 0;
     int* _net_send_buffer = nullptr;
 
-    int* _watch_types = nullptr;                   /* nullptr or 0 terminated array of integers */
-    void* mapping = nullptr;                       /* section to segment mapping information */
+    int* _watch_types = nullptr; /* nullptr or 0 terminated array of integers */
+    void* mapping = nullptr;     /* section to segment mapping information */
+    void* alu_ = nullptr;        /* report to ALU (values of the current summation */
```
olupton, line 134: Can we forward declare `struct ALUMapping;` and avoid this use of `void*`?

It would be nicer to use `std::unique_ptr<ALUMapping>` and remove some explicit `new`/`delete`, but that might take a little more reorganisation(?) to make the full type visible.

coreneuron/io/reports/report_event.cpp
```diff
@@ -22,25 +23,44 @@ namespace coreneuron {
...
+                         double report_dt)
     : dt(dt)
     , tstart(tstart)
-    , report_path(name) {
+    , report_path(name)
+    , report_dt(report_dt) {
     VarsToReport::iterator it;
     nrn_assert(filtered_gids.size());
     step = tstart / dt;
+    reporting_period = static_cast<int>(report_dt / dt);
```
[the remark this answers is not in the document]
jorblancoa: but this is exactly what it is, the number of timesteps after the variable will be recorded. When this is multiple of the current step we do the summation (in any other case it doesn't make sense since it wont be recorded).

coreneuron/io/nrn_setup.cpp
```diff
@@ -951,6 +957,7 @@ void read_phase3(NrnThread& nt, UserParams& userParams) {
 
     // set pointer in NrnThread
     nt.mapping = (void*) ntmapping;
+    nt.alu_ = (void*) alu;
```
olupton, line 960: I don't think this cast is needed.

coreneuron/io/reports/report_configuration_parser.cpp
```diff
@@ -40,13 +40,34 @@ enum class TargetType {
...
 /*
- * Split filter string ("mech.var_name") into mech_id and var_name
+ * Split filter comma separated strings ("mech.var_name") into mech_name and var_name
  */
 void parse_filter_string(const std::string& filter, ReportConfiguration& config) {
-    std::istringstream iss(filter);
-    std::string token;
-    std::getline(iss, config.mech_name, '.');
-    std::getline(iss, config.var_name, '.');
+    std::istringstream in(filter);
```
olupton, line 46: Unused?

coreneuron/io/reports/report_configuration_parser.cpp
```diff
@@ -40,13 +40,34 @@ enum class TargetType {
...
+
+    std::vector<std::string> mechanisms;
+    std::stringstream ss(filter);
+    std::string mechanism;
+    // Separate in case of more than 1 report_on
+    while (getline(ss, mechanism, ',')) {
```
olupton, line 52: I don't immediately see why we need two loops here.

coreneuron/io/reports/report_handler.cpp
```diff
@@ -262,7 +352,10 @@ VarsToReport ReportHandler::get_custom_vars_to_report(const NrnThread& nt,
         if (report.target.find(gid) == report.target.end()) {
             continue;
         }
-        Memb_list* ml = nt._ml_list[report.mech_id];
+        // There can only be 1 mechanism
```
olupton, line 355: Perhaps this should be asserted?

olupton (commented):
I added some minor comments on the code, but I admit I am not super familiar with this part of the codebase. Also :+1: on @pramodk's comment about the "ALU" naming, I don't find it very descriptive.

olupton (approved):
LGTM!

coreneuron/io/reports/report_configuration_parser.cpp
```diff
@@ -40,13 +40,34 @@ enum class TargetType {
...
+    // Split mechanisms
+    for (int i = 0; i < mechanisms.size(); ++i) {
+        std::string mech_name;
+        std::string var_name;
+        std::istringstream iss(mechanisms[i]);
+        std::getline(iss, mech_name, '.');
+        std::getline(iss, var_name, '.');
+        if (var_name.empty()) {
+            var_name = "i";
+        }
```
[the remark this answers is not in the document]
pramodk: @jorblancoa : I assume this require changes in neurodamus as well and hence you want to avoid it doing another iteration?

coreneuron/io/reports/report_handler.cpp
pramodk: @jorblancoa : you marked this resolved but with the comment I still found it confusing :)
So, `scalar` means if we need to `scale` because of units? (in that case `scale` and `scalar` are different things)

jorblancoa: I can rename it to scale if you prefer, I was just following neurodamus naming (since it is copying the behavior)

coreneuron/io/reports/report_configuration_parser.cpp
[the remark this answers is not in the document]
jorblancoa: It could, but I would need to add quite a bit of extra functionality in neurodamus (check if summation report and then split report_on and see if there is variable name...). Right now this is done in neurodamus after calling the addSummationReport (we skip in case of coreneuron) in the method evaluateSummationReportOn() so we would need to repeat all the logic while in coreneuron is just 2 lines of code.

jorblancoa:
please retest

jorblancoa:
retest please

jorblancoa:
please retest

pramodk (approved):
LGTM

Issue #197, opened 2019-09-30
Implement fast_imem_calculation for GPU


iomaganaris:
The new fast_imem calculation introduced to `CoreNeuron` should be implemented for GPU as well using OpenACC.
The related functions to the calculations are: `nrn_calc_fast_imem`, `nrn_finitialize`, `nrn_rhs` and `nrn_lhs`.
Care should be taken also for the code generation for GPU from `mod2c` and `NMODL`.
Related PR: https://github.com/BlueBrain/CoreNeuron/pull/196

PR #515, opened 2021-03-29
Add OpenACC data cleanup.

### LINKED ISSUE #513: coreneuron_modtests::datareturn_py test fails on GPU
**Describe the issue**
The `coreneuron_modtests::datareturn_py` CTest test in NEURON fails when a GPU-enabled CoreNEURON build is used. Note that, because of this issue, the test is disabled by default and to see the problem you will have to re-enable it (remove `CONFLICTS gpu` in NEURON's `test/CMakeLists.txt`)

**To Reproduce**
Build NEURON against CoreNEURON with GPU support enabled, run
```
ctest -R coreneuron_modtests::datareturn_py
```
from the build directory.

**Expected behavior**
Success.

**Logs**
The main error is
```
FATAL ERROR: variable in data clause is partially present on the device: name=(unknown)
```
full log: [datareturn_py.log](https://github.com/BlueBrain/CoreNeuron/files/6204601/datareturn_py.log).

**System (please complete the following information)**
 - OS: BB5
 - Compiler: NVHPC 21.2
 - Version: master
 - Backend: GPU

**Additional context**
Identified in the context of https://github.com/neuronsimulator/nrn/pull/1122.

This test calls CoreNEURON multiple times in a single process, and the failure occurs on the second call. The failure is thought to be related to CoreNEURON not correctly freeing all its device-side data.

cc: @pramodk @iomaganaris

olupton:
**Description**
This PR ensures that calls to `acc_copyin(...)` have corresponding calls to `acc_delete(...)`, fixing errors such as
```
FATAL ERROR: variable in data clause is partially present on the device: name=(unknown)
```
when repeatedly running CoreNEURON on GPU in the same process.

Also fix handling of artificial cells to avoid problems with `IntervalFire`.

#513 remains valid, because #197 means that minor additional changes are required in NEURON.

**How to test this?**
Follow instructions in https://github.com/BlueBrain/CoreNeuron/issues/513.

**Test System**
 - OS: BB5
 - Compiler: PGI 21.2
 - Backend: GPU

### DIFF
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -867,15 +867,6 @@ void update_matrix_to_gpu(NrnThread* _nt) {
 #endif
 }
 
-void finalize_data_on_device() {
-    /*@todo: when we have used random123 on gpu and we do this finalize,
-    I am seeing cuCtxDestroy returned CUDA_ERROR_INVALID_CONTEXT error.
-    THis might be due to the fact that the cuda apis (e.g. free is not
-    called yet for Ramdom123 data / streams etc. So handle this better!
-    */
-    return;
-}
-
 /** Cleanup device memory that is being tracked by the OpenACC runtime.
  *
  *  This function painstakingly calls `acc_delete` in reverse order on all
@@ -1021,6 +1012,8 @@ void delete_nrnthreads_on_device(NrnThread* threads, int nthreads) {
 
 void nrn_newtonspace_copyto_device(NewtonSpace* ns) {
 #ifdef _OPENACC
+    // FIXME this check needs to be tweaked if we ever want to run with a mix
+    //       of CPU and GPU threads.
     if (nrn_threads[0].compute_gpu == 0) {
         return;
     }
@@ -1061,6 +1054,8 @@ void nrn_newtonspace_copyto_device(NewtonSpace* ns) {
 
 void nrn_newtonspace_delete_from_device(NewtonSpace* ns) {
 #ifdef _OPENACC
+    // FIXME this check needs to be tweaked if we ever want to run with a mix
+    //       of CPU and GPU threads.
     if (nrn_threads[0].compute_gpu == 0) {
         return;
     }
@@ -1078,6 +1073,8 @@ void nrn_newtonspace_delete_from_device(NewtonSpace* ns) {
 
 void nrn_sparseobj_copyto_device(SparseObj* so) {
 #ifdef _OPENACC
+    // FIXME this check needs to be tweaked if we ever want to run with a mix
+    //       of CPU and GPU threads.
     if (nrn_threads[0].compute_gpu == 0) {
         return;
     }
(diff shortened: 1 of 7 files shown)


olupton:
I re-ran the failing GitLab pipeline and it worked: https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/2536 -- the failure was related to BB5 infrastructure.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -863,6 +876,152 @@ void finalize_data_on_device() {
...
+                acc_delete(info->cellsize, sizeof(int) * info->nwarp);
+                acc_delete(info->stridedispl, sizeof(int) * (info->nwarp + 1));
+                acc_delete(info->lastnode, sizeof(int) * (info->nwarp + 1));
+                acc_delete(info->firstnode, sizeof(int) * (info->nwarp + 1));
+                acc_delete(info->stride, sizeof(int) * info->nstride);
+                acc_delete(info, sizeof(InterleaveInfo));
+            } else {
+                printf("\n ERROR: only --cell_permute = [12] implemented");
+                abort();
+            }
```
pramodk, line 928: nitpick : this check may not be required at the time of cleanup?

(other stuff looks good but will take a closer look later today)

olupton: I removed the check, and also added some missing `acc_delete` counterparts to `acc_copyin` calls.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -984,6 +1157,27 @@ void nrn_sparseobj_copyto_device(SparseObj* so) {
 #endif
 }
 
+void nrn_sparseobj_delete_from_device(SparseObj* so) {
+#ifdef _OPENACC
+    if (nrn_threads[0].compute_gpu == 0) {
+        return;
+    }
```
pramodk, line 1164: Just a note: `SparseObj* so` argument to this function is for every thread object in `nrn_threads` array. It is technically possible to have some NrnThread running on CPUs and some on GPUs (currently we don't have CLI option though!). In this case, checking just 0th thread `nrn_threads[0].compute_gpu` won't be sufficient. We can add a comment here (or in follow-up PR)

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -903,6 +1059,23 @@ void nrn_newtonspace_copyto_device(NewtonSpace* ns) {
 #endif
 }
 
+void nrn_newtonspace_delete_from_device(NewtonSpace* ns) {
+#ifdef _OPENACC
+    if (nrn_threads[0].compute_gpu == 0) {
+        return;
+    }
```
pramodk, line 1066: same : see my comment below

pramodk (commented):
LGTM - some minor comments to address.

coreneuron/gpu/nrn_acc_manager.cpp
olupton: Done in 5456f65.

coreneuron/gpu/nrn_acc_manager.cpp
olupton: Comment added in 5456f65.

PR #517, opened 2021-04-06
CMake / build system improvements + remove outdated warning

olupton:
**Description**
This uses a CMake helper function (`cpp_cc_build_time_copy`) introduced in https://github.com/BlueBrain/hpc-coding-conventions/pull/95 and refined in https://github.com/BlueBrain/hpc-coding-conventions/pull/96 to define build rules that copy files from the source tree to the build tree. This means that more dependencies can be tracked automatically and that CMake does not have to be re-run so often. Also make some implicit dependencies explicit.

Also remove a warning (`Warning: VectorPlay used but NOT implemented on GPU!`) that is outdated.


pramodk (approved):
LGTM

iomaganaris (approved):
LGTM!

PR #523, opened 2021-04-12
In gitlab try with nmodl

alkino:


pramodk:
@alkino : this is closed because?

PR #528, opened 2021-04-17
Psolve direct

### LINKED ISSUE #416: Incremental simulation in direct memory mode
Hello,
In standard NEURON, I use a checkpointing strategy where the overall simulation time is divided into equal segments, psolve is run sequentially for each segment, and the spike times and state traces are saved at the end of the psolve run for that segment. My understanding is that in standard NEURON, h.t is updated to the stop time for the segment, and when psolve is run again, the simulation starts from that time. However, when using CoreNeuron, it appears that the simulation always starts at t=0, even when h.t is set to a positive number. I'd like to ask whether the outlined behavior of psolve in standard NEURON is intended by design, and whether it would be possible to have consistent psolve start time semantics in both standard NEURON and CoreNEURON? Many thanks.

nrnhines:
**ParallelContext.psolve(tstop) behavior for CoreNEURON direct mode is same as NEURON only.**
This means CoreNEURON direct mode does not do its own version of finitialize().
On entry to CoreNEURON, all state needed to continue a simulation is copied from NEURON (including Event Queue).
On exit from CoreNEURON all state needed to continue the simulation in NEURON is copied from CoreNEURON (including EventQueue)

This pull request is associated with neuronsimulator/nrn#1192

Fixes #416

**How to test this?**

```bash
#clone nrn and checkout psolve-direct
cmake .. -DCMAKE_INSTALL_PREFIX=install  -DPYTHON_EXECUTABLE=`which python3` -DNRN_ENABLE_TESTS=ON -DNRN_ENABLE_CORENEURON=ON
make -j 6 install
make test
```

CI_BRANCHES:NEURON_BRANCH=psolve-direct,

Todos: 

- [x] Make sure to update mod2c submodule


alkino:
Please rebuild

alkino:
Please retest

alexsavulescu (commented):
Just some suggestions, I know it's still WIP

alexsavulescu:
Please test this

alexsavulescu:
Please retest this.

alexsavulescu:
@iomaganaris we'd need your help to figure out failed GPU tests

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -145,7 +243,272 @@ void core2nrn_data_return() {
...
+                int watch_begin = first;
+                for (int iml = 0; iml < nodecount; ++iml) {
+                    int iml_permute = permute ? permute[iml] : iml;
+                    std::vector<int>& wiv = watch_info[iml_permute];
+                    for (int ix = first; ix <= last; ++ix) {
+                        int datum = pdata[nrn_i_layout(iml, nodecount, ix, dparam_size, layout)];
+                        if (datum & 2) {  // activated
+                            // if the assert fails then perhaps need to
+                            // revisit the line in nrn/.../nrncore_callbacks.cpp
+                            // wc->flag_ = false; // this is a mystery
```
alexsavulescu, line 289: Leftover comment? 
```suggestion
```

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -145,7 +243,272 @@ void core2nrn_data_return() {
...
+                self_event_weight_map[weight_index].push_back(q);
+            } else {
+                (*core2nrn_SelfEvent_event_noweight_)(
+                    nt.id, td, tar_type, tar_index, flag, is_movable);
+            }
+            break;
         }
+        case PreSynType: {
+            PreSyn* ps = (PreSyn*) d;
+            printf("PreSynType %g\n", td);
```
alexsavulescu, line 444: `ps` is not used.
```suggestion
            // nothing to transfer
```

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -145,7 +243,272 @@ void core2nrn_data_return() {
...
+        }
+    }
+
+    // For self events with weight, find the NetCon index and send that
+    // to NEURON.
+    if (!self_event_weight_map.empty()) {
+        for (int nc_index = 0; nc_index < nt.n_netcon; ++nc_index) {
+            NetCon& nc = nt.netcons[nc_index];
+            int weight_index = nc.u.weight_index_;
+            auto search = self_event_weight_map.find(weight_index);
```
alexsavulescu, line 469:
```suggestion
            const auto& search = self_event_weight_map.find(weight_index);
```

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -145,7 +243,272 @@ void core2nrn_data_return() {
...
+            if (search != self_event_weight_map.end()) {
+                auto& tqitems = search->second;
```
alexsavulescu, line 471:
```suggestion
                const auto& tqitems = search->second;
```

coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -145,7 +243,272 @@ void core2nrn_data_return() {
...
+            break;
+        }
+        case PlayRecordEventType: {
+            // nothing to transfer
+            break;
+        }
+        default: {
+            // In particular, InputPreSyn does not appear in tqueue as it
+            // immediately fans out to NetCon.
+            assert(0);
```
alexsavulescu, line 458:
```suggestion
            std::stringstream qetype;
            qetype << d->type();
            hoc_execerror("core2nrn_tqueue_item -> unimplemented queue event type:", qetype.str().c_str());
```

coreneuron/io/nrn2core_data_init.cpp
```diff
@@ -0,0 +1,440 @@
...
+                               i,
+                               ncte->type[i],
+                               ncte->td[i]);
+#endif
+                    } break;
+
+                    default: {
+                        static char s[20];
+                        sprintf(s, "%d", ncte->type[i]);
+                        hoc_execerror("Unimplemented transfer queue event type:", s);
```
alexsavulescu, line 286:
```suggestion
                        std::stringstream qetype;
                        qetype <<  ncte->type[i];
                        hoc_execerror("Unimplemented transfer queue event type:", qetype);
```

alexsavulescu (approved):
LGTM

olupton:
Submodule commit can be updated now https://github.com/BlueBrain/mod2c/pull/63 is merged.

olupton:
In https://github.com/neuronsimulator/nrn/pull/1192/commits/1471e648ad1643a85d5fd7ce8f7b20cfb20168de I changed `tstop` from `20` to `0.5` for the `watchrange_py` test.
This is expected to make [this](https://github.com/BlueBrain/CoreNeuron/blob/87f607c434784784c9de74c57147bba72ecbc7ca/coreneuron/io/core2nrn_data_return.cpp#L287-L290) assert fail on CPU. [GitLab pipeline 12311](https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/12311) is testing this. 🤞

iomaganaris:
Regarding the `coreneuron_modtests::test_netmove_py` test which uses the `netmove.mod` and is failing on the GPU with `mod2c`, it seems like there is some issue with the 
```
_net_send_buffering(_ml->_net_send_buffer, 2, _tqitem, -1, _ppvar[1*_STRIDE], t + dur , 0.);
```
which gets called instead of `net_move` for both CPU and GPU executions, according to the code printed by mod2c. The `net_move(t + dur)` call of `netmove.mod` is translated to 
```
#if NET_RECEIVE_BUFFERING
    _net_send_buffering(_ml->_net_send_buffer, 2, _tqitem, -1, _ppvar[1*_STRIDE], t + dur , 0.);
#else
 net_move ( _tqitem, _pnt, t + dur ) ;

#endif
```
It seems like there is something wrong with the update of the `NetSendBuffers` in the device. We should check whether they are copied properly on the device and then from the device to the host, so that NEURON has the proper data to continue the simulation.
For the same test that runs on CPU using `NMODL` translator and fails we should compare the generated code for the `net_receive` kernel to see if there is any difference that generates the issue.

olupton:
A behind-the-scenes fix to the GitLab CI infrastructure should fix the failure above of `test:neuron+nmodl:intel` and now https://github.com/BlueBrain/nmodl/pull/714 is merged it should be picked up.

Note that the GitLab CI does not look at the submodule commit, so to pick up a non-default branch of NMODL or NEURON we need to use the
```
CI_BRANCHES:NEURON_BRANCH=psolve-direct,NMODL_BRANCH=...
```
magic in the PR description.

olupton:
Finally in the CI run https://github.com/BlueBrain/CoreNeuron/pull/528#issuecomment-901163587 the `netmove` test is passing on GPU when MOD2C is used. The remaining failures are:
- `netmove` with GPU + NMODL: not yet investigated
- `watchrange` with GPU + (MOD2C or NMODL): we tentatively have a workaround, at least for MOD2C. to be tested soon.

pramodk:
```
99% tests passed, 1 tests failed out of 99
328Total Test time (real) = 337.28 sec
329The following tests FAILED:
330	 15 - coreneuron_modtests::test_netmove_py (Failed)
```
In last CI, above test failed on GPU with MOD2C as well as NMODL!

olupton:
The last pipeline on GitLab https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/14401 was all green in substance, but a glitch stopped the logs being uploaded and the correct status being reported here.

coreneuron/io/core2nrn_data_return.cpp
olupton: :+1:

coreneuron/io/core2nrn_data_return.cpp
olupton: 👎 from me; IMO cleaner to use `auto` (or `auto const`) for things returned by value.

coreneuron/io/core2nrn_data_return.cpp
olupton: :+1:

coreneuron/io/nrn2core_data_init.cpp
olupton: :+1:

olupton (approved):
Looking good. I quickly looked through the whole diff.

As far as I know, the CI pipelines are all "really" passing, but we're having some GitLab issues...

coreneuron/io/core2nrn_data_return.cpp
alexsavulescu: In this specific case, agreed

coreneuron/io/core2nrn_data_return.cpp
alexsavulescu: Good idea, but printf can pollute logs on MPI runs.

coreneuron/apps/main1.cpp
```diff
@@ -120,6 +120,46 @@ char* prepare_args(int& argc, char**& argv, int use_mpi, const char* arg) {
...
+                tr->n_pr = n_pr;
+                tr->n_trajec = n_trajec;
+                tr->vsize = 0;
+                tr->vpr = vpr;
+                tr->gather = new double*[n_trajec];
+                tr->varrays = varrays;
+                tr->scatter = pvars;
+                for (int i = 0; i < n_trajec; ++i) {
+                    tr->gather[i] = stdindex2ptr(types[i], indices[i], nt);
+                }
```
pramodk, line 155: @olupton : just fyi as a reminder : this is where pointer from index is calculated. So here one could have determined `range` of data arrays for `bulk` update.

coreneuron/io/core2nrn_data_return.cpp
olupton: LGTM now.

tests/jenkins/Jenkinsfile
```diff
@@ -51,7 +51,13 @@ pipeline {
...
             steps {
-                sh 'sh tests/jenkins/install_neuron_reportinglib.sh'
+                script{
+                    alt_branches=setAlternateBranches()
+                    sh("""export ${alt_branches};
+                          source $WORKSPACE/tests/jenkins/install_neuron_reportinglib.sh
+                       """
+                    )
+                }
             }
```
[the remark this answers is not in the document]
olupton: This wasn't my change; I guess we'd need to check with @iomaganaris  (https://github.com/BlueBrain/CoreNeuron/commit/783c33bd02c615634267b54f3a882af8a2335a5c https://github.com/BlueBrain/CoreNeuron/commit/23696aaae091697bdbdeffded711c2a83a1d2f0d).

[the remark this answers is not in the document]
pramodk: Ok, it was on psolve-direct in order specify branches. Seems legit, just wanted to verify.

coreneuron/mechanism/patternstim.cpp
```diff
@@ -47,9 +49,6 @@ void nrn_set_extra_thread0_vdata() {
     // if called, must be called before nrn_setup and after mk_mech.
     int type = nrn_get_mechtype("PatternStim");
     if (!corenrn.get_memb_func(type).initialize) {
-        // the NEURON mod file version is not vectorized so the param size
-        // differs by 1 from the coreneuron version.
-        corenrn.get_prop_param_size()[type] += 1;
```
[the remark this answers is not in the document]
pramodk: from 6002996e7f9e0b52f3994c8202eeb16e0db8cf34 by Michael!

alexsavulescu (approved):

PR #537, opened 2021-05-04
Handle correctly destructor block

alkino:
* This feature was not entirely supported and only placeholders were there
* Add the correct signature of destructor's callbacks (mod_f_t)

### DIFF
--- a/coreneuron/mechanism/membfunc.hpp
+++ b/coreneuron/mechanism/membfunc.hpp
@@ -30,7 +30,7 @@ struct Memb_func {
     mod_f_t jacob;
     mod_f_t state;
     mod_f_t initialize;
-    mod_f_t destructor; /* only for point processes */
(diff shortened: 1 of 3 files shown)


ohm314 (approved):
lgtm

ohm314:
Although there is still something failing in the CI...

coreneuron/mechanism/membfunc.hpp
```diff
@@ -30,7 +30,7 @@ struct Memb_func {
     mod_f_t jacob;
     mod_f_t state;
     mod_f_t initialize;
-    Pfri destructor; /* only for point processes */
+    mod_f_t destructor; /* only for point processes */
```
iomaganaris, line 33: Is this still applicable only for point processes?

alkino:
Please retest

PR #541, opened 2021-05-10
Add support for constructor

alkino:
Add a public function: `register_constructor`, because today constructors can only be enabled for point process through
`point_register_mech`.

We still have to find where to call this function, what about checkpoints?


pramodk:
LGTM. Check why CI is failing?

pramodk (approved):
@alkino @iomaganaris : currently, neuron side execute constructor and pass necessary data to coreneuron and hence constructor is not necessary to call. But this might change in the future.

should we go ahead this PR without calling/using constructor for now?

iomaganaris (approved):
Yes, I agree. This LGTM at this state

PR #546, opened 2021-05-10
Enable auto checkpointing on SIGTERM

ferdonline:
**Motivation**
As a follow up to #252, we want CoreNeuron to be able to create checkpoints right before an allocation expires.
Since most job schedulers send a SIGTERM before sigkill, we implement a handler for such signal. It may, however, be needed  to tune the time to sending this signal, since long simulations may take a bit of time to write everything out.

*Implementation*
Checkpoints are created in a folder `_corenrn_ckpt` inside the output root only if a minimum amount of time elapsed.
This directory is checked for existence on startup if no `--restore` is provided.

### DIFF
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -12,8 +12,9 @@
  * @brief File containing main driver routine for CoreNeuron
  */
 
-#include <cstring>
 #include <climits>
+#include <cstring>
+#include <csignal>
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


ferdonline:
I don't really understand why the CI could fail in GPU. @pramodk Ideas?

coreneuron/apps/main1.cpp
```diff
@@ -402,6 +416,20 @@ std::unique_ptr<ReportHandler> create_report_handler(ReportConfiguration& config
...
 
+/**
+ * \brief Installs a SIGTERM handler so that we finish the current simulation without losing data
+ * \return True if a checkpoint was performed. False otherwise (not enough elapsed time)
+ */
+static void install_sigterm_handler() {
+    auto sigh = [](int) {
+        std::cerr << "SIGTERM caught! Halting sim and dumping checkpoint" << std::endl;
+        coreneuron::stoprun = true;
+    };
```
alkino, line 427: What is often do is to enable a "double sigterm", to really cut things

How long is this checkpoint can be?

pramodk:
> I don't really understand why the CI could fail in GPU. @pramodk Ideas?

sorry for delay - this issue is being investigated. You can ignore this error.

olupton:
Can you rebase this @ferdonline?

PR #562, opened 2021-05-19
Move Phase2::create_tml to coreneuron/sim/multicore

kotsaloscv:
Move Phase2::create_tml to coreneuron/sim/multicore


ohm314 (approved):
LGTM

PR #567, opened 2021-05-21
Cell permutation documentation

kotsaloscv:
Cell permutation documentation

### DIFF
--- a/coreneuron/permute/cellorder1.cpp
+++ b/coreneuron/permute/cellorder1.cpp
@@ -250,7 +250,7 @@ size_t level_from_leaf(VecTNode& nodevec) {
 }
 
 /**
- * \brief Set the cellindex to distinguish the difference cells
+ * \brief Set the cellindex to distinguish the different cells.
  */
 static void set_cellindex(int ncell, VecTNode& nodevec) {
     for (int i = 0; i < ncell; ++i) {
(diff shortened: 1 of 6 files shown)


coreneuron/permute/cellorder1.cpp
```diff
@@ -611,7 +626,8 @@ static void admin2(int ncell,
         size_t i = nodebegin[iwarp];
         while (i < j) {
             i += stride_length(i, j, nodevec);
-            ++nc;
+            ++nc;  // ncycles refers to how many times a warp should cycle this level before going
+                   // to a higher one
```
nrnhines, line 630: I no longer remember. Is this the maximum number of compartments for a given core in a warp. I.e a cycle is
a single triangularization or back substitution step. And, because of load imbalance, some cores in the warp need more cycles than other cores. (toward the end, only a few cores of the warp are doing anything).

kotsaloscv: From what I understand, whenever nc per warp/thread goes to zero (after nc cycles), this means that we are done with all tree depths and we are ready to break. 
```
        // if finished with all tree depths then ready to break
        // (note that break is not allowed in OpenACC)
        if (icycle == 0) {
            has_subtrees_to_compute = false;
            continue;
        }
        --icycle;
        istride = stride[icycle];
        i -= istride;
```
I am not really sure how to formalize nc in a different way.

iomaganaris (approved):
LGTM
We also discussed offline about some small additions to the documentation but in general I think that this is in a good state to merge.
Additional documentation after more discussions can be added in the future if needed

nrnhines (approved):
Thanks! This materially improves the readability.

pramodk (approved):
LGTM

PR #568, opened 2021-05-21
Explicitly specify NVIDIA compute capabilities.

olupton:
**Description**

This means we can force consistency between the CUDA and OpenACC code and avoid linker errors.

Previously, compilation was successful on machines with a sufficiently modern GPU, [as the NVHPC compiler modifies its behaviour based on the current system hardware](https://docs.nvidia.com/hpc-sdk/archive/21.2/compilers/hpc-compilers-user-guide/index.html#compute-cap), but it failed on machines with no GPU. This was because the default list of compute capabilities (3.5, 5.0, 6.0 and 7.0 in NVHPC 21.2) included a value (3.5) that was not compatible with the options used to compile the CoreNEURON CUDA library.

Also explicitly specify the CUDA version to avoid a similar issue with a newer CUDA version being assumed for the OpenACC compilation than was used for the CUDA part on a machine with no GPU.

With this change it should be possible to **build** CoreNEURON and `.mod` files that call CUDA routines on a machine without a modern GPU. This should allow https://github.com/BlueBrain/spack/pull/1172 to work.

**How to test this?**
```
spack install -j 36 neuron+coreneuron~debug~legacy-unit+tests%nvhpc ^coreneuron+gpu~legacy-unit~report+tests ^py-cython%gcc ^py-numpy%gcc
```
on a machine without a GPU with/without this change.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.2
 - Version: master
 - Backend: GPU on a non-GPU machine

### DIFF
--- a/CMake/OpenAccHelper.cmake
+++ b/CMake/OpenAccHelper.cmake
@@ -24,9 +24,23 @@ if(CORENRN_ENABLE_GPU)
 
   # various flags for PGI compiler with GPU build
   if(${CMAKE_C_COMPILER_ID} STREQUAL "PGI" OR ${CMAKE_C_COMPILER_ID} STREQUAL "NVHPC")
-
-    # workaround for old PGI version
-    set(PGI_ACC_FLAGS "-acc")
+    # find_cuda produce verbose messages : use new behavior to use _ROOT variables
+    if(POLICY CMP0074)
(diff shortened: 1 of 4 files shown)


pramodk (approved):
LGTM

PR #571, opened 2021-05-26
Simplify GitLab CI config

olupton:
With https://github.com/BlueBrain/spack/pull/1172 the dependencies of the GPU CI are now deployed. This means we can remove some workarounds that avoided building these dependencies in the CI.

Also, running GPU builds on non-GPU nodes is now supported so we can remove some constraints there.


pramodk (approved):
LGTM

PR #572, opened 2021-05-26
Remove vector 

alkino:


iomaganaris:
Yes, those functions are used in `VERBATIM` blocks of mod files. One of their usage is on the delayed connections implementation in `CoreNEURON` for allocating vectors inside the mod files.
Maybe this can be part of a discussion about removing `VERBATIM` blocks from our mod files.

ohm314:
Yeah, for now, let's not remove. They will be replaced by the NMODL VECTOR type started discussing a few months ago. I would close this PR for now.

PR #574, opened 2021-06-01
Implement fast_imem calculation on GPU.

olupton:
**Description**
This PR adds support for copying the `fast_imem` data structures to/from a compute device/GPU and adds OpenACC `#pragma`s to offload computations on those data. This closes https://github.com/BlueBrain/CoreNeuron/issues/197.

This will allow us to improve the coverage of the NEURON tests, where some checks [are disabled](https://github.com/neuronsimulator/nrn/blob/d9462cc81b5d3a264b947a825499cbea1e22eee1/test/coreneuron/test_direct.py#L47-L49) because of this issue. It is also necessary to get the test suite running with CoreNEURON+NMODL+GPU, because NMODL's generated code crashes if the data structures are not copied to the GPU.

Some fixes were also needed to the `.mod` file translation to make the NEURON tests pass, so this changes the submodule commits to include https://github.com/BlueBrain/mod2c/pull/64 and https://github.com/BlueBrain/nmodl/pull/681.

This PR also includes fixes for two memory errors found with Valgrind/memcheck, and a workaround so that not passing `--mpi` to an MPI build does not cause a crash.

**How to test this?**
Build NEURON with CoreNEURON, NMODL and GPU support enabled and run the tests:
```
cmake ..  -DNRN_ENABLE_TESTS=ON -DNRN_ENABLE_CORENEURON=ON -DCORENRN_ENABLE_GPU=ON -DCORENRN_ENABLE_NMODL=ON
cmake --build . --parallel
ctest -j 8
```
without this PR, some tests will fail with
```
CUDA Exception: Warp Illegal Address
```
(with an NVIDIA GPU), with this PR the only failures should be in the `testcorenrn_gf` and `testcorenrn_watch` tests, which will be fixed separately.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.2
 - Backend: GPU

CI_BRANCHES:NEURON_BRANCH=olupton/gpu-fast-imem,

### DIFF
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -291,6 +291,23 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
             acc_memcpy_to_device(&(d_nt->_shadow_d), &d_shadow_ptr, sizeof(double*));
         }
 
+        /* Fast membrane current calculation struct */
+        if (nt->nrn_fast_imem) {
+            auto* d_fast_imem = reinterpret_cast<NrnFastImem*>(
+                acc_copyin(nt->nrn_fast_imem, sizeof(NrnFastImem)));
+            acc_memcpy_to_device(&(d_nt->nrn_fast_imem), &d_fast_imem, sizeof(NrnFastImem*));
+            {
+                auto* d_ptr = reinterpret_cast<double*>(
+                    acc_copyin(nt->nrn_fast_imem->nrn_sav_rhs, nt->end * sizeof(double)));
+                acc_memcpy_to_device(&(d_fast_imem->nrn_sav_rhs), &d_ptr, sizeof(double*));
+            }
+            {
+                auto* d_ptr = reinterpret_cast<double*>(
+                    acc_copyin(nt->nrn_fast_imem->nrn_sav_d, nt->end * sizeof(double)));
+                acc_memcpy_to_device(&(d_fast_imem->nrn_sav_d), &d_ptr, sizeof(double*));
+            }
+        }
+
         if (nt->n_pntproc) {
             /* copy Point_processes array and fix the pointer to execute net_receive blocks on GPU
              */
@@ -659,6 +676,11 @@ void update_nrnthreads_on_host(NrnThread* threads, int nthreads) {
                 acc_update_self(nt->_shadow_d, pcnt * sizeof(double));
             }
 
+            if (nt->nrn_fast_imem) {
+                acc_update_self(nt->nrn_fast_imem->nrn_sav_rhs, nt->end * sizeof(double));
+                acc_update_self(nt->nrn_fast_imem->nrn_sav_d, nt->end * sizeof(double));
+            }
+
             if (nt->n_pntproc) {
                 acc_update_self(nt->pntprocs, nt->n_pntproc * sizeof(Point_process));
             }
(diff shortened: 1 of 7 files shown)


olupton:
Retest this please Jenkins.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -787,6 +814,19 @@ void update_voltage_from_gpu(NrnThread* nt) {
...
+/**
+ * @brief Copy fast_imem vectors from GPU to CPU.
+ *
+ */
+void update_fast_imem_from_gpu(NrnThread* nt) {
+    if (nt->compute_gpu && nt->end > 0 && nt->nrn_fast_imem) {
+        int num_fast_imem = nt->end;
+        double* fast_imem_d = nt->nrn_fast_imem->nrn_sav_d;
+        double* fast_imem_rhs = nt->nrn_fast_imem->nrn_sav_rhs;
+#pragma acc update host(fast_imem_d [0:num_fast_imem], fast_imem_rhs [0:num_fast_imem])
```
pramodk, line 826: Just as a note: in the old versions of PGI/Cray compiler we had issue of pointer dereference in `pragma`.

coreneuron/sim/treeset_core.cpp
```diff
@@ -144,6 +150,7 @@ static void nrn_lhs(NrnThread* _nt) {
            so here we transform so it only has membrane current contribution
         */
         double* p = _nt->nrn_fast_imem->nrn_sav_d;
+#pragma acc parallel loop present(p, vec_d) if (_nt->compute_gpu) async(_nt->stream_id)
```
pramodk, line 153: This is launched via `async` but I believe the synchronisation via `acc wait` is not immediately required (there should be top level `acc wait` already). cc: @iomaganaris

pramodk (approved):
LGTM

coreneuron/gpu/nrn_acc_manager.cpp
olupton: OK, maybe we can remove some of these temporary variables if/when we change OpenACC -> OpenMP.

coreneuron/sim/treeset_core.cpp
olupton: That was my conclusion too; let's wait for @iomaganaris's review though.

olupton:
I just checked that with NMODL+GPU+the current version of this branch then I get
```
98% tests passed, 2 tests failed out of 83
The following tests FAILED:
         51 - testcorenrn_gf::compare_results (Failed)
         73 - testcorenrn_watch::compare_results (Failed)
```
as expected (https://github.com/BlueBrain/nmodl/issues/675, https://github.com/BlueBrain/nmodl/issues/678).

Because the tests are not included in the CI yet, I think it's fine to merge this without waiting for fixes for those issues.

pramodk:
thanks! good to merge!

coreneuron/sim/treeset_core.cpp
iomaganaris: IIUC `stream_id` should always be `0` unless we use `openmp` threading, which in the normal case we don't. 
More details here: https://github.com/BlueBrain/CoreNeuron/blob/f61ec320b2f36fb9579006a741235f35e4b25e6e/coreneuron/io/phase2.cpp#L843
However since there is still this case I think that the `async(stream_id)` is still needed

iomaganaris (approved):
Thanks a lot for taking care of this! The PR LGTM as well :+1:

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -241,7 +241,7 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
                 d_displ = (int*) acc_copyin(nrb->_displ, sizeof(int) * (nrb->_size + 1));
                 acc_memcpy_to_device(&(d_nrb->_displ), &d_displ, sizeof(int*));
 
-                d_nrb_index = (int*) acc_copyin(nrb->_nrb_index, sizeof(int) * (nrb->_size + 1));
+                d_nrb_index = (int*) acc_copyin(nrb->_nrb_index, sizeof(int) * nrb->_size);
```
alkino, line 244: Can you briefly explain what is behind this change?

olupton: The change is to make it match https://github.com/BlueBrain/CoreNeuron/blob/2c51992f4548210180909dff30e3f8b9a292b289/coreneuron/io/phase2.cpp#L461
without the change then `acc_copyin` reads 4 undefined bytes at the end (Valgrind complained).

PR #575, opened 2021-06-03
Build Coreneuron based on nmodl in gitlab ci tests

alkino:


olupton (approved):
LGTM, it looks like GitLab triggered the pipeline twice and reported the wrong status to GitHub. The last commit in this PR did succeed: https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/7815

PR #583, opened 2021-06-23
Enable CoreNEURON CI with NMODL

pramodk:
Supersede #576 

Apply patch mentioned in https://github.com/neuronsimulator/nrn/pull/1261#issuecomment-866811604

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,7 +61,7 @@ build:nmodl:gpu:
     SPACK_PACKAGE: nmodl
     SPACK_PACKAGE_REF: ''
     SPACK_PACKAGE_SPEC: ~legacy-unit~python
-    SPACK_PACKAGE_DEPENDENCIES:"${SPACK_PACKAGE_DEPENDENCIES}^py-pyyaml%gcc"
+    SPACK_PACKAGE_DEPENDENCIES: "${SPACK_PACKAGE_DEPENDENCIES}^py-pyyaml%gcc"
   extends:
     - .spack_build
     - .spack_nvhpc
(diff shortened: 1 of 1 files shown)


pramodk:
NEURON PR https://github.com/neuronsimulator/nrn/pull/1347 is pending but that is not a blocker for this PR.

@alkino : anything else needs to be done here?

.gitlab-ci.yml
```diff
@@ -74,6 +85,21 @@ build:coreneuron:intel:
     - .spack_build
     - .spack_intel
 
+build:coreneuron+nmodl:gpu:
+  variables:
+    SPACK_PACKAGE: coreneuron
+    # +report pulls in a lot of dependencies and the tests fail.
+    # See https://github.com/BlueBrain/CoreNeuron/issues/518 re: build_type
+    SPACK_PACKAGE_SPEC: +nmodl+gpu+tests~legacy-unit~report build_type=RelWithDebInfo
+    SPACK_PACKAGE_DEPENDENCIES: ^bison%gcc^flex%gcc^eigen%gcc
```
alkino, line 94: Should delete it. It's handled by the `before_script` line 98.

olupton, line 94: I think this line can be dropped (its value will be overridden by the `dotenv` artifacts from the `build:nmodl:gpu` job).
```suggestion
```

olupton (approved):
LGTM, one line that should be removed to avoid confusion.

We could probably reduce repetition in the YAML file with a couple of extra templates (`.spack_coreneuron` and `.spack_nmodl` à la `.spack_neuron`, but it's not a big deal).

alkino (approved):

PR #591, opened 2021-07-19
Check that c++14 is fully supported before compiling

### LINKED ISSUE #228: CMake should check for minimum GCC version requirement for C++14 support
CoreNEURON currently requires compiler with C++14 support. We have following in top level CMake:

```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

This make sure to add C++14 flags (e.g. -std=c++1y) but this is not sufficient for C++14 and result into following error:

```
[  6%] Building CXX object external/coreneuron/coreneuron/CMakeFiles/coreneuron.dir/nrniv/cvodestb.cpp.o
In file included from /home/ext_salvadordura_gmail_com/nrn/external/coreneuron/coreneuron/nrniv/cvodestb.cpp:29:0:
/home/ext_salvadordura_gmail_com/nrn/external/coreneuron/coreneuron/coreneuron.hpp:77:7: internal compiler error: in gen_type_die_with_usage, at dwarf2out.c:19525
 class CoreNeuron {
       ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://bugzilla.redhat.com/bugzilla> for instructions.
Preprocessed source stored into /tmp/ccqWAT4G.out file, please attach this to your bugreport.
make[2]: *** [external/coreneuron/coreneuron/CMakeFiles/coreneuron.dir/nrniv/cvodestb.cpp.o] Error 1
make[1]: *** [external/coreneuron/coreneuron/CMakeFiles/coreneuron.dir/all] Error 2
make: *** [all] Error 2
```

See #219 ([comment](https://github.com/BlueBrain/CoreNeuron/issues/219#issuecomment-568796069)).

alkino:
Setting CMAKE_CXX_STANDARD only add the flag, but maybe that the compiler is
too old and does not fully support it.

This way, it will fail before compiling and avoid understandable errors during
compilation.

Fix #228

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,10 @@ set(CMAKE_BUILD_TYPE
     RelWithDebInfo
     CACHE STRING "Empty or one of Debug, Release, RelWithDebInfo")
 
-if (NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-  message(FATAL_ERROR "This compiler does not fully support C++14, choose an higher version or an other compiler.")
+if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+  message(
+    FATAL_ERROR
+      "This compiler does not fully support C++14, choose an higher version or an other compiler.")
 endif()
 
 # =============================================================================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ set(CMAKE_BUILD_TYPE
     RelWithDebInfo
     CACHE STRING "Empty or one of Debug, Release, RelWithDebInfo")
 
+if (NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+  message(FATAL_ERROR "This compiler does not fully support C++14, choose an higher version or an other compiler.")
+endif()
+
 # =============================================================================
 # Settings to enable project as submodule
 # =============================================================================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,8 @@ set(CMAKE_BUILD_TYPE
     RelWithDebInfo
     CACHE STRING "Empty or one of Debug, Release, RelWithDebInfo")
 
-if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES OR NOT "cxx_digit_separators" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES OR NOT "cxx_digit_separators" IN_LIST
+                                                          CMAKE_CXX_COMPILE_FEATURES)
   message(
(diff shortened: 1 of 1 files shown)


olupton:
I just tried this on BB5:
```
[olupton@bbpv2 build_olli]$ module load unstable cmake hpe-mpi
[olupton@bbpv2 build_olli]$ cmake ..
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
```
and the CMake step didn't give an error. The build failed though:
```
[ 26%] Building CXX object coreneuron/CMakeFiles/coreneuron.dir/apps/corenrn_parameters.cpp.o
In file included from /gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/coreneuron/apps/corenrn_parameters.cpp:9:0:
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/coreneuron/apps/corenrn_parameters.hpp:43:28: warning: missing terminating ' character [enabled by default]
     unsigned spikebuf = 100'000;           /// Internal buffer used on every rank for spikes
                            ^
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/coreneuron/apps/corenrn_parameters.hpp:43:5: error: missing terminating ' character
     unsigned spikebuf = 100'000;           /// Internal buffer used on every rank for spikes
     ^
```
I hadn't actually seen this error before, but it looks like missing support for [digit separators](https://en.cppreference.com/w/cpp/language/integer_literal#Single_quote).

I thought with this PR there should have been an error at CMake execution time?

alexsavulescu:
C++14 support is broken/incomplete for `GCC4.x`: https://gcc.gnu.org/projects/cxx-status.html
We should add a check for it (>5).

alkino:
So it is useless to check it in the way I did here, right?
This cxx_14_support only tell us that the flag is available, right?

should we check version for gcc / clang / icc / vscode?

alexsavulescu:
Thinking about it, it's not that straightforward. 

I remember reading about this: https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html
but not sure that will work out.

Maybe the only valid way would be to have a `test.cpp` with all C++14 features and try to compile that with whatever compiler comes in.  But I wouldn't go that far, especially if we can rely on `CXX_STANDARD_REQUIRED` for most of the time.

olupton:
In the case of GCC 4 then it's enough to change it to
```cmake
if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES OR NOT "cxx_digit_separators" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
```
but for later C++ standards they don't have all of these specific feature macros.

I would vote for doing the best we can with `CMAKE_CXX_COMPILE_FEATURES` and not writing our own table of compilers and minimum versions.

olupton:
Failures should be fixed by #592, but I'm not sure we need to re-run the CI on this branch (it only touches CMake files...)

alkino:
I rerun only failed jobs

olupton (approved):
LGTM!

PR #595, opened 2021-07-20
Use CUDA unified memory for Random123 state

### LINKED ISSUE #345: Random123 in VERBATIM : Allow GPU built executable to run on CPU
Currently, if Random123 is used in the mod file then GPU built binary can't be run on CPU. This limitation should be removed as it's often required to run code on CPU for debugging and performance comparison.

olupton:
**Description**
This modifies how we allocate state variables for Random123 in GPU-enabled builds. This should simplify things, and enable use-cases like in #528 where we need to manipulate this state on the device from CPU code in ways that the existing API did not allow.

Because the state is now accessible from both CPU and GPU code, this also closes #345.

**How to test this?**
Build and run with/without GPU support enabled.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.2
 - Version: master
 - Backend: CPU/GPU

CI_BRANCHES:NEURON_BRANCH=olupton/random123-unified-memory,


coreneuron/utils/randoms/nrnran123.cu
```diff
@@ -5,99 +5,107 @@
...
-__device__ static philox4x32_key_t k = {{0}};
-__device__ static unsigned int instance_count_ = 0;
-__device__ size_t nrnran123_instance_count() {
-    return instance_count_;
-}
 
-__device__ size_t nrnran123_state_size() {
-    return sizeof(nrnran123_State);
-}
+#include <atomic>
```
pramodk, line 18: As mentioned during discussion, we should consider interoperability with OpenMP threading. For now, we can use MUTLOCK.

pramodk (commented):
LGTM!

coreneuron/utils/randoms/nrnran123.cu
olupton: I dropped `<atomic>` and now use the `OMP_Mutex` wrapper we already have.

pramodk (approved):
Once channel-benchmark is validated to run with a single exe on CPU as well as GPU, this is good to go!

pramodk:
Retest this please

olupton:
It was tested locally that `channel-benchmark` produces consistent results with/without `--gpu` in a GPU-enabled build.

Note that [a bug in the NVHPC compiler version 21.3 and 21.5](https://forums.developer.nvidia.com/t/incorrect-cpu-results-with-pragma-acc-atomic-capture/184453) means version 21.2 (or older?) or 21.7 (or hopefully newer!) must be used.

iomaganaris (approved):
LGTM :+1:

pramodk:
Replayed rxd-mpi tests and they passed - https://bbpgitlab.epfl.ch/hpc/coreneuron/-/jobs/38730 (in previous CI there was timeout)

PR #596, opened 2021-07-20
Fix compilation error with derivimplicit solver and use of celsius as extern variable

pramodk:
As described in #498, all extern variables require acc declare
create/copyin caluse. As variables can be updated after device
attachment (like celsius), #498 has taken care to update them.

fixes BlueBrain/nmodl/issues/707

Todos:

- [x] Merge NMODL PR first BlueBrain/nmodl/pull/709
- [x] Update nmodl submodule in this PR and then merge 

Fixes # (issue)

**How to test this?**

Run testderiv.hoc test from testcorenrn repository

**Test System**
 - OS: BB5
 - Compiler: default nvhpc + CUDA
 - Backend: GPU

CI_BRANCHES:NEURON_BRANCH=pramodk/derivimplicit-legacy-fix,

### DIFF
--- a/coreneuron/nrnconf.h
+++ b/coreneuron/nrnconf.h
@@ -36,10 +36,13 @@ extern double celsius;
 #pragma acc declare create(celsius)
 
 extern double pi;
+#pragma acc declare create(pi)
+
+extern int secondorder;
+#pragma acc declare create(secondorder)
 
 extern double t, dt;
 extern int rev_dt;
-extern int secondorder;
 extern bool stoprun;
 extern const char* bbcore_write_version;
 #define tstopbit   (1 << 15)
--- a/coreneuron/nrnconf.h
+++ b/coreneuron/nrnconf.h
@@ -31,7 +31,10 @@ using Symbol = char;
 #define VEC_AREA(i) (_nt->_actual_area[(i)])
 #define VECTORIZE   1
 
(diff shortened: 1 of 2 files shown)


ferdonline:
Worst bugs are those whose resolution are 1LOC

coreneuron/nrnconf.h
```diff
@@ -31,7 +31,10 @@ using Symbol = char;
 #define VEC_AREA(i) (_nt->_actual_area[(i)])
 #define VECTORIZE   1
 
+// extern variables require acc declare
 extern double celsius;
+#pragma acc declare create(celsius)
+
 extern double pi;
```
iomaganaris, line 38: Should we do the same for `pi`?

iomaganaris: At the moment `pi` is only used in `lfp.cpp`, so there is no need for it to be allocated here on the GPU, however if it's used at some point in mod files we should add more logic to update in in GPU in mod2c/NMODL

pramodk: Yes. I did the change for `pi` as well and then revered it because `pi` is typically not used (?). But I can change it.

olupton (approved):
LGTM

iomaganaris (approved):

Issue #601, opened 2021-07-24
CMake Error at external/coreneuron/coreneuron/CMakeLists.txt:37:: file COPY cannot find external/Random123/include/Random123


pramodk:
**Describe the issue**

While building NEURON on Juelich's JUSUF system I get following error:

```console

[kumbhar1@jsfl01 build]$ cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install -DNRN_ENABLE_INTERVIEWS=OFF -DNRN_ENABLE_RX3D=ON -DNRN_ENABLE_CORENEURON=ON
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /p/software/jusuf/stages/2020/software/GCCcore/9.3.0/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /p/software/jusuf/stages/2020/software/GCCcore/9.3.0/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
heads/master
-- Setting build type to 'RelWithDebInfo' as none was specified.
-- Found BISON: /usr/bin/bison (found version "3.0.4")
-- Found FLEX: /usr/bin/flex (found version "2.5.37")
-- Found Readline: /p/software/jusuf/stages/2020/software/libreadline/8.0-GCCcore-9.3.0/include
-- Found PythonInterp: /p/software/jusuf/stages/2020/software/Python/3.8.5-GCCcore-9.3.0/bin/python3 (found version "3.8.5")
-- Found Cython: /p/software/jusuf/stages/2020/software/Python/3.8.5-GCCcore-9.3.0/bin/cython (found version "0.29.21")
-- Found MPI_C: /opt/ddn/ime/lib/libim_client.so (found version "3.1")
-- Found MPI_CXX: /opt/ddn/ime/lib/libim_client.so (found version "3.1")
-- Found MPI: TRUE (found version "3.1")
-- Found PythonInterp: /p/software/jusuf/stages/2020/software/Python/3.8.5-GCCcore-9.3.0/bin/python3 (found suitable version "3.8.5", minimum required is "3.6")
[...]

PR #602, opened 2021-07-26
Be more restrictive when looking for Random123

alkino:
As the file search to find Random123 (LICENSE) is really a common file
we should be restrictive, otherwise we can found other projects.

Try to fix: #601


pramodk (approved):
LGTM! 

With this patch, on the JUSUF system:

```
-- Could NOT find Random123 (missing: Random123_PROJ)
-- Sub-module Random123 missing: running git submodule update --init --recursive
Cloning into 'external/Random123'...
Submodule path 'external/Random123': checked out 'b5cb15614e8f3eb0ff316bdc50c300ddd32fa3c1'
-- Found CLI11: /p/home/jusers/kumbhar1/jusuf/nrn/external/coreneuron/external/CLI11
```

PR #604, opened 2021-07-27
Add a way to enable NRN_PRCELLSTATE inside nrnivmodl-core

alkino:


iomaganaris:
I would also suggest adding a CMake flag for this, so when we build CoreNEURON we don't have to run nrnivmodl to use prcellstate with the mechanisms already inbuilt in CoreNEURON (hh for example)

CMakeLists.txt
```diff
@@ -94,6 +94,7 @@ option(CORENRN_ENABLE_UNIT_TESTS "Enable unit tests execution" ON)
 option(CORENRN_ENABLE_GPU "Enable GPU support using OpenACC" OFF)
 option(CORENRN_ENABLE_SHARED "Enable shared library build" ON)
 option(CORENRN_ENABLE_LEGACY_UNITS "Enable legacy FARADAY, R, etc" OFF)
+option(CORENRN_ENABLE_NRN_PRCELLSTATE "Enable NRN_PRCELLSTATE debug feature" OFF)
```
iomaganaris, line 97: The option could be `CORENRN_ENABLE_PRCELLSTATE` to avoid any confusion with the `NRN` word?

iomaganaris (approved):
LGTM :+1:
If CI passes this can be merged

alexsavulescu (approved):

Issue #599, opened 2021-07-23
GPU-enabled builds should not call GPU APIs without --gpu


olupton:
#595 closes #345, because GPU-enabled builds using Random123 can now be executed on CPU.

When a GPU-enabled build is executed **without** `--gpu` but on a machine that **does** have a GPU, various data transfers and synchronisation calls are made, which can be seen e.g. with `nvprof`.

We should fix these and make sure all GPU/OpenACC operations are conditional on `nt->compute_gpu`.

olupton:
A more problematic aspect of this is that `__managed__` is not safe to use on nodes without GPUs, as demonstrated by
```
#include <iostream>
__managed__ int x = 42;
int main() {
  std::cout << "x = " << x << std::endl;
  return 0;
}
```
which will segfault on a non-GPU node and print 42 on a node with an NVIDIA GPU.
The real-world example of this is
```
(gdb) bt
#0  0x0000000000570a65 in coreneuron::nrnran123_set_globalindex(unsigned int) ()
#1  0x0000000000496a8d in coreneuron::set_globals (path=<optimized out>, cli_global_seed=0 '\000', cli_global_seed_value=-1) at ../external/coreneuron/coreneuron/io/global_vars.cpp:134
#2  0x0000000000487c2a in coreneuron::nrn_init_and_load_data (argc=<optimized out>, argv=<optimized out>, checkPoints=0x7fffffffd7c0, is_mapping_needed=0 '\000', run_setup_cleanup=1 '\001')
    at ../external/coreneuron/coreneuron/apps/main1.cpp:210
#3  0x000000000046e68c in run_solve_core (argc=<optimized out>, argv=<optimized out>) at ../external/coreneuron/coreneuron/apps/main1.cpp:496
#4  0x000000000040eac0 in solve_core (argc=3, argv=0x7fffffffda58) at /gpfs/bbp.cscs.ch/home/olupton/nrn/build_ollimod2c/share/coreneuron/enginemech.cpp:49
#5  0x000000000040ea99 in main (argc=10867584, argv=0x57f240 <cudart::createGlobalState()>) at /gpfs/bbp.cscs.ch/home/olupton/nrn/build_ollimod2c/share/coreneuron/coreneuron.cpp:14
```
so a follow-up to https://github.com/BlueBrain/CoreNeuron/pull/595 will be needed to enable non-GPU execution in GPU builds _on nodes without GPUs._

olupton:
[...]

PR #606, opened 2021-07-29
Do not use __managed__ in GPU builds. Add utils.

olupton:
**Description**
This fixes CPU execution of GPU builds on machines that do not have GPUs, which were previously segfaulting due to the use of the `__managed__` keyword. Now the handling of Random123 state is more explicit for host/device. This was only added in https://github.com/BlueBrain/CoreNeuron/pull/595, but the issue wasn't noticed at the time because the GPU-enabled CI only executes tests on machines with GPUs.

Also add a pair of helper functions `coreneuron::[de]allocate_unified()` that wrap `cudaMallocManaged()` in GPU builds if `--gpu` was passed at runtime and fall back to `new`/`delete` otherwise, and a method `coreneuron::unified_memory_enabled()` that queries whether this condition is met.

Additionally add a C++ allocator template `coreneuron::unified_allocator<T>` that wraps these functions, and a templated `coreneuron::alloc_deleter<T>` for use with `std::unique_ptr<T, D>`. Also add `coreneuron::allocate_unique` helper from SO.

Cleanup Random123 code by dropping an unused `nrnran123_mutconstruct()` method.

Tweak compilation scripts to allow for circular dependencies between `libcoreneuron.a` and `libcudacoreneuron.a`.
In future these should just be merged into a single library.

This addresses https://github.com/BlueBrain/CoreNeuron/issues/599#issuecomment-888416482. #599 should stay open because various OpenACC calls are still not conditional on `--gpu`.

**How to test this?**
Try running a GPU-built `special-core` without `--gpu` on a machine that does not have an NVIDIA GPU.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.7 / CUDA 11.0 / GCC 9.3
 - Version: master
 - Backend: GPU/CPU

### DIFF
--- a/CMake/OpenAccHelper.cmake
+++ b/CMake/OpenAccHelper.cmake
@@ -9,7 +9,7 @@
 # =============================================================================
 if(CORENRN_ENABLE_GPU)
   # Enable cudaProfiler{Start,Stop}() behind the Instrumentor::phase... APIs
-  add_compile_definitions(CORENEURON_CUDA_PROFILING)
+  add_compile_definitions(CORENEURON_CUDA_PROFILING CORENEURON_ENABLE_GPU)
   # cuda unified memory support
   if(CORENRN_ENABLE_CUDA_UNIFIED_MEMORY)
     add_compile_definitions(CORENEURON_UNIFIED_MEMORY)
@@ -49,8 +49,8 @@ if(CORENRN_ENABLE_GPU)
   endif()
   # -acc enables OpenACC support, -cuda links CUDA libraries and (very importantly!) seems to be
   # required to make the NVHPC compiler do the device code linking. Otherwise the explicit CUDA
-  # device code (.cu files in libcudacoreneuron) has to be linked in a separate, earlier, step,
-  # which apparently causes problems with interoperability with OpenACC. Passing -cuda to nvc++ when
+  # device code (.cu files in libcoreneuron) has to be linked in a separate, earlier, step, which
+  # apparently causes problems with interoperability with OpenACC. Passing -cuda to nvc++ when
   # compiling (as opposed to linking) seems to enable CUDA C++ support, which has other consequences
   # due to e.g. __CUDACC__ being defined. See https://github.com/BlueBrain/CoreNeuron/issues/607 for
   # more information about this. -gpu=cudaX.Y ensures that OpenACC code is compiled with the same
@@ -80,7 +80,7 @@ if(CORENRN_ENABLE_GPU)
     GLOBAL
     PROPERTY
       CORENEURON_LIB_LINK_FLAGS
-      "${NVHPC_ACC_COMP_FLAGS} ${NVHPC_ACC_LINK_FLAGS} -rdynamic -lrt -Wl,--whole-archive -L${CMAKE_HOST_SYSTEM_PROCESSOR} -lcorenrnmech -L${CMAKE_INSTALL_PREFIX}/lib -lcoreneuron -lcudacoreneuron -Wl,--no-whole-archive"
+      "${NVHPC_ACC_COMP_FLAGS} ${NVHPC_ACC_LINK_FLAGS} -rdynamic -lrt -Wl,--whole-archive -L${CMAKE_HOST_SYSTEM_PROCESSOR} -lcorenrnmech -L${CMAKE_INSTALL_PREFIX}/lib -lcoreneuron -Wl,--no-whole-archive"
   )
 else()
   set_property(GLOBAL PROPERTY CORENEURON_LIB_LINK_FLAGS
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -108,6 +108,7 @@ if(CORENRN_ENABLE_GPU)
   # ~~~
   # artificial cells and some other cpp files (using Random123) should be compiled
   # without OpenACC to avoid use of GPU Random123 streams
+  # OL210813: this shouldn't be needed anymore, but it may have a small performance benefit
   # ~~~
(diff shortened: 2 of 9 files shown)


CMake/OpenAccHelper.cmake
```diff
@@ -68,7 +69,7 @@ if(CORENRN_ENABLE_GPU)
     GLOBAL
     PROPERTY
       CORENEURON_LIB_LINK_FLAGS
-      "${PGI_ACC_FLAGS} -rdynamic -lrt -Wl,--whole-archive -L${CMAKE_HOST_SYSTEM_PROCESSOR} -lcorenrnmech -L${CMAKE_INSTALL_PREFIX}/lib -lcoreneuron -lcudacoreneuron -Wl,--no-whole-archive ${CUDA_cudart_static_LIBRARY}"
+      "${PGI_ACC_FLAGS} -rdynamic -lrt -Wl,--whole-archive -L${CMAKE_HOST_SYSTEM_PROCESSOR} -lcorenrnmech -L${CMAKE_INSTALL_PREFIX}/lib -lcoreneuron -lcudacoreneuron -lcoreneuron -Wl,--no-whole-archive ${CUDA_cudart_static_LIBRARY}"
```
ferdonline, line 72: I see `-lcoreneuron` twice. Expected?

olupton: Yes, although I haven't actually tested the `nrnivmodl` workflow (only `nrnivmodl-core`) where IIUC this is used -- the CI will do that for us...

It's a crude workaround (commented elsewhere) for circular dependencies between `libcoreneuron.a` and `libcudacoreneuron.a` introduced by this PR. I didn't immediately find a better **simple** way.

IMO the better solution would be for us to produce a single library instead of these two separate ones. I don't think they are really logically separate libraries, the separation is just a consequence of how things are handled in CMake.

Alternatively we could add a third library...
```
libcoreneuron.a (C++) <depends on> libcudacoreneuron.a (CUDA) <depends on> libcorecoreneuron.a (C++)
```
but I fear this will just get messier without much real benefit.

olupton:
This is blocked by #607.

CMake/OpenAccHelper.cmake
olupton: #609 made it easy to merge `libcoreneuron` and `libcudacoreneron`, so I did that instead. Much neater.

ferdonline (commented):
LGTM but I'm not the best person to review this code. Maybe let's wait for a review from @iomaganaris

kotsaloscv (approved):
LGTM

coreneuron/utils/randoms/nrnran123.cu
```diff
@@ -157,6 +220,8 @@ nrnran123_State* nrnran123_newstream3(uint32_t id1,
     s->c.v[3] = id2;
     nrnran123_setseq(s, 0, 0);
     {
+        // TODO: can I assert something useful about the instance count going
+        // back to zero anywhere? Or that it is zero when some operations happen?
```
pramodk, line 224: Answered in previous comment.

@nrnhines : How the ran123 streams allocated with `nrnran123_newstream3` inside `bbcore_read()` should be allocated? any thoughts?

coreneuron/utils/randoms/nrnran123.cu
```diff
@@ -127,27 +170,47 @@ CORENRN_HOST_DEVICE double nrnran123_uint2dbl(uint32_t u) {
...
+        std::lock_guard<OMP_Mutex> _{g_instance_count_mutex};
+        if (g_instance_count != 0) {
+            std::ostringstream oss;
+            oss << "nrnran123_set_globalindex(" << gix
+                << ") called when a non-zero number of Random123 streams (" << g_instance_count
+                << ") were active. This is not safe, some streams will remember the old value ("
+                << get_global_state().v[0] << ')';
+            throw std::runtime_error(oss.str());
+        }
+    }
```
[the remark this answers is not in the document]
olupton: There wasn't any discussion, but there was a pre-existing comment:
https://github.com/BlueBrain/CoreNeuron/blob/170a0bb74b033de020946c96f67f8699e67e10d4/coreneuron/utils/randoms/nrnran123.cu#L48
which made me think that  we could easily add a diagnostic for this. I haven't tried to run the `psolve-direct` tests with this branch. You're probably right that this should be introduced as a warning instead of a hard error.

coreneuron/CMakeLists.txt
```diff
@@ -118,24 +119,19 @@ if(CORENRN_ENABLE_GPU)
       ${CMAKE_CURRENT_SOURCE_DIR}/io/setup_fornetcon.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/io/corenrn_data_return.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/io/global_vars.cpp)
-
+  # OL210831: why?
   set_source_files_properties(${DIMPLIC_CODE_FILE} ${NMODL_INBUILT_MOD_OUTPUTS}
                               PROPERTIES COMPILE_FLAGS "")
```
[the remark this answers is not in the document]
olupton: Removed.

coreneuron/utils/randoms/nrnran123.cu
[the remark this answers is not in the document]
olupton: I demoted this to a warning.

coreneuron/utils/randoms/nrnran123.cu
pramodk: lets keep this as a separate issue. https://github.com/BlueBrain/nmodl/issues/383 would help to implement this easily.

pramodk (approved):
LGTM

Issue #607, opened 2021-07-30
Random123 global state is not propagated correctly


olupton:
**Describe the issue**
After https://github.com/BlueBrain/CoreNeuron/pull/595 (and possibly before?) there is an issue with the device-side global state of Random123 configured by the `--seed` command line option (and otherwise).

When Random123 methods are called from OpenACC regions it **seems** that the default value ([zero](https://github.com/BlueBrain/CoreNeuron/blob/92b443c7cb2c57c10f173f00ed0e6c230262c516/coreneuron/utils/randoms/nrnran123.cu#L27)) of the global seed is seen by device code, even if it is nominally set to something else.

This is more obvious in the branch of https://github.com/BlueBrain/CoreNeuron/pull/606, where the device-side global is a pointer and its default value (`nullptr`) triggers an assert. 

**To Reproduce**
Check out `master` of CoreNEURON (I used 92b443c7cb2c57c10f173f00ed0e6c230262c516), apply a patch like https://github.com/BlueBrain/CoreNeuron/commit/7ee09ed59cc71d2cd5cf0d00704d5d3728d47cf9:
```diff
diff --git a/coreneuron/apps/main1.cpp b/coreneuron/apps/main1.cpp
index 12b3cb1..bfbc304 100644
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -169,6 +169,13 @@ void nrn_init_and_load_data(int argc,
     // read the global variable names and set their values from globals.dat
     set_globals(corenrn_param.datpath.c_str(), (corenrn_param.seed >= 0), corenrn_param.seed);

+    printf("DEBUG: seed=%u [host]\n", nrnran123_get_globalindex());
+    #pragma acc serial
+    {
+      printf("DEBUG: seed=%u [device]#1\n", nrnran123_get_globalindex());
+    }
+    std::exit(0);
+
     // set global variables for start time, timestep and temperature
     t = checkPoints.restore_time();

diff --git a/coreneuron/utils/randoms/nrnran123.cu b/coreneuron/utils/randoms/nrnran123.cu
index 8e56a19..d21772b 100644
--- a/coreneuron/utils/randoms/nrnran123.cu
+++ b/coreneuron/utils/randoms/nrnran123.cu
@@ -45,8 +45,23 @@ static MUTDEC void nrnran123_mutconstruct() {
 void nrnran123_mutconstruct() {}
 #endif

[...]

Issue #520, opened 2021-04-09
CoreNEURON uses deprecated FindCUDA


olupton:
**Is your feature request related to a problem? Please describe.**
[FindCUDA](https://cmake.org/cmake/help/v3.10/module/FindCUDA.html) has been deprecated since CMake v3.10, and [CUDA has had first-class language support since v3.8](https://cmake.org/cmake/help/v3.8/release/3.8.html#cuda).

**Describe the solution you'd like**
We should avoid deprecated features and follow upstream recommendations, i.e. stop using FindCUDA, `cuda_add_library` etc.

**Additional context**
I suspect that the legacy code does not correctly declare all build dependencies, and that it is responsibly for `cmake --build` re-running the CMake configuration more often than necessary.

olupton:
See also: https://cliutils.gitlab.io/modern-cmake/chapters/packages/CUDA.html

alkino:
We have to support both for a long time, minimum version of cmake for CoreNeuron is 3.7.

pramodk:
@alkino : With 8.0 release done, we can bump the minimum requirement to from 3.7 to something more "reasonable".

olupton:
In #609 I moved to using CMake's CUDA language support, initially using CMake v3.20. I then added workarounds for older versions back to v3.15.

An attempted GPU build with CMake v3.10 / CUDA 11.0 / NVHPC 21.7 did not go so well. CMake labels the NVHPC C/C++ compilers as PGI 21.7, but it does not know how to detect the language features that they support. And it does not seem to be possible to distinguish programmatically between the two cases:
 - C++ feature `X` is known to be unsupported by PGI/NVHPC version `Y`
 - No test is implemented for C++ feature `X`  for PGI/NVHPC compilers

The result is that the check introduced in https://github.com/BlueBrain/CoreNeuron/pull/591 fails. If this test is disabled for PGI/NVHPC, we hit a similar issue with the `catch2` submodule of NEURON. If we add workarounds for both of these (I just skipped the check for CoreNEURON and manually kludged `CMAKE_CXX_COMPILE_FEATURES` for `catch2`, but it's not important) then there is a further issue:
```
[19/199] Linking CXX executable bin/lfp_test_bin
FAILED: bin/lfp_test_bin
[...]

Issue #153, opened 2019-03-15
Prefix project macros to prevent conflicts


tristan0x:
Preprocessing macros defined by the project should be prefixed by `CORENEURON_` to prevent possible conflicts.

tapaswenipathak:
Hi @tristan0x, is this open and can be worked upon?

tristan0x:
Hello @tapaswenipathak 
Thank you for your interest in CoreNeuron!

According to the _add_definitions_ references in CMakeLists.txt files:
```
./CMakeLists.txt:    add_definitions("-DISPC_INTEROP=1")
./CMakeLists.txt:    add_definitions("-DNRNMPI=1")
./CMakeLists.txt:    add_definitions("-DNRNMPI=0")
./CMakeLists.txt:    add_definitions("-DNRN_MULTISEND=0")
./CMakeLists.txt:    add_definitions("-DLAYOUT=0")
./CMakeLists.txt:    add_definitions("-DLAYOUT=1")
./CMakeLists.txt:    add_definitions("-DENABLE_OMP_RUNTIME_SCHEDULE")
./CMakeLists.txt:    add_definitions("-DDISABLE_HOC_EXP")
./CMakeLists.txt:    add_definitions("-DENABLE_SPLAYTREE_QUEUING")
./CMakeLists.txt:    add_definitions("-DNET_RECEIVE_BUFFERING=0")
./CMakeLists.txt:        add_definitions("-DENABLE_REPORTING")
./CMakeLists.txt:    add_definitions("-DCORENEURON_CALIPER")
./CMakeLists.txt:  add_definitions("-DHAVE_MALLOC_H")
./CMakeLists.txt:add_definitions(-DCORENEURON_BUILD)
./CMakeLists.txt:    add_definitions(-DSWAP_ENDIAN_DISABLE_ASM)
./CMakeLists.txt:        add_definitions( -DPG_ACC_BUGS)
./CMakeLists.txt:                add_definitions( -DCUDA_PROFILING)
./CMakeLists.txt:        add_definitions(-DCUDA_MODULES_DISABLED)
./CMakeLists.txt:    add_definitions("-DDISABLE_TIMEOUT")
./CMakeLists.txt:    add_definitions("-DEXPORT_MECHS_FUNCTIONS")
./tests/CMakeLists.txt:    add_definitions(-DBOOST_TEST_DYN_LINK=TRUE)
[...]

PR #609, opened 2021-08-12
Modernise CMake/CUDA and fix link issues

olupton:
**Description**
This changeset changes the CMake configuration of GPU builds to:
- [x] Use CMake's language support for CUDA, instead of the deprecated `find_package(CUDA ...)`, `cuda_add_library(...)`, etc.
- [x] Increase the minimum CMake version for GPU builds to v3.15, with an explicit error message if this is not respected. 
- [x] Retire the custom `CORENRN_GPU_CUDA_COMPUTE_CAPABILITY` option; we now use the standard `CMAKE_CUDA_ARCHITECTURES` variable
- [x] Avoid multiple device code linking steps, which apparently caused problems (https://github.com/BlueBrain/CoreNeuron/issues/607) with global state.
- [x] Prefix more preprocessor macro names (https://github.com/BlueBrain/CoreNeuron/issues/153)
- [x] Bump the https://github.com/BlueBrain/hpc-coding-conventions submodule commit to support `clang-format-12`.
- [x] Fix a silly error in https://github.com/BlueBrain/CoreNeuron/pull/595 affecting builds with asserts disabled.

These changes force compilation to proceed slightly differently to a standard mixed CUDA/C++ project. Instead of allowing CMake to generate an explicit device linker step, i.e.
```
nvcc -dc -o foo.cu.o foo.cu ...
nvcc -dlink -o dlink.o *.cu.o ...
ar qc libcudastuff.a *.cu.o dlink.o
```
before compiling OpenACC/GPU-enabled C++ code (which emits additional device code):
```
nvc++ -acc -gpu=cuda11.0,cc70 -o main main.cpp -lcudastuff ...
```
we instead let `nvc++` do the device code linking itself, i.e. something more like
```
nvc++ -acc -gpu=cuda11.0,cc70 -o main.cpp.o -c main.cpp
nvc++ -acc -gpu=cuda11.0,cc70 -cuda -o main main.cpp.o *.cu.o # -cuda but no dlink.o!
```
which seems to give correct results. This is a bit tortuous in CMake presumably because the same pattern would fail with a GPU-unaware C++ compiler, e.g. GCC or Clang. The hypothesis is that the old way of doing things fell foul of
> It is possible to do multiple device links within a single host executable, as long as each device link is independent of the other. This requirement of independence means that they cannot share code across device executables, nor can they share addresses (e.g., a device function address can be passed from host to device for a callback only if the device link sees both the caller and potential callback callee; you cannot pass an address from one device executable to another, as those are separate address spaces). 

from [the CUDA documentation](https://docs.nvidia.com/cuda/archive/11.0/cuda-compiler-driver-nvcc/index.html#examples), while the new way ensures there is a single device link step including both the code generated from `.cu` files and that from OpenACC regions.

We now also prefer to dynamically link the CUDA runtime, `libcudart.so`. `nvc++ -cuda` seems to prefer this and only allows it to be steered by the `-static-nvidia` option, which would also statically link the OpenACC runtime (which has always been dynamically linked). Setting `CMAKE_CUDA_RUNTIME_LIBRARY=Shared` stops CMake from emitting `-lcudart_static`, which causes segfaults at teardown in combination with `-cuda`'s dynamic linking. 

cc: @kotsaloscv 

Closes https://github.com/BlueBrain/CoreNeuron/issues/520. Closes https://github.com/BlueBrain/CoreNeuron/issues/607.

**How to test this?**
Follow instructions in https://github.com/BlueBrain/CoreNeuron/issues/607 to test the link/global state issue.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.7 / CUDA 11.0
 - Version: master
 - Backend: GPU

### DIFF
--- a/CMake/OpenAccHelper.cmake
+++ b/CMake/OpenAccHelper.cmake
@@ -8,51 +8,40 @@
 # Prepare compiler flags for GPU target
 # =============================================================================
 if(CORENRN_ENABLE_GPU)
-
   # cuda unified memory support
   if(CORENRN_ENABLE_CUDA_UNIFIED_MEMORY)
     set(UNIFIED_MEMORY_DEF -DUNIFIED_MEMORY)
   endif()
-
-  # if user don't specify host compiler, use gcc from $PATH
-  if(NOT CUDA_HOST_COMPILER)
-    find_program(GCC_BIN gcc)
-    set(CUDA_HOST_COMPILER
-        ${GCC_BIN}
-        CACHE FILEPATH "" FORCE)
-  endif()
-
-  # various flags for PGI compiler with GPU build
-  if(${CMAKE_C_COMPILER_ID} STREQUAL "PGI" OR ${CMAKE_C_COMPILER_ID} STREQUAL "NVHPC")
-    # find_cuda produce verbose messages : use new behavior to use _ROOT variables
-    if(POLICY CMP0074)
-      cmake_policy(SET CMP0074 NEW)
-    endif()
-    find_package(CUDA 9.0 REQUIRED)
-    set(CUDA_SEPARABLE_COMPILATION ON)
-    set(CUDA_PROPAGATE_HOST_FLAGS OFF)
-    set(CUDA_PROFILING_DEF -DCUDA_PROFILING)
-    set(PGI_ACC_FLAGS "-acc -gpu=cuda${CUDA_VERSION}")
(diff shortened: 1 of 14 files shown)


alexsavulescu:
Please retest yo

alexsavulescu (approved):

PR #610, opened 2021-08-16
Update NVHPC warning suppressions

olupton:
**Description**
Update warning suppression compiler flags to work with modern versions of the NVHPC compilers. Fix some warnings at source. Bumps the submodule commits to include https://github.com/BlueBrain/mod2c/pull/66 and https://github.com/BlueBrain/nmodl/pull/716.

**How to test this?**
Compile with NVHPC 20.7+ and enjoy the lack of compiler warnings (?).

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.7
 - Version: master
 - Backend: GPU


coreneuron/permute/cellorder2.cpp
```diff
@@ -194,61 +159,6 @@ static void move_nodes(size_t start, size_t length, size_t dst, VTN& nodes) {
     }
 }
 
-// least number of nodes to move after nd to eliminate prace
-static size_t need2move(TNode* nd) {
-    size_t d = dist2child(nd);
-    return warpsize - ((nd->nodevec_index % warpsize) + d);
-}
-
-#if DEBUG
```
pramodk, line 203: @olupton : can I suggest to put `permute/cellorder*` related code removal in a separate PR or defer it? the reason is that it has some unused code with some macros like `DEBUG` but I believe there is some useful information in the routines and it will help to understand/debug this part of the code. (@kotsaloscv started it but we will ned some more time + efforts)

olupton: Sure, I changed it so this PR just moves some `#if DEBUG` around to squash the warnings.

pramodk (approved):
LGTM!

PR #613, opened 2021-08-18
Bump minimum supported CMake version to 3.15

alexsavulescu:
Following an internal discussion, we've decided to bump minimum CMake version to 3.15 for maintainability purposes.
Driven by #609.

CI_BRANCHES:NEURON_BRANCH=cmake315,

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 #
 # See top-level LICENSE file for details.
 # =============================================================================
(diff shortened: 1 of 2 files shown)


pramodk (approved):
NMODL can be also updated but that can happen later.

@olupton can confirm if we should do any other cmake changes as part of this PR.

alexsavulescu:
@olupton please update if necessary and merge at your convenience 🚀

olupton (approved):
LGTM. I don't see any urgent need to update the NMODL submodule.

Issue #611, opened 2021-08-18
Buffer TrajectoryRequests data on GPU


olupton:
https://github.com/BlueBrain/CoreNeuron/pull/528 introduces a new test (`test_netmove.py`) that records the values of a mechanism `RANGE` variable. This was not previously possible during GPU execution, because these values were not uploaded from GPU to host at each time step.

In https://github.com/BlueBrain/CoreNeuron/pull/528/commits/185a775ae8824d4905f94922deca1c102e67558d and https://github.com/BlueBrain/CoreNeuron/pull/528/commits/5780019e775bfece02fe3e9daa049cb62c975ee7 a minimal change was made to support recording such variables: https://github.com/BlueBrain/CoreNeuron/blob/5780019e775bfece02fe3e9daa049cb62c975ee7/coreneuron/sim/fadvance_core.cpp#L296-L325
but this is expected to carry a [simulation dependent, but] significant performance penalty.

This should be updated (after #528) so the recorded values are buffered on the GPU during the simulation and uploaded to the host afterwards.

cc: @pramodk

PR #616, opened 2021-08-23
Record variable trajectories on GPU

olupton:
**Description**
This updates the handling of variable trajectories in GPU builds. Previously, each value of interest was individually uploaded to the host every time step. Now the values are accumulated in GPU memory and uploaded at the end of the simulation, eliminating one `acc wait` and many `acc update self` clauses each time step.

Closes https://github.com/BlueBrain/CoreNeuron/issues/611.

**How to test this?**
Run the standard CTest test suite in a GPU build.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 21.7
 - Version: olupton/gpu-trajectories branch
 - Backend: GPU

### DIFF
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -400,6 +400,46 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
         } else {
             printf("\n WARNING: NrnThread %d not permuted, error for linear algebra?", i);
         }
+
+        {
+            TrajectoryRequests* tr = nt->trajec_requests;
+            if (tr) {
+                // Create a device-side copy of the `trajec_requests` struct and
+                // make sure the device-side NrnThread object knows about it.
+                auto* d_trajec_requests = reinterpret_cast<TrajectoryRequests*>(
+                    acc_copyin(tr, sizeof(TrajectoryRequests)));
+                acc_memcpy_to_device(&(d_nt->trajec_requests),
+                                     &d_trajec_requests,
+                                     sizeof(TrajectoryRequests*));
+                // Initialise the double** varrays member of the struct.
+                auto* d_tr_varrays = reinterpret_cast<double**>(
+                    acc_copyin(tr->varrays, sizeof(double*) * tr->n_trajec));
+                acc_memcpy_to_device(&(d_trajec_requests->varrays),
+                                     &d_tr_varrays,
+                                     sizeof(double**));
+                // Initialise the double** gather member of the struct.
+                auto* d_tr_gather = reinterpret_cast<double**>(
+                    acc_copyin(tr->gather, sizeof(double*) * tr->n_trajec));
+                acc_memcpy_to_device(&(d_trajec_requests->gather), &d_tr_gather, sizeof(double**));
+                // `varrays` and `gather` are minimal, the host-size buffers have
+                // `bsize` elements.
+                for (int i = 0; i < tr->n_trajec; ++i) {
+                    // tr->varrays[i] is a buffer of tr->bsize doubles on the host,
+                    // make a device-side copy of it and store a pointer to it in
+                    // the device-side version of tr->varrays.
+                    auto* d_buf_traj_i = reinterpret_cast<double*>(
+                        acc_copyin(tr->varrays[i], tr->bsize * sizeof(double)));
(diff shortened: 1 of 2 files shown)


coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -400,6 +400,46 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
...
+                        acc_copyin(tr->varrays[i], tr->bsize * sizeof(double)));
+                    acc_memcpy_to_device(&(d_tr_varrays[i]), &d_buf_traj_i, sizeof(double*));
+                    // tr->gather[i] is a double* referring to (host) data in the
+                    // (host) _data block
+                    auto* d_gather_i = acc_deviceptr(tr->gather[i]);
+                    acc_memcpy_to_device(&(d_tr_gather[i]), &d_gather_i, sizeof(double*));
+                }
+                // TODO: other members of the TrajectoryRequests struct:
+                // `scatter`, `n_pr`, `n_trajec`, `bsize`, `vsize` are not
+                // updated on device.
```
pramodk, line 440: scalar members are implicitly updated by the copy?
```suggestion
                // TODO: `scatter` member of the TrajectoryRequests struct
                // is not updated on device.
```

coreneuron/sim/fadvance_core.cpp
```diff
@@ -293,28 +293,29 @@ void nrncore2nrn_send_values(NrnThread* nth) {
...
+            // Note that this is rather inefficient: we generate one `acc update
+            // self` call for each `double` value (voltage, membrane current,
+            // mechanism property, ...) that is being recorded, even though in most
+            // cases these values will actually fall in a small number of contiguous
+            // ranges in memory. A better solution, if the performance of this
+            // branch becomes limiting, might be to offload this loop to the
+            // device and populate some `scatter_values` array there and copy it
+            // back with a single transfer. See also:
+            // https://github.com/BlueBrain/CoreNeuron/issues/611
             for (int i = 0; i < tr->n_trajec; ++i) {
```
pramodk, line 319: So - this part is still inefficient because we don't know "ranges" to update in one go and these needs to be sent back in every timestep, right?

pramodk (approved):
LGTM

coreneuron/gpu/nrn_acc_manager.cpp
olupton: Yes..it was a bad comment. There is also a `void** vpr` that was not mentioned in the comment.

For the scalar members: you're right these are automatically updated when we call `acc_copyin`. That said, `vsize` changes during the simulation and is not kept up to date on the device.

I'll update the comment...

coreneuron/sim/fadvance_core.cpp
olupton: Yes.

Not knowing which values are contiguous means we have `n_trajec` updates per time step instead of something like `n_properties_being_recorded` per time step (something like `n_trajec / n_sections` if variables are recorded on all sections/cells). We need to have **some** updates every timestep because apparently we have to call the callback every timestep.

Looking at this again this morning I think I need to re-add the `acc wait` I moved out of `nrn_fixed_step_lastpart` to be just before this loop, otherwise it might not pick up the correct values of mechanism data that is modified in `nrn_state` methods (inside `nonvint`).

I should also add that this branch is being modified "blind"...I am not aware of any test coverage of the `else if (tr->scatter) { ... }` part.

pramodk: > Looking at this again this morning I think I need to re-add the acc wait I moved out of nrn_fixed_step_lastpart to be just before this loop, otherwise it might not pick up the correct values of mechanism data that is modified in nrn_state methods (inside nonvint).

For the first loop / if block :
>  if (tr->varrays) {  // full trajectories into Vector data
> ...
>         tr->varrays[i][vs] = *tr->gather[i];

Yesterday it came to my mind I thought: `nrn_state` are updating data on GPU and the loop is copying that data into `tr->varrays[i][vs]` which is also on GPU. As the kernels are launched in same stream `async(nth->stream_id)`, they will be executed in order and hence will pick-up correct values. So there is no need for explicit wait ?

olupton: Indeed, I meant I needed to re-add the `acc wait` for the `else if(tr->scatter)` block: https://github.com/BlueBrain/CoreNeuron/blob/14b083632a92654954014efd7f28c4eb3d0294fc/coreneuron/sim/fadvance_core.cpp#L324-L333

Issue #603, opened 2021-07-26
Implementation of solve_interleaved1 and solve_interleaved2 using CUDA instead of  OpenACC ?


pramodk:
**Is your feature request related to a problem? Please describe.**

Currently we have [solve_interleaved1](https://github.com/BlueBrain/CoreNeuron/blob/7ea549590f15e80600ab0f61fa16ab9e9949ce1a/coreneuron/permute/cellorder.cpp#L654) and [solve_interleaved2](https://github.com/BlueBrain/CoreNeuron/blob/7ea549590f15e80600ab0f61fa16ab9e9949ce1a/coreneuron/permute/cellorder.cpp#L585) implementations using OpenACC. These routines can become significant bottleneck when there are smaller number of cells available to execute (e.g. using channel-benchmark with few thousand cells). 

**Describe the solution you'd like**

We haven't done detailed performance analysis yet (Olli/Christos should have more info here) but as OpenACC doesn't allow to tune kernels (e.g. memory types exposure), one of the thought is if these kernels can be efficiently implemented in CUDA.

**Describe alternatives you've considered**

Not really - do performance analysis of OpenACC implementation, do CUDA implementation and compare performance.

**Additional context**

How this could be tested? kernels are relatively small and should be straightforward to implement. Compile and link CUDA implementations to coreneuron library. See the existence of `solve_interleaved_launcher` and `ENABLE_CUDA_INTERFACE` in cellorder.cpp.

PR #618, opened 2021-08-25
Implementation of solve_interleaved2 using CUDA + Improvement of OpenACC counterpart (issue #603)

kotsaloscv:
**Description**

Implementation of solve_interleaved2 using CUDA + Improvement of OpenACC counterpart (issue #603)

The solve_interleaved1 should not be used since the algorithm is not optimized for GPUs. On the other hand, solve_interleaved2 is constructed with GPU-architecture in mind, i.e. the computations (forward/backward substitution on Hines matrices) are done in batches of warpsize(=32). This guarantees coalesced memory access and minimal branch divergence.

Regarding the openACC version of solve_interleaved2, I have forced the vector_length to be warpsize(=32) given the structure of the implementation. _This change resulted in 15-20% increase in performance._

Regarding the CUDA version of solve_interleaved2, I have tried many alternatives using shared/constant memory but because the algorithm is based on strided accesses of various matrices, these types of memories do not help. As is, the CUDA version has exactly the same performance as the OpenACC one. This branch is activated through the **--cuda-interface** cli argument.

**How to test this?**

```bash
cmake .. \
-DCMAKE_INSTALL_PREFIX=./install \
-DCORENRN_ENABLE_OPENMP=OFF \
-DCORENRN_ENABLE_GPU=ON \
-DCORENRN_ENABLE_NMODL=OFF \
-DCMAKE_C_COMPILER=nvc \
-DCMAKE_CXX_COMPILER=nvc++ \
-DCMAKE_CUDA_COMPILER=nvcc

make -j && make install

./bin/nrnivmodl-core **channel-benchmark/benchmark/channels/lib/modlib**
```

**Test System**
 - BB5

### DIFF
--- a/coreneuron/apps/main1.cpp
+++ b/coreneuron/apps/main1.cpp
@@ -151,6 +151,12 @@ void nrn_init_and_load_data(int argc,
                 "missing --gpu\n");
         exit(1);
     }
+    if (!corenrn_param.gpu && corenrn_param.cuda_interface) {
+        fprintf(stderr,
+                "compiled with OpenACC/CUDA does not allow the combination of --cuda-interface and "
+                "missing --gpu\n");
+        exit(1);
+    }
 #endif
 
 // if multi-threading enabled, make sure mpi library supports it
--- a/coreneuron/permute/cellorder.cpp
+++ b/coreneuron/permute/cellorder.cpp
@@ -568,10 +568,6 @@ static void bksub_interleaved2(NrnThread* nt,
     }
 }
 
-int temp1[1024] = {0};
-int temp2[1024] = {0};
-int temp3[1024] = {0};
-
 /**
  * \brief Solve Hines matrices/cells with compartment-based granularity.
  *
@@ -595,7 +591,6 @@ void solve_interleaved2(int ith) {
         solve_interleaved2_launcher(d_nt, d_info, ncore, acc_get_cuda_stream(nt->stream_id));
     } else {
 #endif
-        static int foo = 1;
         int* ncycles = ii.cellsize;         // nwarp of these
         int* stridedispl = ii.stridedispl;  // nwarp+1 of these
         int* strides = ii.stride;           // sum ncycles of these (bad since ncompart/warpsize)
@@ -625,9 +620,6 @@ void solve_interleaved2(int ith) {
             int lastroot = rootbegin[iwarp + 1];
             int firstnode = nodebegin[iwarp];
             int lastnode = nodebegin[iwarp + 1];
-// temp1[icore] = ic;
-// temp2[icore] = ncycle;
-// temp3[icore] = stride - strides;
 #if !defined(_OPENACC)
             if (ic == 0) {  // serial test mode. triang and bksub do all cores in warp
 #endif
@@ -640,13 +632,6 @@ void solve_interleaved2(int ith) {
 #ifdef _OPENACC
 #pragma acc wait(nt->stream_id)
 #endif
-        if (foo == 1) {
-            return;
-        }
-        foo = 0;
-        for (int i = 0; i < ncore; ++i) {
-            printf("%d => %d %d %d\n", i, temp1[i], temp2[i], temp3[i]);
-        }
 #ifdef _OPENACC
     }
 #endif
--- a/coreneuron/permute/cellorder.cpp
+++ b/coreneuron/permute/cellorder.cpp
@@ -588,11 +588,13 @@ void solve_interleaved2(int ith) {
 
     int ncore = nwarp * warpsize;
 
+#ifdef _OPENACC
     if (corenrn_param.gpu && corenrn_param.cuda_interface) {
         auto* d_nt = static_cast<NrnThread*>(acc_deviceptr(nt));
         auto* d_info = static_cast<InterleaveInfo*>(acc_deviceptr(interleave_info + ith));
         solve_interleaved2_launcher(d_nt, d_info, ncore, acc_get_cuda_stream(nt->stream_id));
     } else {
+#endif
         static int foo = 1;
         int* ncycles = ii.cellsize;         // nwarp of these
         int* stridedispl = ii.stridedispl;  // nwarp+1 of these
@@ -645,7 +647,9 @@ void solve_interleaved2(int ith) {
         for (int i = 0; i < ncore; ++i) {
             printf("%d => %d %d %d\n", i, temp1[i], temp2[i], temp3[i]);
         }
+#ifdef _OPENACC
     }
+#endif
 }
 
 /**
--- a/coreneuron/permute/cellorder.cpp
+++ b/coreneuron/permute/cellorder.cpp
@@ -15,6 +15,7 @@
 #include "coreneuron/network/tnode.hpp"
 #include "coreneuron/utils/lpt.hpp"
 #include "coreneuron/utils/memory.h"
+#include "coreneuron/apps/corenrn_parameters.hpp"
 
 #include "coreneuron/permute/node_permute.h"  // for print_quality
 
@@ -567,10 +568,6 @@ static void bksub_interleaved2(NrnThread* nt,
     }
 }
 
-#ifdef ENABLE_CUDA_INTERFACE
-void solve_interleaved_launcher(NrnThread* nt, InterleaveInfo* info, int ncell);
-#endif
-
 int temp1[1024] = {0};
 int temp2[1024] = {0};
 int temp3[1024] = {0};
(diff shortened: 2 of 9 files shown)


coreneuron/permute/cellorder.cpp
```diff
@@ -601,13 +601,19 @@ void solve_interleaved2(int ith) {
 #endif
 
     int ncore = nwarp * warpsize;
+
+#ifdef ENABLE_CUDA_INTERFACE
+    NrnThread* d_nt = (NrnThread*) acc_deviceptr(nt);
+    InterleaveInfo* d_info = (InterleaveInfo*) acc_deviceptr(interleave_info + ith);
+    solve_interleaved2_launcher(d_nt, d_info, ncore);
+#else
```
pramodk, line 609: IMO, we should remove this `ENABLE_CUDA_INTERFACE` compile time switch and make the possibility to launch this via CLI option at runtime.

coreneuron/permute/cellorder.cpp
```diff
@@ -601,13 +601,19 @@ void solve_interleaved2(int ith) {
...
 #ifdef _OPENACC
     // clang-format off
-
-    #pragma acc parallel loop present(                  \
-        nt[0:1], strides[0:nstride],                    \
-        ncycles[0:nwarp], stridedispl[0:nwarp+1],       \
-        rootbegin[0:nwarp+1], nodebegin[0:nwarp+1])     \
+    
+    #pragma acc parallel loop gang vector vector_length(warpsize) \
```
pramodk, line 613: Just curious: how this was parallelising before by default? (i.e. `-Minfo=accel` output)

coreneuron/permute/cellorder.cu
```diff
@@ -0,0 +1,101 @@
+#ifdef ENABLE_CUDA_INTERFACE
+
```
pramodk, line 2: nitpick: to remove; we can compile cuda file in GPU build and use CLI option as mentioned before.

coreneuron/permute/cellorder.cu
```diff
@@ -0,0 +1,101 @@
...
+//#include <iostream>
+//#include <stdio.h>
```
pramodk, line 4: nitpick: remove

coreneuron/permute/cellorder.cu
```diff
@@ -0,0 +1,101 @@
...
+#include "coreneuron/network/tnode.hpp"
+#include "coreneuron/sim/multicore.hpp"
+
+namespace coreneuron {
+
+#define GPU_A(i)      nt->_actual_a[i]
+#define GPU_B(i)      nt->_actual_b[i]
+#define GPU_D(i)      nt->_actual_d[i]
+#define GPU_RHS(i)    nt->_actual_rhs[i]
+#define GPU_PARENT(i) nt->_v_parent_index[i]
```
pramodk, line 16: nitpick: this is older pattern; as we want to write code with less macros, would be nice to avoid this?

coreneuron/permute/cellorder.hpp
```diff
@@ -12,6 +12,12 @@
 #include "coreneuron/utils/memory.h"
 #include <algorithm>
 namespace coreneuron {
+
+#ifdef ENABLE_CUDA_INTERFACE
```
pramodk, line 16: same

pramodk (changes requested):
looks good; have suggested small improvements / nitpicks

coreneuron/permute/cellorder.cu
```diff
@@ -0,0 +1,101 @@
...
+    int lastroot = rootbegin[iwarp + 1];
+    int firstnode = nodebegin[iwarp];
+    int lastnode = nodebegin[iwarp + 1];
+
+    triang_interleaved2_device(nt, ic, ncycle, stride, lastnode);
+    bksub_interleaved2_device(nt, root + ic, lastroot, ic, ncycle, stride, firstnode);
+}
+
+void solve_interleaved2_launcher(NrnThread* nt, InterleaveInfo* info, int ncore) {
+    cudaDeviceSynchronize();
```
olupton, line 91: Not a blocker here, because we do the same thing elsewhere, but maybe we should only be waiting on the device stream corresponding to `nt->stream_id`? There is an OpenACC API method to help with this, `acc_get_cuda_stream`.

olupton (changes requested):
LGTM in general. Would be good to make it a runtime switch as @pramodk suggested so we can test this without having to add new builds.

IMO we should not add this without at least some tests that use it.

coreneuron/apps/main1.cpp
```diff
@@ -191,6 +191,12 @@ void nrn_init_and_load_data(int argc,
                 "missing --gpu\n");
         exit(1);
     }
+    if (!corenrn_param.gpu && corenrn_param.cuda_interface) {
+        fprintf(stderr,
+                "compiled with OpenACC/CUDA does not allow the combination of --cuda-interface and "
+                "missing --gpu\n");
+        exit(1);
+    }
```
olupton, line 199: I'm surprised we're not hitting this, as the new `cudaInterface` tests seem to be running in non-GPU builds (that do not pass `--gpu`).

tests/integration/CMakeLists.txt
```diff
@@ -23,11 +24,13 @@ set(TEST_CASES_WITH_ARGS
     "ring_spike_buffer!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_spike_buffer --spikebuf 1"
     "ring_permute1!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute1 ${PERMUTE1_ARGS}"
     "ring_permute2!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute2 ${PERMUTE2_ARGS}"
+    "ring_permute2_cudaInterface!${RING_COMMON_ARGS} ${GPU_ARGS} --outpath ${CMAKE_CURRENT_BINARY_DIR}/ring_permute2_cudaInterface ${PERMUTE2_ARGS} ${CUDA_INTERFACE}"
```
olupton, line 27: I think we should disable these `CUDA_INTERFACE` tests in non-GPU builds.

coreneuron/permute/cellorder.cpp
```diff
@@ -583,65 +576,65 @@ int temp3[1024] = {0};
...
-    int stream_id = nt->stream_id;
-#endif
 
     int ncore = nwarp * warpsize;
+
 #ifdef _OPENACC
-    // clang-format off
+    if (corenrn_param.gpu && corenrn_param.cuda_interface) {
+        auto* d_nt = static_cast<NrnThread*>(acc_deviceptr(nt));
+        auto* d_info = static_cast<InterleaveInfo*>(acc_deviceptr(interleave_info + ith));
```
olupton, line 590: I haven't checked locally [yet], but I suspect the errors in the unified memory Jenkins build might be because `nullptr` is returned here.

olupton (commented):
Looking good, I think there are just a few (hopefully minor) loose ends to tidy up.

kotsaloscv:
retest please

iomaganaris (approved):

coreneuron/permute/cellorder.cu
```diff
@@ -0,0 +1,99 @@
...
+void solve_interleaved2_launcher(NrnThread* nt, InterleaveInfo* info, int ncore, void* stream) {
+    auto cuda_stream = static_cast<cudaStream_t>(stream);
+
+    int threadsPerBlock = warpsize;
+    // TODO: Should blocksPerGrid be a fixed number and have a while block inside the kernel?
+    int blocksPerGrid = (ncore + threadsPerBlock - 1) / threadsPerBlock;
+
+    solve_interleaved2_kernel<<<blocksPerGrid, threadsPerBlock, 0, cuda_stream>>>(nt, info, ncore);
+
+    cudaStreamSynchronize(cuda_stream);
```
pramodk, line 96: @kotsaloscv : Could we do error checking here? to make sure the kernels are executed and avoid silent errors...

pramodk (commented):
One small change for error checking and we are good to go!

coreneuron/permute/cellorder.cu
kotsaloscv: @pramodk : I just added a check. Shall we move the check in the utils folder for future CUDA-related development?

pramodk: Sure! sounds good! (realised `coreneuron/utils/randoms/nrnran123.cu` also has some cuda calls without error checking)

kotsaloscv: regarding `coreneuron/utils/randoms/nrnran123.cu` shall I open a ticket or include a fix in this PR?

pramodk: lets do that separately. Go ahead with a ticket and then one can look at other parts of the code as well.

pramodk (approved):
LGTM

PR #656, opened 2021-10-11
Check mechanisms in direct mode

### LINKED ISSUE #636: Check mechanism details between NEURON and CoreNEURON in direct mode
Similar to the file mode transfer of data between `NEURON` and `CoreNEURON` we should check that the mechanism data are the same between the generated file of `NEURON` and `CoreNEURON`. This currently done using the `check_mechanism()` function in https://github.com/BlueBrain/CoreNeuron/blob/b28a00f8e16434c1a221c4166b86c36cdc2397f7/coreneuron/io/phase2.cpp#L411
Otherwise it's possible in direct mode transfer to pass wrong data to the pointers of the instance struct of different mechanisms.
**Describe the solution you'd like**
Run the `check_mechanism()` function also in direct mode.

alkino:
Fix #636

### DIFF
--- a/coreneuron/io/phase2.cpp
+++ b/coreneuron/io/phase2.cpp
@@ -259,6 +259,7 @@ void Phase2::read_direct(int thread_id, const NrnThread& nt) {
     nodecounts = std::vector<int>(nodecounts_, nodecounts_ + n_mech);
     delete[] nodecounts_;
 
+    check_mechanism();
 
     // TODO: fix it in the future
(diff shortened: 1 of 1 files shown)


alkino:
tested locally it works:
```
Error: ExpSyn is a different MOD file than used by NEURON!
Error : NEURON and CoreNEURON must use same mod files for compatibility, 1 different mod file(s) found. Re-compile special and special-core!
```

PR #666, opened 2021-10-19
Fix reading spike info in report.conf in case of 0 voltage report

WeinaJi:
Fixes # (issue)

Fix reading spike info in report.conf in case of 0 voltage report

**How to test this?**
SimulationStack


WeinaJi:
@jorblancoa  Thanks for the approval. Do you have the permission to merge this PR? I cannot do it.

Issue #670, opened 2021-10-20
test/coreneuron/test_direct.py fails with GPU based installation via wheel 


pramodk:
**Describe the issue**

When we use gpu enabled neuron wheel produced via Azure, [test/coreneuron/test_direct.py](https://github.com/neuronsimulator/nrn/blob/master/test/coreneuron/test_direct.py) fails.

**To Reproduce**

* Download the attached wheel [NEURON_nightly-8.0a.702.gpu-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.zip](https://github.com/BlueBrain/CoreNeuron/files/7382418/NEURON_nightly-8.0a.702.gpu-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.zip) (you can also grab from Azure [pipeline](https://dev.azure.com/neuronsimulator/nrn/_build/results?buildId=4141&view=artifacts&type=publishedArtifacts) but the artefacts will be deleted in a week or two)
* Install on BB5

```console
salloc --partition=prod --account=proj7 --nodes=1 --time=8:00:00 --exclusive --constraint=volta --ntasks-per-node=1
module load unstable python-dev/0.4
pip install --upgrade pip
pip install wheelhouse/NEURON_nightly-8.0a.702.gpu-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

Create special:

```console
module load unstable nvhpc/21.2
cd nrn # repo
nrnivmodl -coreneuron test/coreneuron/mod/
```

And run the test on CPU:

```console
$ ./x86_64/special test/coreneuron/test_direct.py
INFO : Using neuron-nightly Package (Developer Version)
Warning: no DISPLAY environment variable.
--No graphics will be displayed.
NEURON -- VERSION 8.0a-702-g50eb265ea+ epic/gpu_wheel (50eb265ea+) 2021-10-19
Duke, Yale, and the BlueBrain Project -- Copyright 1984-2021
See http://neuron.yale.edu/neuron/credits

Additional mechanisms from files
 "test/coreneuron/mod//fornetcon.mod" "test/coreneuron/mod//invlfire.mod" "test/coreneuron/mod//netmove.mod" "test/coreneuron/mod//sample.mod" "test/coreneuron/mod//watchrange.mod"
Traceback (most recent call last):
  File "test/coreneuron/test_direct.py", line 98, in <module>
    test_direct_memory_transfer()
[...]

PR #671, opened 2021-10-20
Update mod2c submodule to include #72

olupton:
**Description**
Update submodule to include https://github.com/BlueBrain/mod2c/pull/72, which should fix https://github.com/BlueBrain/CoreNeuron/issues/670.

The branch of https://github.com/neuronsimulator/nrn/pull/1510 is being used so that the CI is sensitive to the issue being fixed.

CI_BRANCHES:NEURON_BRANCH=olupton/test-gpu-builds-on-cpu,


pramodk:
As BB5 Jenkins is down, I will go ahead with the merge. I tested this as well and now wheel tests pass without skipping imembrane comparison.

PR #685, opened 2021-10-29
Drop MUT* macros and C support.

olupton:
**Description**
The `MUT*` macros were only used in one place, and I believe the 0 in `MUTCONSTRUCT(0)` meant that the mutex was never actually used.

This is partly motivated by having been led on a wild goose chase by these macros when debugging another issue...


ohm314:
Should we not anyway redo those queues completely using stdlib queues and a proper C++ implementation? Also, probably a stupid question, but would running some tests with ThreadSanitizer help here?

alkino (approved):

pramodk:
> TQueue in coreneuron/network/tqueue.ipp seems to be derived from src/nrncvode/spt2queue.cpp in neuron. It doesn't have code related to mutex locks

I was wrong with the above comment - the TQueue implementation is derived from [nrncvode/sptbinq.cpp](https://github.com/neuronsimulator/nrn/blob/master/src/nrncvode/sptbinq.cpp) which has locks.

The issue / inconsistency in CoreNEURON was that the mutex was never constructed (`0` as an argument to  MUTCONSTRUCT):

```cpp
template <container C>
TQueue<C>::TQueue() {
    MUTCONSTRUCT(0)
    nshift_ = 0;
```

whereas neuron has constructor with a default argument `mkmut=0`:

```cpp
class TQueue {
public:
    TQueue(TQItemPool*, int mkmut = 0);
....

TQueue::TQueue(TQItemPool* tp, int mkmut) {
	MUTCONSTRUCT(mkmut)
	tpool_ = tp;
```

In the neuron source tree I don't see any TQueue instantiation with `mkmut=1`:

```cpp
kumbhar@bbpv1:~/.../nrn/src$ grep -r "new TQueue" *
nrncvode/netcvode.cpp:	tqe_ = new TQueue(tpool_, 0);
nrncvode/netcvode.cpp:			d.tq_ = new TQueue(d.tpool_);
nrncvode/netcvode.cpp:		p.lcv_[i].neosim_self_events_ = new TQueue();
nrncvode/netcvode.cpp:		d.tqe_ = new TQueue(p[i].tpool_);
nrncvode/tqueue.cpp:	TQueue* q = new TQueue(0);
```

Does this mean we can remove all `MUT*` macros in `TQueue`? I would say YES.

There was one more thing where I saw `mkmut=1` and that is in `TQItemPool`:

```cpp
$ grep -r TQItemPool *
....
nrncvode/netcvode.h:	TQItemPool* tpool_;
nrncvode/netcvode.cpp:	tpool_ = new TQItemPool(1000, 1);


nrncvode/tqueue.h:declarePool(TQItemPool, TQItem)

# and in nrncvode/pool.h

#define declarePool(Pool,T) \
class Pool { \
public: \
    Pool(long count, int mkmut = 0); \

```
but this doesn't interact with `TQueue`. 

With above summary, I had quick call with @nrnhines to verify the origin of mutex macros and if the removal of those is safe to do. He pointed out that the next/newer iteration of original per queue mutex was [`inter-thread-event` buffers](https://github.com/BlueBrain/CoreNeuron/blob/ff93a938cd56ca4e0b2c343f57f6ed39959916da/coreneuron/network/netcvode.hpp#L50) i.e. instead of copying events to TQueue using mutex, there is a  buffer which is populated by the thread who owns the `PreSyn` (see [interthread_send()](https://github.com/BlueBrain/CoreNeuron/blob/ff93a938cd56ca4e0b2c343f57f6ed39959916da/coreneuron/network/netcvode.cpp#L136)) and this is where mutex is required/used. The thread who owns the queue will then copy all those buffered event into it's queue (and hence mutex is not required now). In [PreSyn::send()](https://github.com/BlueBrain/CoreNeuron/blob/ff93a938cd56ca4e0b2c343f57f6ed39959916da/coreneuron/network/netcvode.cpp#L408) we can see how thread decide to put event into it's queue or inter-thread-event buffer.

This also explains why coreneuron always had `MUTCONSTRUCT(0)`.

In summary, the merge of this PR is safe.

> Should we not anyway redo those queues completely using stdlib queues and a proper C++ implementation?

I believe so and this was what attempted by Tim already with the use of [std::priority_queue](https://github.com/BlueBrain/CoreNeuron/blob/ff93a938cd56ca4e0b2c343f57f6ed39959916da/coreneuron/network/tqueue.hpp#L151). This implementation is not supporting moving events which is used by `net_move()` NMODL constructs . (I vaguely remember some perf/efficiently related discussion about this with default SplayTree implementation).

> Also, probably a stupid question, but would running some tests with ThreadSanitizer help here?

It would but such test is missing. I am creating ticket in neuron to adapt [tqperf](https://github.com/nrnhines/tqperf) model with threads so that it can be used for such validations.

PR #694, opened 2021-11-23
[Hackathon] disable a lot of CI

olupton:
**Description**
Disable various CI checks. The `hackathon_main` target branch is now blacklisted on the Jenkins side.


iomaganaris (approved):
LGTM
Maybe we could blacklist also `hackathon/*` so that we can also create and merge branches that start with this like `hackathon/magkanar/new_feature` with other similar ones?

PR #695, opened 2021-11-23
[Hackathon] Add a temporary option for benchmark data.

olupton:
**Description**
Adds an option `CORENRN_EXTERNAL_BENCHMARK_DATA` that we can point to a shared directory on Permutter/Ascent/BB5. A draft directory on BB5 is `/gpfs/bbp.cscs.ch/project/proj16/olupton/nersc-gpu-hackathon-dec-2021`, which exists (possibly out of date) at `$CFS/ntrain9/neuron/nersc-gpu-hackathon-dec-2021` on Perlmutter and `/ccsopen/proj/gen170/neuron/nersc-gpu-hackathon-dec-2021` on Ascent.

If this is available, building CoreNEURON will also create a `special-core` binary that can execute the `channel-benchmark` input data provided in `CORENRN_EXTERNAL_BENCHMARK_DATA`.

### DIFF
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -310,10 +310,15 @@ if(NOT ${CORENRN_EXTERNAL_BENCHMARK_DATA} STREQUAL "")
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/benchmark
     COMMENT "Running nrnivmodl-core for channel-benchmark mechanisms")
   list(APPEND all_output_binaries ${output_binaries})
-  add_test(
-    NAME benchmark
-    COMMAND "${CMAKE_BINARY_DIR}/benchmark/${CMAKE_SYSTEM_PROCESSOR}/special-core" --datpath
-            "${CORENRN_EXTERNAL_BENCHMARK_DATA}/channel-benchmark-all-440-cells-2-ranks" --tstop 1)
+  string(
+    CONCAT
+      benchmark_command
+      "'${CMAKE_BINARY_DIR}/benchmark/${CMAKE_SYSTEM_PROCESSOR}/special-core'"
+      " --datpath '${CORENRN_EXTERNAL_BENCHMARK_DATA}/channel-benchmark-all-440-cells-2-ranks'"
+      " --tstop 1 &&"
+      "diff out.dat '${CORENRN_EXTERNAL_BENCHMARK_DATA}/channel-benchmark-all-440-cells-2-ranks.gpu.spikes'"
+  )
+  add_test(NAME benchmark COMMAND sh -c "${benchmark_command}")
 endif()
 set(modfile_directory "${CORENEURON_PROJECT_SOURCE_DIR}/tests/integration/ring_gap/mod")
 file(GLOB modfiles "${modfile_directory}/*.mod")
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -293,6 +293,28 @@ set_target_properties(
 # =============================================================================
 # create special-core with halfgap.mod for tests
 # =============================================================================
+set(all_output_binaries)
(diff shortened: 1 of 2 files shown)


coreneuron/CMakeLists.txt
```diff
@@ -293,6 +293,28 @@ set_target_properties(
...
+    DEPENDS scopmath coreneuron ${NMODL_TARGET_TO_DEPEND} ${modfiles} ${CORENEURON_BUILTIN_MODFILES}
+    COMMAND ${CMAKE_BINARY_DIR}/bin/nrnivmodl-core -b STATIC -m ${CORENRN_MOD2CPP_BINARY} -p 1
+            "${modfile_directory}"
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/benchmark
+    COMMENT "Running nrnivmodl-core for channel-benchmark mechanisms")
+  list(APPEND all_output_binaries ${output_binaries})
+  add_test(
+    NAME benchmark
+    COMMAND "${CMAKE_BINARY_DIR}/benchmark/${CMAKE_SYSTEM_PROCESSOR}/special-core" --datpath
+            "${CORENRN_EXTERNAL_BENCHMARK_DATA}/channel-benchmark-all-440-cells-2-ranks" --tstop 1)
```
olupton, line 316: Thoughts on whether it's worth adding an automatic comparison to reference spikes here?

iomaganaris: It's not a bad idea to avoid having to run all the cases of the `channel-benchmark` in case we want to check the validity of our changes

iomaganaris (approved):
LGTM
Just one question. How long does this take to run? Should we also include ringtest in case it runs faster? Otherwise for validation of standard functionality I don't think it adds any benefit and we can keep only channel-benchmark

kotsaloscv (approved):
LGTM

olupton:
> LGTM Just one question. How long does this take to run? Should we also include ringtest in case it runs faster? Otherwise for validation of standard functionality I don't think it adds any benefit and we can keep only channel-benchmark

On CPU it takes ~8s (setup ~ simulation), on GPU it's ~25s (setup >> simulation, we should fix that 😅).
I didn't remove anything related to the ringtest, so it's easy to run that too if we want to.

coreneuron/CMakeLists.txt
olupton: OK, I'll add a spike file to `CORENRN_EXTERNAL_BENCHMARK_DATA` and a `diff` to the test command.

iomaganaris:
> > LGTM Just one question. How long does this take to run? Should we also include ringtest in case it runs faster? Otherwise for validation of standard functionality I don't think it adds any benefit and we can keep only channel-benchmark
> 
> On CPU it takes ~8s (setup ~ simulation), on GPU it's ~25s (setup >> simulation, we should fix that sweat_smile) on a Perlmutter node.
> 
> I didn't remove anything related to the ringtest, so it's easy to run that too if we want to.

By setup you mean the NEURON setup or loading the coredat files?

olupton:
> > > LGTM Just one question. How long does this take to run? Should we also include ringtest in case it runs faster? Otherwise for validation of standard functionality I don't think it adds any benefit and we can keep only channel-benchmark
> > 
> > 
> > On CPU it takes ~8s (setup ~ simulation), on GPU it's ~25s (setup >> simulation, we should fix that sweat_smile) on a Perlmutter node.
> > I didn't remove anything related to the ringtest, so it's easy to run that too if we want to.
> 
> By setup you mean the NEURON setup or loading the coredat files?

Loading the `coredat` files, which are about 1GB. I think the GPU one is slower because there are currently a lot of tiny unified memory allocations for Random123.

coreneuron/CMakeLists.txt
olupton: Done in 84a7f50b.

PR #699, opened 2021-11-29
Give possibility to change mpiexec configuration for integration tests

alkino:
Some mpiexec version does not have the option -n, so it should be configurable


pramodk:
@alkino : could you check failing CI?

pramodk:
@alkino : converting this to draft until CI is fixed. If you think this is not relevant now, feel free to close this.

PR #701, opened 2021-11-29
BinQ and spike compression fixes

nrnhines:
See neuronsimulator/nrn#1548
Superceded by neuronsimulator/nrn#1556

CI_BRANCHES:NEURON_BRANCH=hines/binq-fix-rebase,

- [x] Merge https://github.com/nrnhines/tqperf/pull/6
- [x] Merge https://github.com/BlueBrain/nmodl/pull/791 and update submodule


coreneuron/apps/main1.cpp
```diff
@@ -467,6 +467,8 @@ static void* load_dynamic_mpi(const std::string& libname) {
 #endif
 
 extern "C" void mk_mech_init(int argc, char** argv) {
+    corenrn_param.multisend = false;
+    corenrn_param.binqueue = false;
```
nrnhines, line 471: Originally hoping for something like ```corenrn_param = corenrn_parameters()``` but that is not allowed and I didn't want to convert to a pointer.

coreneuron/network/netcvode.cpp
```diff
@@ -232,7 +236,7 @@ void NetCvode::clear_events() {
         d.unreffed_event_cnt_ = 0;
         d.inter_thread_events_.clear();
         d.tqe_->nshift_ = -1;
-        d.tqe_->shift_bin(nrn_threads->_t);
+        d.tqe_->shift_bin(nrn_threads->_t - 0.5 * nrn_threads->_dt);
```
nrnhines, line 239: There is a similar statement in the following function I did not change. A specific test would have to be written to see if there could be an issue with file mode demonstrating the need for an equivalent change.

coreneuron/network/netpar.cpp
```diff
@@ -366,6 +369,7 @@ void nrn_spike_exchange(NrnThread* nt) {
             ps->send(spikein[i].spiketime, net_cvode_instance, nt);
         }
     }
+    nrn_multithread_job(interthread_enqueue);
```
nrnhines, line 372: I'm not sure this change is needed. But it can't hurt.

coreneuron/network/netcvode.cpp
nrnhines: The corresponding change was made in NetCvode::init_events.  In retrospect, it is likely that BinQ is not supported for ```--restore``` since there is no mention of ```shift_bin``` in the ```nrn_checkpoint``` files. (Analogous to ```shift_bin``` in nrn/src/nrniv/bbsavestate.cpp)

coreneuron/utils/randoms/nrnran123.cu
```diff
@@ -179,7 +179,7 @@ void nrnran123_set_globalindex(uint32_t gix) {
     // If the global seed is changing then we shouldn't have any active streams.
     {
         std::lock_guard<OMP_Mutex> _{g_instance_count_mutex};
-        if (g_instance_count != 0 && nrnmpi_myid == 0) {
+        if (g_instance_count != 0 && nrnmpi_myid == 0 && get_global_state().v[0] != gix) {
```
olupton, line 182: Is stdout unacceptably verbose without this change?

When I added this warning then I didn't add this check because I thought it increased the chances of hiding bugs behind lucky choices of random seeds.

olupton (commented):
First pass through this changeset on its own looks good, I'll move on to the NEURON side.

alexsavulescu (approved):
LGTM. Maybe merge Olli's change from #720 before.

alexsavulescu:
Please retest

coreneuron/utils/randoms/nrnran123.cu
nrnhines: The change is not really necessary because the item that was causing the warning has been fixed. ie use of a random123 instance to help with multisend two phase performance. The worry about "hiding bugs" is valid since in the absence of the multisend two phase fix, multiple psolve would have different transfer patterns, and I wouldn't have noticed that was happening. So I will revert line 182 above.

nrnhines: By the way. On the NEURON side the old "nondeterminism" still occurs but it is perhaps not a bug since the transfer patterns are different only if the model changes, and not different  merely because psolve is called multiple times.
Nevertheless it might be reasonable to backport the multisend two phase random123 fix  from CoreNEURON to NEURON.

nrnhines:
> LGTM. Maybe merge Olli's change from #720 before.

Yes.  But is this best done after #720 is merged to master?

alexsavulescu:
> Yes. But is this best done after #720 is merged to master?

Indeed

nrnhines:
please retest

nrnhines:
@alexsavulescu I notice ```test:neuron:intel``` has (among 9 test failures) a tqperf problem ```MPT ERROR: invalid option: --oversubscribe```. It occurs to me that in nrn/test/external/tqperf/CMakeLists.txt the ```tqperf_neuron_prefix``` may be misleading as it is not used in ```COMMAND bash "./ci-test.sh"```. In the latter it is tempting to just replace ```${MPIEXEC_OVERSUBSCRIBE---oversubscribe}``` with ```${MPIEXEC_OVERSUBSCRIBE}``` but I'm wondering if the full ```${tqperf_neuron_prefix} should be somehow passed into ci-test.sh in case those special flags are needed in some cases.

alexsavulescu:
https://github.com/nrnhines/tqperf/pull/5

nrnhines:
> ```${MPIEXEC_OVERSUBSCRIBE---oversubscribe}```
But I was thinking the problem is that the test in nrn/test/CMakeLists.txt  unsets using
```
  if(NOT MPIEXEC_OVERSUBSCRIBE_TEST EQUAL 0)
    message(STATUS "mpiexec does not support ${MPIEXEC_OVERSUBSCRIBE}")
    unset(MPIEXEC_OVERSUBSCRIBE)
  endif()
```
so that the ci_test.sh file would do exactly the wrong thing if ```--oversubscribe``` is not a valid argument to mpiexec.

alexsavulescu:
It is my understanding that we test via CMake if `oversubscribe` works with the current MPI library and sets or unsets `MPIEXEC_OVERSUBSCRIBE`. That is then passed to the test itself via `MPIEXEC_OVERSUBSCRIBE=${MPIEXEC_OVERSUBSCRIBE}` https://github.com/neuronsimulator/nrn/pull/1556/commits/a43bd8bc66290fa98f18b450c88d72f5199656d8

So if `oversubscribe` is not available, then we will have an empty `$MPIEXEC_OVERSUBSCRIBE` in the test scope. 

With the changeset above, we now hit: 
```
+ mpiexec -n 16 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P29957/J111765/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv -mpi -python test1.py
MPT ERROR: Not enough slots from job scheduler for requested ranks
```

whereas without the  `${MPIEXEC_OVERSUBSCRIBE---oversubscribe}` we previously had: 
```
 mpiexec -oversubscribe -n 16 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P29772/J110989/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv -mpi -python test1.py
MPT ERROR: invalid option: -oversubscribe
```

Looks like we need another strategy when we have HPE MPT

alexsavulescu:
Also got a bit puzzled by `-` vs `--`, but we have the same error: 

```
+ mpiexec --oversubscribe -n 16 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P29811/J111242/_/spack-build/spack-stage-neuron-develop-mvn7s3za436y6xrhbspwpy7pbewubnvm/spack-build-mvn7s3z/bin/nrniv -mpi -python test1.py
MPT ERROR: invalid option: --oversubscribe
```

nrnhines:
> Looks like we need another strategy when we have HPE MPT

Do I have this right?  mpiexec on that machine does not accept ```--oversubscribe``` so ci-test.sh sees an empty ```${MPIEXEC_OVERSUBSCRIBE```. But ```-n 16``` is too many for that machine.

Note that 16 ranks are used solely for complete coverage of the phase2 multisend code.  Perhaps we can skip the test if oversubscribe is not allowed and 16 real cores are not available.

alexsavulescu:
> Do I have this right? mpiexec on that machine does not accept `--oversubscribe` so ci-test.sh sees an empty `${MPIEXEC_OVERSUBSCRIBE`. But `-n 16` is too many for that machine.

Looks that way to me. 

> Note that 16 ranks are used solely for complete coverage of the phase2 multisend code. Perhaps we can skip the test if oversubscribe is not allowed and 16 real cores are not available.

Well we only care about coverage in the NEURON Code Coverage CI. I guess we can skip the test in the CoreNEURON CI based on some condition, i,e. if `--oversubscribe` is not available.

nrnhines:
> skip the test in the CoreNEURON CI based on some condition, i,e. if --oversubscribe is not available.

That's reasonable.  Or run with ```-n 2``` if not available. Although the coverage won't be as good, many of the tests are still substantive.

alexsavulescu:
Was thinking about something along the lines of: https://github.com/neuronsimulator/nrn/pull/1556/commits/e686c1d66f6be51f7409fcca996c6486c63b7bac

olupton (approved):

alexsavulescu:
Now we get 
```
CoreNEURON single and multiple threads
4 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P30108/J112635/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv: Could not find CoreNEURON library
4  near line 0
4  ^
        4 ParallelContext[0].psolve(50)
MPT ERROR: Rank 4(g:4) is aborting with error code -1.
	Process ID: 51011, Host: ldir01u05.bbp.epfl.ch, Program: /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P30108/J112635/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv
	MPT Version: HPE HMPT 2.25  10/22/21 03:18:39

MPT: --------stack traceback-------
13 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P30108/J112635/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv: Could not find CoreNEURON library
13  near line 0
13  ^
        13 ParallelContext[0].psolve(50)
12 /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P30108/J112635/_/spack-build/spack-stage-neuron-develop-vkogqddlipn6wfzq2o4mkbt36t7uknzy/spack-build-vkogqdd/bin/nrniv: Could not find CoreNEURON library
```

pramodk:
```
/nrniv: Could not find CoreNEURON library
```
@alexsavulescu : In case of GPU build, coreneuron is a static library and hence we have to use `special` to launch the simulation. `nrniv` can not be used to run coreneuron simulation.

pramodk:
```console
/gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcproj12/P30438/J115285/_/spack-build/spack-stage-neuron-develop-fwu47zj5ksgprg63usqqvmnfvuvjnlmc/spack-build-fwu47zj/share/nrn/nrnmain.cpp:
622ptxas warning : Conflicting options --device-debug and --generate-line-info specified, ignoring --generate-line-info option
623ptxas warning : Conflicting options --device-debug and --generate-line-info specified, ignoring --generate-line-info option
624Successfully created x86_64/special
625++ find . -type f -name special -print -quit
626+ special_exe=./x86_64/special
627+ mpiexec -n 16 ./x86_64/special -mpi -python test1.py
628MPT ERROR: Not enough slots from job scheduler for requested ranks
629	(HPE HMPT 2.25  10/22/21 03:19:55)
```
I thought this was fixed. Will take a look soon...

nrnhines:
My intention for a substantive gpu test is to allow use of a POINT_PROCESS version of ```ARTIFICIAL_CELL IntervalFireSHA```, assuming I can arrange for a gpu version of ```#include <openssl/sha.h>``` (though we could drop back to the less strict test using ```IntervalFire``` which only counts input and output events without checking event times.).  Anyway, that will provide good performance and validation of the spike buffering algorithms on the GPU.

pramodk:
GPFS finally reasonably working:
image

PR #705, opened 2021-12-09
Implement omp_get_mapped_ptr until it is available

alkino:
omp_get_mapped_ptr is only for omp >= 5.1
gcc and clang still does not implement it

### DIFF
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -188,13 +188,10 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
             int szdp = corenrn.get_prop_dparam_size()[type];
             int is_art = corenrn.get_is_artificial()[type];
 
-            if (is_art) {
-                // data has been allocated somewhere else
-                dptr = cnrn_target_copyin(tml->ml->data);
-            } else {
-                // get device pointer for corresponding mechanism data
-                dptr = cnrn_target_deviceptr(tml->ml->data);
-            }
+            // If the mechanism is artificial data are not inside nt->_data but in a newly
+            // allocated block. As we never run code for artificial cell inside GPU
+            // we don't copy it.
+            dptr = is_art ? nullptr : cnrn_target_deviceptr(tml->ml->data);
             cnrn_target_memcpy_to_device(&(d_ml->data), &(dptr));
 
 
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -188,8 +188,13 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
             int szdp = corenrn.get_prop_dparam_size()[type];
             int is_art = corenrn.get_is_artificial()[type];
 
-            // get device pointer for corresponding mechanism data
-            dptr = cnrn_target_deviceptr(tml->ml->data);
+            if (is_art) {
+                // data has been allocated somewhere else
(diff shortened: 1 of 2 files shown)


coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -89,10 +89,17 @@ void* cnrn_target_deviceptr(void* h_ptr) {
 #if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
     return acc_deviceptr(h_ptr);
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
+    void *d_ptr = nullptr;
```
pramodk, line 92: Looks good but note that `cnrn_target_delete` also has `omp_get_mapped_ptr()`.
One question about `cnrn_target_delete()`: 

```
    omp_target_disassociate_ptr(h_ptr, device_id);
    auto* d_ptr = omp_get_mapped_ptr(h_ptr, device_id);
    omp_target_free(d_ptr, device_id);
```

Once we dissociate pointer, I don't think we can use `omp_get_mapped_ptr()` because it's removed from the present table.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -71,26 +71,33 @@ void cnrn_memcpy_to_device(void* d_ptr, void* h_ptr, size_t len) {
...
-    omp_target_disassociate_ptr(h_ptr, device_id);
-    auto* d_ptr = omp_get_mapped_ptr(h_ptr, device_id);
-    omp_target_free(d_ptr, device_id);
+
+    nrn_pragma_omp(target data device(device_id) use_device_ptr(h_ptr))
+    {
+        d_ptr = h_ptr;
+    }
+
+    return d_ptr;
```
olupton, line 86: I think we don't need the `device(device_id)` clause.
Maybe it's simpler to actually define `omp_get_mapped_ptr` in the global namespace in an `#ifdef` block, and only enable that block for specific compilers? (or, I guess, include a CMake check that detects if it's already defined)

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -71,26 +71,33 @@ void cnrn_memcpy_to_device(void* d_ptr, void* h_ptr, size_t len) {
...
+void cnrn_target_delete(void* h_ptr, size_t len) {
 #if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
-    return acc_deviceptr(h_ptr);
+    acc_delete(h_ptr, len);
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
+    (void)len;
     auto device_id = omp_get_default_device();
-    return omp_get_mapped_ptr(h_ptr, device_id);
+    auto* d_ptr = cnrn_target_deviceptr(h_ptr);
+    omp_target_disassociate_ptr(h_ptr, device_id);
```
olupton, line 99: This is not needed with https://github.com/BlueBrain/CoreNeuron/pull/708, right?

olupton (commented):
The implementation of `omp_get_mapped_ptr` looks good, but let's merge https://github.com/BlueBrain/CoreNeuron/pull/708 first and then reorganise this a bit.

alkino:
Hurray

olupton (approved):

PR #708, opened 2021-12-10
OMP: use pragmas instead of c-api to allocate and delete target data

alkino:
This way is more convenient for us.

### DIFF
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -46,7 +46,7 @@ void nrn_VecPlay_delete_from_device(NrnThread* nt);
 template <typename T>
 T* cnrn_target_deviceptr(const T* h_ptr) {
 #if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
-    return acc_deviceptr(static_cast<T*>(h_ptr));
+    return acc_deviceptr(const_cast<T*>(h_ptr));
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
     return static_cast<T*>(omp_get_mapped_ptr(const_cast<T*>(h_ptr), omp_get_default_device()));
 #else
--- a/coreneuron/gpu/nrn_acc_manager.cpp
+++ b/coreneuron/gpu/nrn_acc_manager.cpp
@@ -55,12 +55,12 @@ void* cnrn_target_deviceptr(void* h_ptr) {
 }
 
 template <typename T>
-void* cnrn_gpu_copyin(T* h_ptr, std::size_t len = 1) {
+T* cnrn_gpu_copyin(T* h_ptr, std::size_t len = 1) {
 #if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
-    return acc_copyin(h_ptr, len * sizeof(T));
+    return static_cast<T*>(acc_copyin(h_ptr, len * sizeof(T)));
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
     #pragma omp target enter data map(to:h_ptr[:len])
-    return cnrn_target_deviceptr(h_ptr);
+    return static_cast<T*>(cnrn_target_deviceptr(h_ptr));
 #else
     throw std::runtime_error("cnrn_gpu_copyin() not implemented without OpenACC/OpenMP and gpu build");
 #endif
@@ -107,13 +107,13 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
         NrnThread* nt = threads + i;  // NrnThread on host
 
         if (nt->n_presyn) {
-            PreSyn* d_presyns = (PreSyn*) cnrn_gpu_copyin(nt->presyns, nt->n_presyn);
+            PreSyn* d_presyns = cnrn_gpu_copyin(nt->presyns, nt->n_presyn);
         }
 
         if (nt->n_vecplay) {
             /* copy VecPlayContinuous instances */
             /** just empty containers */
-            void** d_vecplay = (void**) cnrn_gpu_copyin(nt->_vecplay, nt->n_vecplay);
+            void** d_vecplay = cnrn_gpu_copyin(nt->_vecplay, nt->n_vecplay);
             // note: we are using unified memory for NrnThread. Once VecPlay is copied to gpu,
             // we dont want to update nt->vecplay because it will also set gpu pointer of vecplay
             // inside nt on cpu (due to unified memory).
@@ -131,7 +131,7 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
      * find
      * corresponding NrnThread using Point_process in NET_RECEIVE block
      */
-    NrnThread* d_threads = (NrnThread*) cnrn_gpu_copyin(threads, nthreads);
+    NrnThread* d_threads = cnrn_gpu_copyin(threads, nthreads);
 
     if (interleave_info == nullptr) {
         printf("\n Warning: No permutation data? Required for linear algebra!");
@@ -150,7 +150,7 @@ void setup_nrnthreads_on_device(NrnThread* threads, int nthreads) {
         /* -- copy _data to device -- */
 
         /*copy all double data for thread */
-        d__data = (double*) cnrn_gpu_copyin(nt->_data, nt->_ndata);
+        d__data = cnrn_gpu_copyin(nt->_data, nt->_ndata);
 
 
         /* Here is the example of using OpenACC data enter/exit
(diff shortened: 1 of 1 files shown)


coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
 void nrn_VecPlay_copyto_device(NrnThread* nt, void** d_vecplay);
 void nrn_VecPlay_delete_from_device(NrnThread* nt);
 
-void* cnrn_gpu_copyin(void* h_ptr, std::size_t len) {
+void* cnrn_target_deviceptr(void* h_ptr) {
```
olupton, line 46: Maybe it would be more consistent to make a template too?
```c++
template <typename T>
T* cnrn_target_deviceptr(T* h_ptr) {
```
and I also think we should be more consistent with the naming.
In the current draft we seem to use 1-2 of `device`, `gpu` and `target` at random 🙃 
- `cnrn_target_deviceptr`
- `cnrn_gpu_copyin`
- `cnrn_target_delete`
- `cnrn_memcpy_to_device`

I don't mind too much as long as they're consistent. For example, the endings could be
- `deviceptr`
- `copyin`
- `delete`
- `memcpy_to_device`

following OpenACC, and the prefix could be `cnrn_target_` or `coreneuron::target::`?

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
+    return acc_deviceptr(h_ptr);
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
-    auto host_id = omp_get_initial_device();
     auto device_id = omp_get_default_device();
-    auto* d_ptr = omp_target_alloc(len, device_id);
-    nrn_assert(d_ptr != nullptr);
-    nrn_assert(omp_target_memcpy(d_ptr, h_ptr, len, 0, 0, device_id, host_id) == 0);
-    nrn_assert(omp_target_associate_ptr(h_ptr, d_ptr, len, 0, device_id) == 0);
-    return d_ptr;
+    return omp_get_mapped_ptr(h_ptr, device_id);
```
olupton, line 51: No need for the named temporary `device_id`.

IIUC we also need to provide an implementation of `omp_get_mapped_ptr` for some compilers (https://github.com/BlueBrain/CoreNeuron/pull/705)?

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
+    return static_cast<T*>(cnrn_target_deviceptr(h_ptr));
 #else
-    throw std::runtime_error("cnrn_memcpy_to_device() not implemented without OpenACC/OpenMP and gpu build");
+    throw std::runtime_error("cnrn_gpu_copyin() not implemented without OpenACC/OpenMP and gpu build");
 #endif
 }
 
-void cnrn_target_delete(void* h_ptr, size_t len) {
+template <typename T>
+void cnrn_target_delete(T* h_ptr, size_t len = 1) {
```
olupton, line 70:
```suggestion
void cnrn_target_delete(T* h_ptr, std::size_t len = 1) {
```

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
+T* cnrn_gpu_copyin(T* h_ptr, std::size_t len = 1) {
 #if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
-    acc_memcpy_to_device(d_ptr, h_ptr, len);
+    return static_cast<T*>(acc_copyin(h_ptr, len * sizeof(T)));
 #elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENMP)
-    auto host_id = omp_get_initial_device();
-    auto device_id = omp_get_default_device();
-    omp_target_memcpy(d_ptr, h_ptr, len, 0, 0, device_id, host_id);
+    #pragma omp target enter data map(to:h_ptr[:len])
+    return static_cast<T*>(cnrn_target_deviceptr(h_ptr));
```
olupton, line 63: If `cnrn_target_deviceptr` is templated then I think this cast should be in its definition, not here.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
-    auto* d_ptr = omp_get_mapped_ptr(h_ptr, device_id);
-    omp_target_free(d_ptr, device_id);
+    #pragma omp target exit data map(delete: h_ptr[:len])
 #else
     throw std::runtime_error("cnrn_target_delete() not implemented without OpenACC/OpenMP and gpu build");
 #endif
 }
 
-void* cnrn_target_deviceptr(void* h_ptr) {
+void cnrn_memcpy_to_device(void* d_ptr, void* h_ptr, size_t len) {
```
olupton, line 80:
```suggestion
template <typename T>
void cnrn_memcpy_to_device(T* d_ptr, T* h_ptr, std::size_t len) {
```
would provide a little more type-safety, I think.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
+    return omp_get_mapped_ptr(h_ptr, device_id);
 #else
-    throw std::runtime_error("cnrn_gpu_copyin() not implemented without OpenACC/OpenMP and gpu build");
+    throw std::runtime_error("cnrn_target_delete() not implemented without OpenACC/OpenMP and gpu build");
 #endif
 }
 
-void cnrn_memcpy_to_device(void* d_ptr, void* h_ptr, size_t len) {
+template <typename T>
+T* cnrn_gpu_copyin(T* h_ptr, std::size_t len = 1) {
```
olupton, line 58:
```suggestion
T* cnrn_gpu_copyin(T const* h_ptr, std::size_t len = 1) {
```
?

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -522,12 +506,12 @@ void copy_ivoc_vect_to_device(const IvocVect& from, IvocVect& to, bool vector_co
     /// if we need to copy IvocVect vector then newly alloated vector
     /// on the device is a new destination pointer
     if(vector_copy_needed) {
-        d_iv = (IvocVect*) cnrn_gpu_copyin((void*) &from, sizeof(IvocVect));
+        d_iv = cnrn_gpu_copyin(const_cast<IvocVect*>(&from));
```
olupton, line 509: Wondering if the `..copyin` function should take a const pointer and return a non-const pointer, in which case this `const_cast` would not be needed. 🤔

olupton (commented):
Looks nice! I think we should do a little renaming to make things more consistent, but I think it should all be easy search/replace stuff once we decide what to do.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -43,56 +43,49 @@ void nrn_ion_global_map_delete_from_device();
...
     auto device_id = omp_get_default_device();
-    auto* d_ptr = omp_target_alloc(len, device_id);
-    nrn_assert(d_ptr != nullptr);
-    nrn_assert(omp_target_memcpy(d_ptr, h_ptr, len, 0, 0, device_id, host_id) == 0);
-    nrn_assert(omp_target_associate_ptr(h_ptr, d_ptr, len, 0, device_id) == 0);
-    return d_ptr;
+    return omp_get_mapped_ptr(h_ptr, device_id);
 #else
-    throw std::runtime_error("cnrn_gpu_copyin() not implemented without OpenACC/OpenMP and gpu build");
+    throw std::runtime_error("cnrn_target_delete() not implemented without OpenACC/OpenMP and gpu build");
```
iomaganaris, line 53: This should have been like the following?
```suggestion
    throw std::runtime_error("cnrn_target_deviceptr() not implemented without OpenACC/OpenMP and gpu build");
```

coreneuron/gpu/nrn_acc_manager.cpp
iomaganaris: `cnrn_target` prefix looks good to me as well. Also the same for the rest of the functions

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -521,13 +498,13 @@ void copy_ivoc_vect_to_device(const IvocVect& from, IvocVect& to, bool vector_co
     /// if we need to copy IvocVect vector then newly alloated vector
     /// on the device is a new destination pointer
     if(vector_copy_needed) {
-        d_iv = (IvocVect*) cnrn_gpu_copyin((void*) &from, sizeof(IvocVect));
-        cnrn_memcpy_to_device(&to, &d_iv, sizeof(IvocVect*));
+        d_iv = cnrn_target_copyin(&from);
+        cnrn_target_memcpy_to_device(&to, d_iv);
```
olupton, line 502: I am still a bit confused by this part. I will investigate more.

olupton (approved):
🚀

PR #711, opened 2021-12-11
Remove dead code in gpu

alkino:
Remove code than seems dead


coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -532,66 +532,6 @@ void delete_ivoc_vect_from_device(IvocVect& vec) {
 #endif
 }
 
-void realloc_net_receive_buffer(NrnThread* nt, Memb_list* ml) {
```
olupton, line 535: Not dead, can be called from NMODL/MOD2C:
https://github.com/BlueBrain/nmodl/blob/4a67187535eb6d79d7ef74095dacd6434948a43d/src/codegen/codegen_c_visitor.cpp#L4104

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -686,44 +626,6 @@ void update_net_receive_buffer(NrnThread* nt) {
     nrn_pragma_omp(taskwait)
 }
 
-void update_net_send_buffer_on_host(NrnThread* nt, NetSendBuffer_t* nsb) {
```
olupton, line 689: Not dead: https://github.com/BlueBrain/nmodl/blob/4a67187535eb6d79d7ef74095dacd6434948a43d/src/codegen/codegen_acc_visitor.cpp#L322

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -872,136 +774,6 @@ void update_nrnthreads_on_host(NrnThread* threads, int nthreads) {
 #endif
 }
 
-void update_nrnthreads_on_device(NrnThread* threads, int nthreads) {
```
olupton, line 875: I think this one is really not used at the moment. I am OK with removing it, we can recover it from git history if needed...if it is not used/tested, chanced are this implementation has bugs anyway.

olupton (changes requested):
It's good to cleanup, but 2/3 of these methods are called from code generated by NMODL (and probably MOD2C).

olupton (approved):

PR #717, opened 2021-12-14
Re-enable CI tests on hackathon branch

olupton:
Re-enable various bits of CI and don't try and enable OpenMP target offload with MOD2C.

With this change we have one new GitLab CI build, giving a total of:
- CPU-only with the Intel compiler and MOD2C
- CPU-only with the Intel compiler and NMODL, no SymPy
- GPU-enabled using OpenACC, using MOD2C and OpenMP host parallelism
- GPU-enabled using OpenACC, using NMODL and SymPy, no OpenMP host parallelism
- GPU-enabled using OpenMP target offload, using NMODL, no SymPy (*), and OpenMP host parallelism

(*) SymPy doesn't work because of OpenMP/Eigen issues: https://forums.developer.nvidia.com/t/enabling-openmp-offload-breaks-openacc-code/196643

The other main change concerning OpenMP support with the NVIDIA compilers is that we no longer enable OpenACC. During the hackathon, when we had not migrated the data transfer code to use OpenMP, we were enabling both OpenACC and OpenMP (`-acc -mp=gpu`) and relying on the compilers' interoperability. This PR drops the `-acc` in that case. Making this work required numerous small fixes, with a lot of overlap with the draft changes for LLVM and XLC offload support.

- [x] Needs https://github.com/BlueBrain/nmodl/pull/784.

CI_BRANCHES:NEURON_BRANCH=master,NMODL_BRANCH=hackathon_main,


olupton:
The CI failures seem to be somehow related to differences between the phase 1 and phase 2 GPU nodes on the cluster. I can only reproduce the failure locally on a phase 2 node, and only with 2 MPI ranks.

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -41,9 +41,44 @@ void nrn_ion_global_map_delete_from_device();
 void nrn_VecPlay_copyto_device(NrnThread* nt, void** d_vecplay);
 void nrn_VecPlay_delete_from_device(NrnThread* nt);
 
+int cnrn_target_get_num_devices() {
```
alkino, line 44: should it be inside offload.hpp?

alkino (approved):

coreneuron/gpu/nrn_acc_manager.cpp
olupton: I think it's OK here. I think we only need to have things in `offload.hpp` if they are used from NMODL.

coreneuron/mechanism/mech/mod2c_core_thread.hpp
```diff
@@ -89,15 +92,17 @@ struct SparseObj {          /* all the state information */
     int do_flag;
 };
 
-#pragma acc routine seq
+nrn_pragma_acc(routine seq)
```
iomaganaris, line 95: `nrn_pragma_omp(declare target)` not needed here?

coreneuron/permute/cellorder.cpp
```diff
@@ -596,7 +596,7 @@ void solve_interleaved2(int ith) {
         int* strides = ii.stride;           // sum ncycles of these (bad since ncompart/warpsize)
         int* rootbegin = ii.firstnode;      // nwarp+1 of these
         int* nodebegin = ii.lastnode;       // nwarp+1 of these
-#if defined(_OPENACC) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD)
+#if defined(_OPENACC) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
```
iomaganaris, line 599:
```suggestion
#if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && defined(_OPENACC)
```
?

coreneuron/sim/scopmath/ssimplic_thread.cpp
```diff
@@ -9,11 +9,12 @@
 #include "coreneuron/mechanism/mech/cfile/scoplib.h"
 #include "coreneuron/mechanism/mech/mod2c_core_thread.hpp"
 #include "coreneuron/sim/scopmath/errcodes.h"
+#include "coreneuron/utils/offload.hpp"
 
 namespace coreneuron {
 #define s_(arg) _p[s[arg] * _STRIDE]
 
-#pragma acc routine seq
+nrn_pragma_acc(routine seq)
```
iomaganaris, line 17: `nrn_pragma_omp(declare target)` not needed here?

iomaganaris (commented):
LGTM in general. Some small questions here and there.
Another question I have whether the OpenMP + Unified memory has been tested either in the CI or locally

coreneuron/mechanism/mech/mod2c_core_thread.hpp
olupton: I tried to follow the approach of only adding it when its absence caused problems. I guess this isn't actually called in the test suite 🤷

coreneuron/sim/scopmath/ssimplic_thread.cpp
olupton: Same as above I think.

olupton:
> Another question I have whether the OpenMP + Unified memory has been tested either in the CI or locally
I think that the Jenkins CI running on https://github.com/BlueBrain/CoreNeuron/pull/713 will be trying to do something there, but I have not checked it carefully. I never tried this myself locally.

coreneuron/mechanism/mech/mod2c_core_thread.hpp
olupton: I added it anyway.

coreneuron/sim/scopmath/ssimplic_thread.cpp
olupton: I added it anyway.

iomaganaris:
> > Another question I have whether the OpenMP + Unified memory has been tested either in the CI or locally
> 
> I think that the Jenkins CI running on #713 will be trying to do something there, but I have not checked it carefully. I never tried this myself locally.

I see. Something to check before merging to `master` then

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -41,9 +41,44 @@ void nrn_ion_global_map_delete_from_device();
...
+#if defined(CORENEURON_ENABLE_GPU) && !defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && \
+    defined(_OPENACC)
+    acc_set_device_num(device_num, acc_device_nvidia);
+#elif defined(CORENEURON_ENABLE_GPU) && defined(CORENEURON_PREFER_OPENMP_OFFLOAD) && \
+    defined(_OPENMP)
+    omp_set_default_device(device_num);
+    // It seems that with NVHPC 21.9 then only setting the default OpenMP device
+    // is not enough: there were errors on some nodes when not-the-0th GPU was
+    // used. These seemed to be related to the NMODL instance structs, which are
+    // allocated using cudaMallocManaged.
```
pramodk, line 70: not sure if this is still an issue but curious if this is still an issue.

pramodk (approved):
LGTM!
(Skimmed through changes except CI part and LGTM. When the PR will be created against master then we can discuss overall changes.)

coreneuron/gpu/nrn_acc_manager.cpp
olupton: Still an issue in 21.11 you mean?

PR #718, opened 2021-12-16
GPU implementation improvements

iomaganaris:
Set nwarp to very big number for optimal parallelization and improve a bit grid config of CUDA solve_interleaved2

**Description**

- [x] After profiling of various configurations of the channel-benchmark we figured out that the optimal number of nwarp is the number of cells to achieve the best parallelization and more coalesced memory access in the solve_interleaved2 kernel and the current updates
- [x] Improved (?) the way the CUDA implementation of solve_interleave2 is launched

### DIFF
--- a/coreneuron/permute/cellorder.cu
+++ b/coreneuron/permute/cellorder.cu
@@ -94,7 +94,7 @@ void solve_interleaved2_launcher(NrnThread* nt, InterleaveInfo* info, int ncore,
 
     /// the selection of these parameters has been done after running the channel-benchmark for
     /// typical production runs, i.e. 1 MPI task with 1440 cells & 6 MPI tasks with 8800 cells.
-    /// In the OpenACC/OpenMP implementations threadsPerBlock is set to 1. From profiling the
+    /// In the OpenACC/OpenMP implementations threadsPerBlock is set to 32. From profiling the
     /// channel-benchmark circuits mentioned above we figured out that the best performance was
     /// achieved with this configuration
     int threadsPerBlock = warpsize;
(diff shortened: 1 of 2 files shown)


coreneuron/permute/cellorder.cu
```diff
@@ -92,12 +92,17 @@ __global__ void solve_interleaved2_kernel(NrnThread* nt, InterleaveInfo* ii, int
...
 
-    // the selection of these parameters has been done after running the channel-benchmark for typical production runs, i.e.
-    // 1 MPI task with 1440 cells & 6 MPI tasks with 8800 cells.
-    // The main idea is to have multiple warps per SM and sufficient blocks to fill the GPU. 
-    // In our case, given that multiple threads share the available GPUs, we "guarantee" a sufficient occupancy of the GPUs.
-    int threadsPerBlock = 128;
-    int blocksPerGrid = 512;
+    /// the selection of these parameters has been done after running the channel-benchmark for
+    /// typical production runs, i.e. 1 MPI task with 1440 cells & 6 MPI tasks with 8800 cells.
+    /// In the OpenACC/OpenMP implementations threadsPerBlock is set to 1. From profiling the
```
kotsaloscv, line 97: I do not think that in OpenACC/OpenMP implementations threadsPerBlock is set to 1. There should be at least 32 threads per block in order to utilize the SM. Are you sure about this comment?

kotsaloscv (approved):

coreneuron/permute/cellorder.cu
iomaganaris: You're right. This is 32 (warpsize) and not 1 :+1:

PR #723, opened 2021-12-21
Request 16 cores for NEURON test jobs

olupton:
**Description**
The NEURON test suite is quite large now, and https://github.com/neuronsimulator/nrn/pull/1556 is going to add a larger single test that requests 16 ranks, so it seems reasonable to increase the number of allocated cores. The old default was 8.


pramodk (approved):
LGTM

alexsavulescu (approved):

PR #730, opened 2021-12-23
Add a function to delete gap indices from device

### LINKED ISSUE #727: CoreNEURON does not free GPU buffers allocated for gap junctions
**Describe the issue**
CoreNEURON does not free all GPU memory that it allocates for gap junction handling:
https://github.com/BlueBrain/CoreNeuron/blob/53b0c5fda91e98c24000238285f7ef1330fc06a4/coreneuron/network/partrans.cpp#L133-L174

This could cause problems if we repeatedly run CoreNEURON in a single process, so we should add cleanup code.

**Expected behavior**
CoreNEURON should clean up after itself.

alkino:
Fix #727

### DIFF
--- a/coreneuron/network/partrans.cpp
+++ b/coreneuron/network/partrans.cpp
@@ -144,23 +144,19 @@ void nrn_partrans::copy_gap_indices_to_device() {
             continue;
         }
 
-        TransferThreadData& ttd = transfer_thread_data_[tid];
+        const TransferThreadData& ttd = transfer_thread_data_[tid];
 
         if (!ttd.src_indices.empty()) {
-            size_t n_src_indices = ttd.src_indices.size();
-            int* src_indices = ttd.src_indices.data();
-            cnrn_target_copyin(src_indices, n_src_indices);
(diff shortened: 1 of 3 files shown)


coreneuron/network/partrans.cpp
```diff
@@ -129,45 +129,69 @@ void nrnthread_v_transfer(NrnThread* _nt) {
...
+
+            size_t n_src_gather = ttd.src_gather.size();
+            double* src_gather = ttd.src_gather.data();
+            cnrn_target_delete(src_gather, n_src_gather);
+        }
+
+        if (!ttd.insrc_indices.empty()) {
+            int* insrc_indices = ttd.insrc_indices.data();
+            size_t n_insrc_indices = ttd.insrc_indices.size();
+            cnrn_target_delete(insrc_indices, n_insrc_indices);
```
olupton, line 194: I would have written
```suggestion
            cnrn_target_delete(ttd.insrc_indices.data(), ttd.insrc_indices.size());
```
and similar elsewhere, but I don't have a strong opinion.

olupton (approved):
LGTM, 🤞 the CI agrees.

PR #731, opened 2022-01-04
GitLab CI improvements

olupton:
**Description**
Streamline the pipeline a little by using a single build of NMODL with GCC instead of building it with both Intel and NVHPC. 

Also allocate more resources for build jobs to avoid too many of them landing on the same nodes and killing the filesystem.

Extend an existing fix for NVHPC to avoid building `py-{cython,numpy}` with Intel, which seems fragile.

~Start the CUDA MPS daemon before running GPU tests. Unfortunately this does not work well if we do not allocate the GPU node exclusively, as only a single instance can run per node, so we also allocate the node exclusively..which is a bit resource-heavy.~

Hopefully make the job names easier to parse.

Bump the mod2c submodule to include https://github.com/BlueBrain/mod2c/pull/76 to improve `ccache` performance in GitLab CI.

This is being tested against the NEURON branch of https://github.com/neuronsimulator/nrn/pull/1574, but it should also be compatible with the `master` of NEURON.

TODO:
 - [x] Re-disable ccache debugging.
 - [x] Revert to using the default branch of `gitlab-pipelines`.
 - [x] Use relative paths in mod2c/neuron transpilers.
 - [x] Decide whether to revert the MPS + exclusive part

CI_BRANCHES:NEURON_BRANCH=olupton/relative-paths,

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,16 +54,8 @@ spack_setup:
     - SPACK_PACKAGE_DEPENDENCIES="${SPACK_PACKAGE_DEPENDENCIES}^py-cython%gcc^py-numpy%gcc"
     - !reference [.spack_build, before_script]
 .gpu_node:
-  before_script:
-    - !reference [.ctest, before_script]
-    - echo "Starting MPS daemon for GPU tests."
-    - nvidia-cuda-mps-control -d || true
-    - ps aux | grep nvidia-cuda-mps
   variables:
     bb5_constraint: volta
-    bb5_exclusive: full
-    bb5_ntasks: 16
-    bb5_memory: 0
 .test_neuron:
   extends: [.ctest]
   variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,7 +61,7 @@ variables:
     - !reference [.ctest, before_script]
     - echo "Starting MPS daemon for GPU tests."
     - nvidia-cuda-mps-control -d || true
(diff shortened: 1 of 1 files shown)


olupton:
Retest this please Jenkins.

.gitlab-ci.yml
```diff
@@ -35,187 +27,158 @@ spack_setup:
...
+  before_script:
+    # Build py-cython and py-numpy with GCC instead of Intel/NVHPC.
+    - SPACK_PACKAGE_DEPENDENCIES="${SPACK_PACKAGE_DEPENDENCIES}^py-cython%gcc^py-numpy%gcc"
+    - !reference [.spack_build, before_script]
 .gpu_node:
+  before_script:
+    - !reference [.ctest, before_script]
+    - echo "Starting MPS daemon for GPU tests."
+    - nvidia-cuda-mps-control -d || true
+    - ps aux | grep nvidia-cuda-mps
```
pramodk: one of the reason for this not working is that daemon needs to be started in same ssh/process session.

pramodk (approved):
LGTM

.gitlab-ci.yml
olupton: I believe that this command was run in the same shell that launched `ctest` 🤔

olupton:
Retest this please Jenkins.

alexsavulescu (approved):

iomaganaris (approved):
LGTM!

PR #744, opened 2022-01-13
Fetch submodules eagerly so git status/log work.

olupton:


olupton:
This fix will be included in https://github.com/BlueBrain/CoreNeuron/pull/745 instead.

PR #750, opened 2022-01-14
Use new libsonatareport

alkino:
Libsonatareport change API to let us set filename instead of hardcoded "out", follow the change.

CI_BRANCHES:SPACK_BRANCH=cornu/libsonatareport_for_coreneuron_pr,

### DIFF
--- a/coreneuron/io/output_spikes.cpp
+++ b/coreneuron/io/output_spikes.cpp
@@ -64,10 +64,10 @@ void spikevec_unlock() {
     mut.unlock();
 }
 
-void local_spikevec_sort(std::vector<double>& isvect,
-                         std::vector<int>& isvecg,
-                         std::vector<double>& osvect,
-                         std::vector<int>& osvecg) {
+static void local_spikevec_sort(std::vector<double>& isvect,
+                                std::vector<int>& isvecg,
+                                std::vector<double>& osvect,
+                                std::vector<int>& osvecg) {
     osvect.resize(isvect.size());
     osvecg.resize(isvecg.size());
     // first build a permutation vector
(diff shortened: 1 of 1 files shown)


coreneuron/io/output_spikes.cpp
```diff
@@ -64,10 +64,10 @@ void spikevec_unlock() {
     mut.unlock();
```
pramodk, line 64: is above submodule change intentional?

coreneuron/io/output_spikes.cpp
```diff
@@ -288,11 +289,11 @@ void output_spikes(const char* outpath,
...
 #if NRNMPI
     if (corenrn_param.mpi_enable && nrnmpi_initialized()) {
-        output_spikes_parallel(outpath, population_name_offset);
+        output_spikes_parallel(outpath, "out", population_name_offset);
     } else
 #endif
     {
-        output_spikes_serial(outpath);
+        output_spikes_serial(outpath, "out");
     }
```
pramodk, line 297: So ascii file will be always out.dat. Its fine just want to confirm with @jorblancoa.

jorblancoa: correct

coreneuron/io/output_spikes.cpp
```diff
@@ -253,9 +254,9 @@ void output_spikes_parallel(
 }
 #endif
 
-void output_spikes_serial(const char* outpath) {
+static void output_spikes_serial(const char* outpath, const char* filename) {
     std::stringstream ss;
-    ss << outpath << "/out.dat";
+    ss << outpath << "/" << filename << ".dat";
```
jorblancoa, line 259: as @pramodk mentioned, the .dat file should always be "out.dat"

coreneuron/io/output_spikes.cpp
alkino: yes, but it brokes everything :p

PR #753, opened 2022-01-14
Add missing dependency between mod2c_core and nrnivmodl-core

### LINKED ISSUE #749: Mis-declared build dependency in CoreNEURON
**Describe the issue**
https://github.com/neuronsimulator/nrn-build-ci/runs/4812063721?check_suite_focus=true#step:6:791 seems to indicate that the dependency on the `mod2c` binary is not correctly declared. This doesn't seem to be a common problem, but it would be nice to fix it if we can.

pramodk:
fixes #749


olupton (approved):
LGTM, thanks!

PR #789, opened 2022-03-16
Default/Minimal coreneuron should be always available in NEURON (cpu)

pramodk:
**Description**

* Before this PR, user always had to do `nrnivmodl -coreneuron .``
  to have coreneuron with default mechanisms.
* This was because we were not installing default mechanism library
  libcorenrnmech.so into install dir. Note that this was done in order
  to force user to re-build optimised library from vendor compiler
  like Intel.
* But, for basic demonstration and testing, I think it's always
  helpful to ship default mechanism library as a fallback. The
  library <arch>/libcorenrnmech.so built by user will be preferred anyway.
* Update nmodl to latest master

**How to test this?**

If you install neuron+coreneuron as:

```console
cmake .. -DCMAKE_INSTALL_PREFIX=`pwd`/install -DNRN_ENABLE_INTERVIEWS=OFF -DNRN_ENABLE_RX3D=OFF  -DNRN_ENABLE_CORENEURON=ON -DNRN_ENABLE_TESTS=ON -DNRN_CMAKE_FORMAT=ON
make -j
make install
```

Now we can run basic coreneuron test without `nrnivmodl -coreneuron .`:

```console
export PYTHONPATH=~/some-dir/install/lib/python:$PYTHONPATH
export PATH=~/some-dir/install/bin:$PATH
python3 ../test/coreneuron/test_psolve.py
```

**Test System**
 - OS: MacOS
 - Compiler: Clang

Hope this works elsewhere!

CI_BRANCHES:NEURON_BRANCH=pramodk/corenrn-mechlib-install,

### DIFF
--- a/coreneuron/CMakeLists.txt
+++ b/coreneuron/CMakeLists.txt
@@ -326,13 +326,25 @@ set_target_properties(
 # =============================================================================
 set(modfile_directory "${CORENEURON_PROJECT_SOURCE_DIR}/tests/integration/ring_gap/mod")
 file(GLOB modfiles "${modfile_directory}/*.mod")
+
+if(CORENRN_ENABLE_SHARED)
+  set(corenrn_mech_library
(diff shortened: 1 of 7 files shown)


iomaganaris (approved):
LGTM

PR #794, opened 2022-03-31
Update hpc-coding-convention

alkino:


CMakeLists.txt
```diff
@@ -59,10 +59,10 @@ set(CORENRN_ClangFormat_FILES_RE
     "^.*\\\\.cu$$" "^.*\\\\.[chi]pp$$"
     CACHE STRING "List of regular expressions matching C/C++ filenames" FORCE)
 set(CORENRN_ClangFormat_EXCLUDES_RE
-    "${CORENRN_PROJECT_SOURCE_DIR}/external/.*$$"
+    "external/.*$$"
```
olupton, line 62:
```suggestion
    ""
```

CMakeLists.txt
```diff
@@ -59,10 +59,10 @@ set(CORENRN_ClangFormat_FILES_RE
...
     CACHE STRING "list of regular expressions to exclude C/C++ files from formatting" FORCE)
 set(CORENRN_CMakeFormat_EXCLUDES_RE
-    "${CORENRN_PROJECT_SOURCE_DIR}/external/.*$$" ".*/CMake/packages/.*$$"
+    "external/.*$$" "CMake/packages/.*$$"
```
olupton, line 65:
```suggestion
    "CMake/packages/.*$$"
```

olupton (commented):
I don't think we need to exclude submodule directories any more, do we?

CMakeLists.txt
tristan0x: that's right, paths are relative to ${CORENRN_PROJECT_SOURCE_DIR} now

olupton (approved):

PR #796, opened 2022-04-07
Add *NRN_VERSION_* macros and variables.

olupton:
**Description**

This is the CoreNEURON counterpart of  https://github.com/neuronsimulator/nrn/pull/1762, which addresses https://github.com/neuronsimulator/nrn/issues/1696.

- Generate a new header and define `NRN_VERSION*` macros that can be used in `VERBATIM` blocks. Include this from `coreneuron/nrnconf.h` so that these macros are implicitly available in all translated mechanisms. Tests are based on the NEURON version, distinguishing between different CoreNEURON versions within the same NEURON version is not foreseen.
- These resolve to the CoreNEURON `PROJECT_VERSION_{MAJOR,MINOR,PATCH}` when CoreNEURON is built standalone. This is now supposed to be kept in sync with the NEURON version.
- Checks are added at the CMake level on the NEURON side to make sure these values do not drift out of sync

**How to test this?**

https://github.com/neuronsimulator/nrn/pull/1762 adds a test mechanism and script that is run both in NEURON and CoreNEURON modes.

**Test System**
 - OS: macOS
 - Compiler: AppleClang
 - Backend: CPU

CI_BRANCHES:NEURON_BRANCH=olupton/version-macros


CMakeLists.txt
```diff
@@ -33,14 +33,51 @@ if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES OR NOT "cxx_digit_separat
       "This compiler does not fully support C++14, choose an higher version or an other compiler.")
 endif()
 
+# If CoreNEURON is built standalone then it has to know what version of NEURON to claim
+# compatibility with via NRN_VERSION_* macros in VERBATIM blocks.
+set(CORENRN_COMPATIBLE_NRN_VERSION_MAJOR 8)
+set(CORENRN_COMPATIBLE_NRN_VERSION_MINOR 2)
+set(CORENRN_COMPATIBLE_NRN_VERSION_PATCH 0)
```
pramodk, line 40: Just a note: I wondered if coreneuron could follow the same version numbers as neuron for simplicity. (but we want to get rid of coreneuron as standalone builds anyway...)

CMakeLists.txt
```diff
@@ -33,14 +33,51 @@ if(NOT "cxx_std_14" IN_LIST CMAKE_CXX_COMPILE_FEATURES OR NOT "cxx_digit_separat
...
+  message(${level} "CoreNEURON could not determine the NEURON version, using the hardcoded "
+          "${NRN_VERSION_MAJOR}.${NRN_VERSION_MINOR}.${NRN_VERSION_PATCH}")
+endif()
+# Regardless of whether we are being built as a submodule of NEURON, NRN_VERSION_{MAJOR,MINOR,PATCH}
+# are now set to the version that we should claim compatibility with when compiling translated MOD
+# files. Generate a header under a special `generated` prefix in the build directory, so that
+# -I/path/to/src -I/path/to/build/generated is safe (headers from the source prefix are copied
+# elsewhere under the build prefix, so there is scope for confusion)
+configure_file(coreneuron/config/neuron_version.hpp.in
+               generated/coreneuron/config/neuron_version.hpp)
```
pramodk, line 80: could you explain a bit why we first generate it under `generated` and then do copy again under `/include` ?

pramodk (approved):
LGTM. I added some comments for clarification.

CMakeLists.txt
olupton: I think that would be fine, it just seemed a [little] bit more invasive. If you think it's worth it then I'm happy to change?

pramodk: if you also think it's cleaner and simple, I am happy to make the change!

alexsavulescu (approved):
LGTM also

CMakeLists.txt
olupton: This is indirectly because of the `nrnivmodl` workflow. Inside `nrnivmodl` then the include path is always `{prefix}/include`, which is either inside the build or install directory, so you need to make sure all headers are there.

For the actual (CMake-driven) build of CoreNEURON, the include paths point to the source directory and (now) this new `generated/` directory, but not the `include/` directory. It would be dangerous to add `{build}/include` to the search path, as then you would have two copies of the same headers visible. I don't remember the exact mechanism, but I'm pretty sure this caused problems with `#pragma once`.

pramodk: ok, thanks for explanation!

CMakeLists.txt
pramodk: 👍, good to go!

olupton:
New pipeline is https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/47911, the old one failed because I rebased the NEURON PR at a bad time. Hopefully once it succeeds the status will go green here 🤷

PR #797, opened 2022-04-12
Random123 + OpenACC: do not compile as CUDA.

olupton:
**Description**
Compile nrnran123.cpp as C++ with OpenACC annotations in GPU builds using OpenACC. This is aligning OpenACC with what we already did for OpenMP target offload.

Drop `cuda_profile.{cu,h}` that did not need to be CUDA code in the first place.


pramodk (approved):
LGTM

PR #802, opened 2022-04-22
Check variable return by get_target_property before it's usage

pramodk:
**Description**

* CMake's `get_target_property` might return `<VAR>-NOTFOUND` and
  hence it should be checked before it's usage
* As mentioned in neuronsimulator/nrn/issues/1788, it can result
  into error like:
```
    => LINKING executable ./special LDFLAGS are:     -lreadline -lncurses
       /users/bp000174/nrn-install/share/nrn/nrnmain.cpp:
       /usr/bin/ld: cannot find -lCORENRN_LINK_LIBRARIES-NOTFOUND
```

**Test System**

- Piz-Daint

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -473,23 +473,25 @@ add_subdirectory(coreneuron)
 
 if(CORENRN_ENABLE_GPU)
   get_target_property(CORENRN_LINK_LIBRARIES coreneuron INTERFACE_LINK_LIBRARIES)
-  foreach(LIB ${CORENRN_LINK_LIBRARIES})
-    get_filename_component(dir_path ${LIB} DIRECTORY)
-    if(TARGET ${LIB})
-      # See, for example, caliper where the coreneuron target depends on the caliper target (so we
-      # get LIB=caliper in this loop), but -l and -L are already added manually here:
-      # https://github.com/BlueBrain/CoreNeuron/blob/856cea4aa647c8f2b0d5bda6d0fc32144c5942e3/CMakeLists.txt#L411-L412
-      message(
-        NOTICE
-        "Ignoring dependency '${LIB}' of 'coreneuron' and assuming relevant flags have already been added to CORENEURON_LIB_LINK_FLAGS."
-      )
-    elseif(NOT dir_path)
-      # In case LIB is not a target but is just the name of a library, e.g. "dl"
-      set_property(GLOBAL APPEND_STRING PROPERTY CORENEURON_LIB_LINK_FLAGS " -l${LIB}")
-    else()
-      set_property(GLOBAL APPEND_STRING PROPERTY CORENEURON_LIB_LINK_FLAGS " ${LIB}")
-    endif()
-  endforeach()
+  if(CORENRN_LINK_LIBRARIES)
+    foreach(LIB ${CORENRN_LINK_LIBRARIES})
+      get_filename_component(dir_path ${LIB} DIRECTORY)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -473,23 +473,25 @@ add_subdirectory(coreneuron)
...
-        "Ignoring dependency '${LIB}' of 'coreneuron' and assuming relevant flags have already been added to CORENEURON_LIB_LINK_FLAGS."
-      )
-    elseif(NOT dir_path)
-      # In case LIB is not a target but is just the name of a library, e.g. "dl"
-      set_property(GLOBAL APPEND_STRING PROPERTY CORENEURON_LIB_LINK_FLAGS " -l${LIB}")
-    else()
-      set_property(GLOBAL APPEND_STRING PROPERTY CORENEURON_LIB_LINK_FLAGS " ${LIB}")
-    endif()
-  endforeach()
+  if(CORENRN_LINK_LIBRARIES)
```
olupton, line 476: Doesn't this need to be something like
```cmake
if(NOT CORENRN_LINK_LIBRARIES STREQUAL "CORENRN_LINK_LIBRARIES-NOTFOUND")
```
? https://cmake.org/cmake/help/latest/command/get_target_property.html

pramodk: I tested earlier and I believe it's not required. e.g. https://github.com/neuronsimulator/nrn/blob/548348a2de18c9ea7d5585d9196338e25bd6d37e/cmake/CMakeListsNrnMech.cmake#L16

Let me double check...

pramodk: So with:

```cmake
  get_target_property(CORENRN_LINK_LIBRARIES coreneuron INTERFACE_LINK_LIBRARIES)
  message(INFO "CORENRN_LINK_LIBRARIES: ${CORENRN_LINK_LIBRARIES}")
  if(CORENRN_LINK_LIBRARIES)
    message(STATUS "if(CORENRN_LINK_LIBRARIES) is TRUE")
  endif()
  if(NOT CORENRN_LINK_LIBRARIES)
    message(STATUS "if(NOT CORENRN_LINK_LIBRARIES) is TRUE")
  endif()
```

when libraries are not found:

```console
CORENRN_LINK_LIBRARIES: CORENRN_LINK_LIBRARIES-NOTFOUND
if(NOT CORENRN_LINK_LIBRARIES) is TRUE
```

and when libraries are found:

```console
CORENRN_LINK_LIBRARIES: /opt/cray/pe/atp/3.14.5/lib/libAtpSigHandler.so;/opt/cray/rca/2.2.20-7.0.3.1_2.20__g8e3fb5b.ari/lib64/librca.so;/opt/cray/pe/mpt/7.7.18/gni/mpich-nvidia/20.7/lib/libmpich_nvidia.so;/opt/cray/pe/mpt/7.7.18/gni/mpich-nvidia/20.7/lib/libmpichcxx_nvidia.so
if(CORENRN_LINK_LIBRARIES) is TRUE
```

And seems similar pattern is used elsewhere:
https://github.com/playmer/YTE/blob/c26320a84efe12f4f6555b8f36eb7e99dc3b3b7c/cmake/YTE_Target_Folder.cmake#L59

But, I had same thought as you when I read the doc but when I saw usage in NRN, I thought it "should be" safe. Let me know if you think otherwise.

olupton: OK, thanks for checking!

olupton (approved):

PR #809, opened 2022-05-04
Make legacy scopmath solvers header only, drop kinderiv.py and dimplic.cpp

olupton:
**Description**
Move scopmath solvers into header files, drop kinderiv.py/dimplic.cpp workarounds.

TODO:
- [x] Update submodule once https://github.com/BlueBrain/mod2c/pull/81 is merged.
- [x] Update submodule once https://github.com/BlueBrain/nmodl/pull/859 is merged.


pramodk:
It seems like the simulation stack is failing with unrelated errors and could be ignored.

image

coreneuron/CMakeLists.txt
```diff
@@ -146,17 +126,6 @@ set_property(
...
-# run KINDERIV_PYTHON_SCRIPT to generate _kinderiv.h
-# =============================================================================
-add_custom_command(
-  OUTPUT "${KINDERIV_HEADER_FILE}"
-  DEPENDS ${NMODL_INBUILT_MOD_OUTPUTS} "${KINDERIV_PYTHON_SCRIPT}"
-  COMMAND ${PYTHON_EXECUTABLE} ${KINDERIV_PYTHON_SCRIPT} ${CMAKE_CURRENT_BINARY_DIR}
-  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-  COMMENT "Generating kinderiv.h by inspecting MOD files")
-add_custom_target(kin_deriv_header DEPENDS "${KINDERIV_HEADER_FILE}")
-
```
pramodk, line 159: this really deserves :bow: !

coreneuron/apps/corenrn_parameters.cpp
```diff
@@ -184,13 +184,21 @@ void corenrn_parameters::parse(int argc, char** argv) {
...
     } catch (const CLI::ParseError& e) {
         // use --help is also ParseError; in this case exit by showing all options
         app.exit(e);
         exit(0);
     }
 
+#ifndef CORENEURON_ENABLE_GPU
+    if (gpu) {
+        std::cerr << "GPU support was not enabled at build time but --gpu was passed. This is "
+                     "probably not what you meant."
```
pramodk, line 196: this could be simplified a bit:
```suggestion
        std::cerr << "Error: GPU support was not enabled at build time but GPU execution is requested."
```

coreneuron/io/phase2.cpp
```diff
@@ -483,7 +483,7 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
             nrb->_pnt_offset = pnt_offset[type];
 
             // begin with a size of 5% of the number of instances
-            nrb->_size = ml->nodecount;
+            nrb->_size = 0.05 * ml->nodecount;
```
pramodk, line 486: I am not sure if the comment was outdated i.e. it should be `ml->nodecount` ? May be @nrnhines can confirm.

coreneuron/sim/scopmath/crout_thread.hpp
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * Copyright (c) 1987, 1988, 1989, 1990
+ *   Duke University
+ *
+ ******************************************************************************/
+#pragma once
+#include "coreneuron/sim/scopmath/errcodes.h"
+#include "coreneuron/sim/scopmath/newton_struct.h"
+
+// #include <cmath>
```
pramodk, line 13: `nitpick`: remove?

By the way, were there any logical changes in this file? (asking as Github doesn't show this as mere renaming)

coreneuron/sim/scopmath/newton_thread.hpp
```diff
@@ -0,0 +1,183 @@
+/*
+# =============================================================================
+# Originally newton.c from SCoP library, Copyright (c) 1987-90 Duke University
+# =============================================================================
```
pramodk, line 4: same question for this file: just copy from original file I assume?

coreneuron/sim/scopmath/ssimplic_thread.hpp
```diff
@@ -0,0 +1,74 @@
...
+    for (int i = 0; i < n; i++) {
+        if (scopmath_ssimplic_s(i) < -1e-6) {
+            scopmath_ssimplic_s(i) = 0.;
+            flag = false;
+        }
+    }
+    return flag;
+}
+#undef scopmath_ssimplic_s
+
```
pramodk, line 28: same question

coreneuron/gpu/nrn_acc_manager.cpp
```diff
@@ -626,44 +626,50 @@ void realloc_net_receive_buffer(NrnThread* nt, Memb_list* ml) {
...
+        using T = std::remove_pointer_t<std::remove_reference_t<decltype(ptr)>>;
+        static_assert(std::is_trivial<T>::value,
+                      "Only trivially constructible and copiable types are supported.");
+        static_assert(std::is_same<decltype(ptr), T*&>::value,
+                      "ptr should be reference-to-pointer");
+        auto* const new_data = static_cast<T*>(ecalloc_align((nrb->_size + extra_size), sizeof(T)));
+        std::memcpy(new_data, ptr, (old_size + extra_size) * sizeof(T));
+        free_memory(ptr);
+        ptr = new_data;
+    };
```
pramodk, line 641: could remind me if this is done for consistency and to use unified memory in case unified-memory-build?

pramodk (commented):
Quickly skimmed through this and overall everything looks fine!

coreneuron/gpu/nrn_acc_manager.cpp
olupton: It's for consistency. The unified memory builds are a more extreme example, where this would have called `realloc` on a unified memory pointer, but even in standard builds I think this was pairing `realloc` with `posix_memalign`, which seems dubious (https://linux.die.net/man/3/realloc).

coreneuron/sim/scopmath/newton_thread.hpp
olupton: They're not **just** copies, I made some defensive changes with symbol/function names, as now these headers will be included in many different .cpp files (including translated MOD files), so there is more scope for clashes.

For example, see https://github.com/BlueBrain/CoreNeuron/blob/8c42d359aeea351536d1135af14410d1b91b6c39/coreneuron/sim/scopmath/sparse_thread.hpp#L22-L24 and macro renaming + sanity checking https://github.com/BlueBrain/CoreNeuron/blob/8c42d359aeea351536d1135af14410d1b91b6c39/coreneuron/sim/scopmath/sparse_thread.hpp#L429-L433 and various bits of allocation/deallocation/initialisation use `new`/`delete`/constructors now that didn't before.

coreneuron/network/netpar.cpp
```diff
@@ -32,7 +32,9 @@
 #include "coreneuron/mpi/core/nrnmpi.hpp"
 int localgid_size_;
 int ag_send_nspike;
+namespace coreneuron {
 int* nrnmpi_nin_;
+}
```
olupton, line 37: This was because of an ODR clash with NEURON, I think.

coreneuron/io/phase2.cpp
nrnhines: The comment is outdated. The original was:
```
commit 921eabb6f65548d6c1fec8b6761d202d0a830abe (HEAD)
Author: Pramod Kumbhar <pramod.kumbhar@epfl.ch>
Date:   Wed May 25 16:25:38 2016 +0200

    NetSendBuffer_t is copied to GPU and updated back to CPU
    
    Change-Id: If0e3fe8ce74895f4378719465ef3686adcb5127c

diff --git a/coreneuron/nrniv/nrn_setup.cpp b/coreneuron/nrniv/nrn_setup.cpp
index 47c6c945..e61c36a5 100644
--- a/coreneuron/nrniv/nrn_setup.cpp
+++ b/coreneuron/nrniv/nrn_setup.cpp
@@ -1303,7 +1303,7 @@ for (int i=0; i < nt.end; ++i) {
       nrb->_pnt_offset = pnt_offset[type];
 
       // begin with a size of 5% of the number of instances
-      nrb->_size = ml->nodecount/20;
+      nrb->_size = ml->nodecount;
       // or at least 8
       if (nrb->_size < 8) {
         nrb->_size = 8;
```
I believe the earlier version was a space saving measure and referred to the presumption that incoming spikes are rare during a minimum delay interval. I'd replace the comment with
```
    // begin with a size equal to the number of instances
```
It would take a bit of thought to know if the size could, in reality , grow beyond that (e.g. many self events per instance) but there is, I believe, provision for growth in each translated mod file.

olupton: Just to say that the reason this change is here is that I was debugging an issue that was triggered by the growth/reallocation. So: without this change then the reallocation was sometimes happening, and I originally made the change so that the bug was triggered earlier in execution.

I left the change in the PR in because of the inconsistency between code and comment, and wanting to trigger this discussion.

nrnhines: > issue that was triggered by the growth/reallocation.

That is a bit disturbing.  Is growth itself problematic on the GPU or is it straightforward to fix the growth code? But the context of you seeing it in 2022 when the change was made in 2016 is confusing. I could write a nonsense mod file that generates several self events so that the growth code is triggered even though we are starting out with size nodecount.

nrnhines: > nonsense mod file that generates several self events

Not exactly nonsense, but it is always better to minimize the size of the event priority queue by generating a single self event in response to the reception of the previous event rather that generating a flurry of self events all at once. In practice it is rare to have multiple self events outstanding for a given instance. Also, for a given instance only unique flag == 1 self events are movable to a new delivery time.

olupton: Sorry, I mean that I (inadvertently) introduced a new bug "elsewhere" that happened to be triggered by the resizing code; to aid debugging that, I [re-]added the 5% so it crashed more quickly when I was debugging.

> I could write a nonsense mod file that generates several self events so that the growth code is triggered even though we are starting out with size nodecount.

I believe this already happens in the current test suite, in the `reduced_dentate` model, with the caveat:
https://github.com/neuronsimulator/nrn/blob/3cc83cf7b09044a478665490d52acfc31f606036/test/external/reduced_dentate/CMakeLists.txt#L12-L15

(although as I say, this was part of a debugging exercise, so this was an observation in a build that was known to have problems...)

nrnhines: Ok. Understood.

olupton: I updated this to initially allocate `ml->nodecount` entries again.

coreneuron/sim/scopmath/crout_thread.hpp
olupton: Removed that #include, see https://github.com/BlueBrain/CoreNeuron/pull/809#discussion_r865613379 for a description of the changes in these files.

coreneuron/sim/scopmath/ssimplic_thread.hpp
olupton: See https://github.com/BlueBrain/CoreNeuron/pull/809#discussion_r865613379

coreneuron/network/netpar.cpp
pramodk: @olupton : could you bit explain the context for this?

coreneuron/network/tqueue.hpp
```diff
@@ -65,9 +65,10 @@ extern void splay(SPBLK*, SPTREE*);    /* reorganize tree */
 extern SPBLK* sphead(SPTREE*);         /* return first node in tree */
 extern void spdelete(SPBLK*, SPTREE*); /* delete node from tree */
 
+struct DiscreteEvent;
 class TQItem {
   public:
-    void* data_ = nullptr;
+    DiscreteEvent* data_ = nullptr;
```
pramodk, line 71: Just to be sure: all event types are derived from `DiscreteEvent` class and hence the change, right?

coreneuron/io/phase2.cpp
```diff
@@ -482,12 +482,8 @@ void Phase2::set_net_send_buffer(Memb_list** ml_list, const std::vector<int>& pn
             ml->_net_receive_buffer = nrb;
             nrb->_pnt_offset = pnt_offset[type];
 
-            // begin with a size of 5% of the number of instances
+            // begin with a size equal to the number of instances
             nrb->_size = ml->nodecount;
-            // or at least 8
-            nrb->_size = std::max(8, nrb->_size);
-            // but not more than nodecount
-            nrb->_size = std::min(ml->nodecount, nrb->_size);
```
pramodk, line 490: Not sure if there was any strong reason to not have `non-zero` buffer size even if a mechanism is empty. (esp. keeping some min in order to avoid reallocation)

coreneuron/sim/scopmath/crout_thread.hpp
```diff
@@ -0,0 +1,190 @@
...
+ */
+inline void nrn_scopmath_solve_thread(int n,
+                                      double** a,
+                                      double* b,
+                                      int* perm,
+                                      double* p,
+                                      int* y,
+                                      _threadargsproto_) {
+    /* Perform forward substitution with pivoting */
+    // if (y) { // pgacc bug. nullptr on cpu but not on GPU
```
pramodk, line 135: Just a note: looking at bed333d4a6e1088f3e224efdfe04d498362db6a7, looks like this was relevant in ~2016. May be this is not required anymore.

pramodk (commented):
1-2 minor comments but overall LGTM!

coreneuron/network/tqueue.hpp
olupton: Yes..this was a change I made while debugging, and it seems like a straightforward improvement so I left it in.

coreneuron/io/phase2.cpp
olupton: You didn't get that anyway. If `ml->nodecount` was zero then `nrb->_size` was set to `0`, then `8`, then `0` again.

coreneuron/sim/scopmath/crout_thread.hpp
olupton: Maybe not, but apparently it hasn't been causing problems so I didn't think I needed to touch it...

coreneuron/network/netpar.cpp
olupton: `nrnmpi_nin_` is defined at global scope in NEURON: https://github.com/neuronsimulator/nrn/blob/3f2f83d5339223c99f3d710424ffec383d038f27/src/nrnmpi/nrnmpi_def_cinc#L10.

Normally for CoreNEURON we duplicate these variables inside the `coreneuron` namespace, but here we are defining these things at global scope too. This violates the one definition rule, so it's undefined behaviour, and some tool (I think ASan, though I don't have the error message to hand) flagged this.

Looking at it now, some of the other variables here are probably falling into the same trap.

coreneuron/io/phase2.cpp
pramodk: yeah I saw that 🙈! (may be was gradually modified) But just saying that in case there was a reason, we could keep `std::max(8, ml->nodecount);`

coreneuron/sim/scopmath/crout_thread.hpp
pramodk: yeah, we don't need to touch it in this PR.

coreneuron/io/phase2.cpp
olupton: Done

pramodk (approved):
LGTM 🚀

PR #811, opened 2022-05-09
Fix building without MPI.

olupton:
**Description**
Without this, builds without MPI have linker errors about `dlsym` and `dlerror`, which are used in `core/nrnmpi.cpp`.


pramodk (approved):
LGTM

alkino (approved):

PR #823, opened 2022-06-02
Configure sanitizers in CMake

olupton:
**Description**
Use the utilities added in https://github.com/BlueBrain/hpc-coding-conventions/pull/131 to configure (Clang) sanitizers when a new `CORENRN_SANITIZERS` option is passed.

So far AddressSanitizer, LeakSanitizer and UndefinedBehaviourSanitizer have been investigated. LeakSanitizer identifies hard-to-suppress problems in every MPI implementation that I tested, but it did seem to run cleanly without MPI after a few (included) suppressions -- I have not currently included a build of that, though. AddressSantizer runs cleanly and UndefinedBehaviourSanitizer flags one thing from Random123 that I have included in a suppression file.

The new CI builds use Ubuntu 22.04 and Clang 14, one build has AddressSanitizer (with LeakSanitizer disabled at runtime), the other build has UndefinedBehaviourSanitizer. As well as adding these builds, the CI plan has been updated to:
- include problem matchers, so GitHub should flag sanitizer and compiler warnings in the PR diff view (see https://github.com/BlueBrain/CoreNeuron/pull/823#issuecomment-1148483509)
- use ccache to speed up CI builds (note that this is currently required for ☝️ as we rely on ccache's absolute-to-relative path rewriting to generate the ../path/to/source/file.cpp pattern matched by the problem matchers)
- use ninja to speed up CI builds
- use macOS-11 instead of macOS-10.15
- move the documentation and gcc-10 builds out of the matrix so we don't run pointless macOS versions of them

Also:
- Add CORENRN_EXTRA_CXX_FLAGS and CORENRN_EXTRA_MECH_CXX_FLAGS CMake variables for injecting compiler flags to be used when compiling CoreNEURON (or CoreNEURON mechanism) source files, unlike CMAKE_CXX_FLAGS that touches all submodules and other software that is pulled into the build.
- Use `cpp_cc_git_submodule` helper for CLI11
- Fix building documentation with Ninja
- Drop empty `sim/scopmath/crout_thread.cpp`
- Fix a bug in `stdindex2ptr` that was flagged by AddressSanitizer
- Fix some warnings where it seemed easier than suppressing them

TODO:
- [x] Add CI pipeline that uses the new option
- [x] Update submodule when https://github.com/BlueBrain/hpc-coding-conventions/pull/131 is merged
- [x] Update submodule when https://github.com/BlueBrain/nmodl/pull/878 is merged

**How to test this?**
Build CoreNEURON with `-DCORENRN_SANITIZERS=undefined`, `-DCORENRN_SANITIZERS=address` or `-DCORENRN_SANITIZERS=address,leak` and Clang.

**Test System**
 - OS: BB5
 - Compiler: Clang 13
 - Backend: CPU

CI_BRANCHES:NEURON_BRANCH=master,SPACK_BRANCH=develop


coreneuron/io/nrn_setup.cpp
```diff
@@ -635,21 +635,19 @@ void read_phasegap(NrnThread& nt, UserParams& userParams) {
...
-        int ix = index;  // relative to _actual_v
+        int ix{index};  // relative to _actual_v
         nrn_assert((ix >= 0) && (ix < nt.end));
         if (nt._permute) {
             node_permute(&ix, 1, nt._permute);
         }
-        return nt._data + (v0 + ix);    // relative to nt._data
+        return nt._actual_v + ix;
     } else if (mtype == i_membrane_) {  // membrane current from fast_imem calculation
-        int i_mem = nt.nrn_fast_imem->nrn_sav_rhs - nt._data;
```
olupton, line 646: This was leading to segfaults when AddressSanitizer was enabled, as in those builds `nrn_sav_rhs` and `_data` (which are independent allocations) had addresses that were too different to fix in an `int`.

olupton:
With the current state of the PR I see:
Screenshot 2022-06-07 at 12 25 03

alexsavulescu (approved):

coreneuron/io/nrn_setup.cpp
pramodk: 👍 (didn't see any reason for previous logic anyway...)

.sanitizers/undefined.supp
```diff
@@ -0,0 +1 @@
+unsigned-integer-overflow:_philox4x32bumpkey(r123array2x32)
```
pramodk, line 1: curious about this - is this something work reporting upstream or false-positive?

pramodk (approved):
LGTM

tests/CMakeLists.txt
```diff
@@ -9,6 +9,12 @@ include(TestHelpers)
 include_directories(${CORENEURON_PROJECT_SOURCE_DIR} ${CORENEURON_PROJECT_BINARY_DIR}/generated
                     ${Boost_INCLUDE_DIRS})
 
+# Add compiler flags that should apply to all CoreNEURON targets, but which should not leak into
+# other included projects.
```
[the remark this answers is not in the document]
olupton: https://github.com/BlueBrain/CoreNeuron/blob/940d0e7802e0dad1c5f55084797914b52314f7d4/coreneuron/CMakeLists.txt#L7-L11
the same lines are there in `coreneuron/CMakeLists.txt` 🙂

.sanitizers/undefined.supp
olupton: It's not actually undefined behaviour, see https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
> `-fsanitize=unsigned-integer-overflow`: Unsigned integer overflow, where the result of an unsigned integer computation cannot be represented in its type. Unlike signed integer overflow, this is not undefined behavior, but it is often unintentional. This sanitizer does not check for lossy implicit conversions performed before such a computation (see `-fsanitize=implicit-conversion`).

So this isn't a big red flag. I haven't yet investigated what's going on in the cases caught by our test suite; I thought it was better to get these checks in and then incrementally try to remove the suppressions in other, smaller PRs.

olupton: I disabled this check in the NEURON PR because it was too noisy, BTW: https://github.com/neuronsimulator/nrn/blob/e347a9dfdcb475c63fc8a34ed7e058f3c2600593/CMakeLists.txt#L250

PR #829, opened 2022-06-19
Allow multiple gids per cell that reference different voltages.

nrnhines:
** TODOs **
- [x] @pramodk is testing Robin's model on BB5

**Description**
Allow a cell to have multiple gids that reference voltage at different locations on the cell.
This change is a companion to neuronsimulator/nrn#1850 and is part of the fix for issue neuronsimulator/nrn#1839
This changes the bbcore_write_version to "1.6"

CoreNEURON NetCon source variable references must still be to voltage. This could be generalized to any range variable as is the case with NEURON.  However, as with the substantive problem of multiple gids per cell, a reference to any range variable in CoreNEURON can also be accomplished with the use of a POINT_PROCESS that WATCHes a range variable. I am not aware of any NEURON model that uses a NetCon source reference to anything except voltage.

**How to test this?**
Build NEURON with ```-DNRN_ENABLE_CORENEURON=ON``` 
```bash
cd test/pynrn
nrnivmodl -coreneuron
python test_multigid.py
```

CI_BRANCHES:NMODL_BRANCH=master,SPACK_BRANCH=develop


alexsavulescu:
please retest this

pramodk (approved):
LGTM

PR #839, opened 2022-07-06
CLI11 & GCC8 combo have special consideration

alexsavulescu:
* add CLI11_HAS_FILESYSTEM=0
 
See "Note: Special instructions for GCC 8" on https://github.com/CLIUtils/CLI11

CI_BRANCHES:NEURON_BRANCH=master,NMODL_BRANCH=wheel,SPACK_BRANCH=develop

### DIFF
--- a/CMake/config/CompilerFlagsHelpers.cmake
+++ b/CMake/config/CompilerFlagsHelpers.cmake
@@ -35,8 +35,7 @@ foreach(COMPILER_LANGUAGE ${SUPPORTED_COMPILER_LANGUAGE_LIST})
     # XLC -qinfo=all is awfully verbose on any platforms that use the GNU STL
     # Enable by default only the relevant one
     # ~~~
-    set(CMAKE_${COMPILER_LANGUAGE}_WARNING_ALL
-        "-qformat=all -qinfo=lan:trx:ret:zea:cmp:ret")
+    set(CMAKE_${COMPILER_LANGUAGE}_WARNING_ALL "-qformat=all -qinfo=lan:trx:ret:zea:cmp:ret")
 
     set(CMAKE_${COMPILER_LANGUAGE}_DEBUGINFO_FLAGS "-g")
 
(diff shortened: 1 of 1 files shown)


CMake/config/CompilerFlagsHelpers.cmake
```diff
@@ -152,3 +152,12 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang" OR ${CMAKE_SYSTEM_NAME} MATCHES "D
...
   string(APPEND CMAKE_SHARED_LIBRARY_CREATE_C_FLAGS " ${UNDEFINED_SYMBOLS_IGNORE_FLAG}")
 endif()
+
+# ===============================================================================
+# See "Note: Special instructions for GCC 8" on https://github.com/CLIUtils/CLI11
+# ===============================================================================
+if(CMAKE_CXX_COMPILER_IS_GCC
+   AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL 8
+   AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 9)
+  add_compile_definitions(CLI11_HAS_FILESYSTEM=0)
```
olupton, line 162: I think this ends up included at the top level in CoreNEURON, that might mean it's inherited by NMODL when that's built as a submodule, but not 100% sure (that might hide the same issue in NMODL in "developer builds").

alexsavulescu: I tested this in the ISPC build and seemed to work ... I may be wrong

olupton: I thought the issue was CoreNEURON crashing? I think this should "definitely" fix that.

I'm wondering why we don't(?) hit the same thing in standalone builds of NMODL + GCC8 when we pass a path to be validated as a commandline argument to `nmodl`. Maybe because NMODL doesn't use the `ExistingDirectory` validator that appeared in your backtrace? 🤷

alexsavulescu: Not sure, kind of tricky with all the CIs and matching stuff

Issue #826, opened 2022-06-16
CoreNEURON does not compile with NVHPC 22.5


olupton:
**Describe the issue**
As first seen in https://github.com/BlueBrain/spack/pull/1606 and subsequently reported in https://github.com/BlueBrain/CoreNeuron/issues/825, CoreNEURON does not compile with `nvc++` 22.5, at least if GPU support is enabled.

This should be fixed, probably in the context of https://github.com/BlueBrain/CoreNeuron/pull/795 which is already revisiting how global variables are handled in GPU code.

olupton:
See also https://github.com/BlueBrain/CoreNeuron/pull/841#issuecomment-1181523812 for information about issues hidden behind the compilation issues.

olupton:
I believe this was closed by https://github.com/BlueBrain/CoreNeuron/pull/795.

PR #841, opened 2022-07-11
Add unit tests for the Hines solver.

olupton:
**Description**
Add some tests of the various CPU/OpenACC/OpenMP/CUDA implementations of the Hines matrix solver.

These use toy input data that is generated on the fly. Tests cover scenarios such as:
- many small cells
- few large cells

both with hardcoded constant values for the matrix elements and pseudorandom number generation.

The results (`d` and `rhs`) from all available solver implementations are compared to one another in each scenario.

See also #826, as some of these tests do not seem to pass at -O0 using NVHPC 22.5. See: https://github.com/BlueBrain/CoreNeuron/pull/841#issuecomment-1181523812.

Also update NMODL to the latest `master`, which now requires Python 3.7+.

**How to test this?**
Build CoreNEURON with tests enabled (and Boost), run the `test-solver` test. Or just execute `bin/test-solver`.

**Test System**
 - OS: BB5
 - Compiler: NVHPC 22.3/22.5
 - Version: this feature branch
 - Backend: GPU

### DIFF
--- a/.github/workflows/coreneuron-ci.yml
+++ b/.github/workflows/coreneuron-ci.yml
@@ -36,9 +36,9 @@ jobs:
           - {cmake_option: "-DCORENRN_ENABLE_MPI_DYNAMIC=ON", flag_warnings: ON}
           - {cmake_option: "-DCORENRN_ENABLE_MPI_DYNAMIC=ON -DCORENRN_ENABLE_SHARED=OFF"}
           - {cmake_option: "-DCORENRN_ENABLE_MPI=OFF"}
-          - {use_nmodl: ON, py_version: 3.6.7}
+          - {use_nmodl: ON, py_version: 3.7}
           - {use_nmodl: ON}
-          - {use_ispc: ON, py_version: 3.6.7}
+          - {use_ispc: ON, py_version: 3.7}
         include:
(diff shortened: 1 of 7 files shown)


olupton:
cc: @nrnhines in case you have any particular suggestions about toy data characteristics/topologies that would provide useful extra coverage.

So far the number/size of cells varies, as do the `a`/`b`/`d`/`rhs` values, but each cell is just a binary tree (possibly a bit unbalanced, depending on the total number of segments per cell that is chosen).

olupton:
With the additional fixes in https://github.com/BlueBrain/CoreNeuron/pull/842, I can build and run the new tests with NVHPC 22.5 and `-O0 -g` plus `-tp haswell` for `nvc`/`nvc++` and `-lineinfo` for `nvcc`.

The `CellPermute2_GPU` configuration does not match the reference (`CellPermute0_CPU`) in the `SingleCellAndThread`, `UnbalancedCellSingleThread`, `LargeCellSingleThread` and `LargeCellSingleThreadRandom` tests, but the ones with smaller cells do pass. It appears that there is some data race issue with parallelisation across a single cell.

At `-O1`, the issue disappears and the tests added in this PR all pass with NVHPC 22.5.

PR #845, opened 2022-08-06
nrnivmodl_core_makefile: fix SDKROOT

alexsavulescu:
**Description**

Needed for neuronsimulator/nrn#1939

CI_BRANCHES:NEURON_BRANCH=fixmf


olupton (approved):
I guess the logic here is that the old code baked in an SDK path from CMake-time that might not be appropriate on a user machine at nrnivmodl-core-time?

alexsavulescu:
> I guess the logic here is that the old code baked in an SDK path from CMake-time that might not be appropriate on a user machine at nrnivmodl-core-time?

Indeed. We were "lucky" the SDKs were there for testing :)

PR #847, opened 2022-08-09
Fixing race condition in cell permute 2

kotsaloscv:


olupton (approved):
LGTM...I guess you have checked locally the correctness and performance with one/both of 22.5/22.7 (I don't think it's straightforward to test that in the automatic CI)?

I'm also wondering whether this change makes it easy to address https://github.com/BlueBrain/CoreNeuron/blob/aef38da7ac29fa931a586812b0880504505c8396/tests/unit/solver/test_solver.cpp#L29 and improve the coverage a little there? What do you think?

kotsaloscv:
> LGTM...I guess you have checked locally the correctness and performance with one/both of 22.5/22.7 (I don't think it's straightforward to test that in the automatic CI)?
> 
> I'm also wondering whether this change makes it easy to address
> 
> https://github.com/BlueBrain/CoreNeuron/blob/aef38da7ac29fa931a586812b0880504505c8396/tests/unit/solver/test_solver.cpp#L29
> 
> and improve the coverage a little there? What do you think?

Yes it is tested with both 22.5 & 22.7. 

Indeed we can include the cell permute 2 CPU version to the test. I will do it.
Thanks for the feedback!

coreneuron/permute/cellorder.cpp
```diff
@@ -600,15 +588,24 @@ void solve_interleaved2(int ith) {
     defined(_OPENACC)
         int nstride = stridedispl[nwarp];
 #endif
-        nrn_pragma_acc(parallel loop gang vector vector_length(
-            warpsize) present(nt [0:1],
+        /* If we compare this loop with the one from cellorder.cu (CUDA version), we will understand
```
kotsaloscv, line 591: @olupton @pramodk 
Here is the explanation on how the loop works in comparison to the CUDA counterpart.

olupton (approved):

pramodk (approved):
LGTM.

(discussed some of the questions with Christos offline)

PR #848, opened 2022-08-10
Header-only mode for Boost unit tests

olupton:
**Description**
The new solver unit test is failing on newer Red-Hat-based distributions:
- https://github.com/neuronsimulator/nrn-build-ci/runs/7758969675?check_suite_focus=true#step:7:34958
- https://github.com/neuronsimulator/nrn-build-ci/runs/7758969737?check_suite_focus=true#step:7:34889
- https://github.com/neuronsimulator/nrn-build-ci/runs/7758969895?check_suite_focus=true#step:7:34954

This seems to be because that test included the Boost unit test framework in header-only mode but then links against an external copy of the library.

https://www.boost.org/doc/libs/1_79_0/libs/test/doc/html/boost_test/usage_variants.html implies that we need to define different macros for linking against shared/static libraries, and https://cmake.org/cmake/help/latest/module/FindBoost.html makes it look difficult to guarantee static and/or shared, so header-only mode seems like the simple way out.

**How to test this?**
Build the tests, run them -- you might get an error, or you might need to run them in `valgrind` to see it.


olupton:
https://github.com/neuronsimulator/nrn-build-ci/actions/runs/2831964189 is testing this

pramodk (approved):
LGTM

PR #850, opened 2022-08-24
Use last hpc-coding-conventions formatter script

alkino:
Usa last version of hpc-coding-conv

### DIFF
--- a/.bbp-project.yaml
+++ b/.bbp-project.yaml
@@ -1,0 +1,8 @@
+tools:
+  ClangFormat:
+    enable: True
+    include:
+      match:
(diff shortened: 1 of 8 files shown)


olupton:
We need to translate this: https://github.com/BlueBrain/CoreNeuron/blob/4a25a6bb789b84456e9f220d5978b9c10bcbdf49/CMakeLists.txt#L88-L98 for the new hpc-coding-conventions.

For the single Python file, I guess we can just commit the updated version.

pramodk (approved):
looks reasonable to me!

PR #854, opened 2022-08-31
Link to libdl.so to avoid linking issues with mpich on centos7

pramodk:
**Description**

Main coreneuron library needs to be linked to libdl.so
only in case of dynamic mpi build. But on an old system
like centos7, we saw mpich library require explicit
ink to libdl.so.

See https://github.com/neuronsimulator/nrn-build-ci/pull/51

**How to test this?**

CI is launched via nrn-build-pipeline: https://github.com/neuronsimulator/nrn-build-ci/pull/51


pramodk:
Duplicate of https://github.com/BlueBrain/CoreNeuron/pull/856

PR #861, opened 2022-09-22
Allow for explicit non-aligned allocation

ohm314:
- Protect aligned_alloc from potentially invalid 0 alignment argument
- Handle alignment = 0 as normal malloc
- Allocate MembList and ThreadMembList objects without alignment. They are small objects and cannot be allocated with alignment when using the std::aligned_alloc function (without extra fill).


ohm314:
Getting some benchmarks here: https://bbpgitlab.epfl.ch/hpc/cellular/coreneuron-perf-tests/-/pipelines/75464

ohm314:
@olupton : I forgot to merge this one, could you take a look please?

olupton (approved):

PR #863, opened 2022-09-26
Undo std::aligned_alloc introduction

ohm314:
turns out it's slower than posix_memalign

Just to be sure I'm running the benchmark [CI](https://bbpgitlab.epfl.ch/hpc/cellular/coreneuron-perf-tests/-/pipelines/75916).


alexsavulescu (dismissed):

ohm314:
hum.. it didn't get faster again... so...

ohm314:
OK - so finally the performance degradation must have been for some other reason and it was coincidence, but the performance is back to normal and it looks like all is good. I'm not going to roll back the change then.

image

PR #868, opened 2022-10-14
Add a fatal error if user give CORENRN_NMODL_DIR and nmodl not found …

alkino:
If the user specify nmodl he want it of course.


iomaganaris (approved):
LGTM

PR #875, opened 2022-10-27
Event callback slightly different per report

jorblancoa:
**Description**

Enforce deterministic report events

### DIFF
--- a/coreneuron/io/reports/report_configuration_parser.cpp
+++ b/coreneuron/io/reports/report_configuration_parser.cpp
@@ -155,6 +155,7 @@ std::vector<ReportConfiguration> create_report_configurations(const std::string&
             // extra new line: skip
             report_conf.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
         }
+        report.report_index = i;
         reports.push_back(report);
     }
     // read population information for spike report
--- a/coreneuron/io/reports/report_event.cpp
+++ b/coreneuron/io/reports/report_event.cpp
@@ -24,12 +24,14 @@ ReportEvent::ReportEvent(double dt,
                          double tstart,
                          const VarsToReport& filtered_gids,
                          const char* name,
-                         double report_dt)
+                         double report_dt,
+                         int report_index)
     : dt(dt)
     , tstart(tstart)
     , report_path(name)
     , report_dt(report_dt)
-    , vars_to_report(filtered_gids) {
+    , vars_to_report(filtered_gids)
+    , report_index_(report_index) {
     nrn_assert(filtered_gids.size());
     step = tstart / dt;
     reporting_period = static_cast<int>(report_dt / dt);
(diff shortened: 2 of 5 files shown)


jorblancoa:
still issues when launching with libsonata-report master
https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/jobs/409163

coreneuron/io/reports/report_event.cpp
```diff
@@ -88,8 +90,11 @@ void ReportEvent::deliver(double t, NetCvode* nc, NrnThread* nt) {
                                 gids_to_report.data(),
                                 report_path.data());
 #endif
-        send(t + dt, nc, nt);
+        // Deterministic event time per report to avoid deadlocks
+        send(t + dt + report_t_shift_, nc, nt);
```
ohm314, line 94: I guess that's the only way to do it without changing the API of reportinglib? it's a bit ugly that we need to add this artificial `1e-6*report_idx` to fix the order.

coreneuron/io/reports/report_configuration_parser.cpp
```diff
@@ -155,6 +155,7 @@ std::vector<ReportConfiguration> create_report_configurations(const std::string&
             // extra new line: skip
             report_conf.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
         }
+        report.report_index = report.format == "SONATA" ? i : 0;
```
pramodk, line 158:
```suggestion
        report.report_index =  i;
```

I think we don't need to have this specialised for sonata only?

coreneuron/io/reports/report_event.cpp
```diff
@@ -24,12 +24,14 @@ ReportEvent::ReportEvent(double dt,
...
-                         double report_dt)
+                         double report_dt,
+                         int report_index)
     : dt(dt)
     , tstart(tstart)
     , report_path(name)
     , report_dt(report_dt)
-    , vars_to_report(filtered_gids) {
+    , vars_to_report(filtered_gids)
+    , report_t_shift_(1e-6 * report_index) {
```
pramodk, line 34: just to avoid hardcoded factor - what if we replace `1e-6` with  `dt/1e-3`? i.e. whatever will be the value of dt in the future, we can always have this factor 1e3 times lower.

pramodk (commented):
Some minor comments otherwise LGTM

coreneuron/io/reports/report_configuration_parser.cpp
jorblancoa: This was causing differences in the binary reports in the CI.

pramodk: Hmm ok! 

How many such reports are different? the new logic of "determinist ordering" of report recording seems cleaner right? In that case, I wonder if we should just change the report files.

PR #877, opened 2022-10-28
Tuning the CI params to avoid GPU out-of-mem issues

kotsaloscv:


.gitlab-ci.yml
```diff
@@ -66,6 +66,7 @@ simulation_stack:
 .gpu_node:
   variables:
     bb5_constraint: volta
+    bb5_ntasks: 8
```
olupton: I don't know what the best approach will prove to be, but...some thoughts:
- We enable `tqperf-heavy` in the GitLab CI, which means that we run those tests with 16 ranks: https://github.com/neuronsimulator/nrn/blob/d4254f01db6066041ce4df19b3bf3d0e103d78f7/test/external/tqperf/CMakeLists.txt#L10 -- this implies that we need at least 16 tasks in the allocation
- We could set `bb5_cpus_per_task` to a number larger than 1 if we want to occupy a larger fraction of a node (and thereby reduce how many other jobs are assigned concurrently) without increasing the number of tasks (increasing the number of tasks will change how many tests are launched in parallel by CTest inside the test jobs). Tuning the requested memory would achieve something similar.
- There are some shortcomings with how "subjobs" inside an allocation are pinned in this GitLab context: https://bbpteam.epfl.ch/project/issues/browse/HELP-15614
- There are some ongoing efforts to integrate the GPU nodes and MPS better with Slurm: https://bbpteam.epfl.ch/project/issues/browse/INFRA-6657
- If needed, we have the `bb5_exclusive` option to request exclusive allocations...

pramodk: Do we know the reason? i.e. we suspect this is something after maintenance? (Or, from crout solver?)

kotsaloscv: @pramodk there are many tasks loading the GPU and it fails. This is for sure not related to crout but to overutilization. Actually, with `bb5_cpus_per_task: 2` and not `bb5_ntasks: 8` the problem seems to be fixed, i.e. less tasks accessing the GPU at the same time.

kotsaloscv:
@olupton it seems that the `bb5_cpus_per_task: 2` fixes the out-of-memory issue.
Shall we use it for now?

.gitlab-ci.yml
iomaganaris: I'm curious about this because in the past we used 40 ranks and MPS to run simulations. Is this something reproducible on BB5 or just a problem in the CI?

kotsaloscv: maybe the absense of MPS causes the issue? Not sure though.

olupton (approved):
LGTM as a workaround...

PR #878, opened 2022-11-02
Change num_spikes datatype to unsigned long

jorblancoa:
**Description**

MMB 70M cell simulation was failing when writing out.dat spikes.
This PR fixes the issue by changing the datatype of the number of spikes and the buffer size to unsigned long.

Jira ticket: https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-941


iomaganaris (approved):
LGTM :+1: 
I think that a similar thing needs to be fixed in some other parts of the code, for example where we report the total number of synapses.

pramodk (approved):

PR #880, opened 2022-11-03
Change num_spikes datatype to unsigned long (#878)

jorblancoa:
* Change datatype from the number of spikes and buffer size to unsigned long
* Change unsigned long for size_t

CI_BRANCHES:NEURON_BRANCH=release/8.2


pramodk (approved):

pramodk:
NMODL tag-related usage is failing but that change could happen independently later.

PR #882, opened 2022-11-08
Improve memory usage in report initialization

jorblancoa:
**Description**

A high memory usage was taking place in the report initialization when reporting big targets.

**Solution**

Use a std::vector instead of a std::set to store the target gids and do the intersection between the target and the NrnThread gids to loop through them instead of searching in the full target.

### DIFF
--- a/coreneuron/io/reports/nrnreport.hpp
+++ b/coreneuron/io/reports/nrnreport.hpp
@@ -19,6 +19,7 @@
 #include <set>
 #include <unordered_map>
 #include <cstdint>
+#include <malloc.h>
 
 #define REPORT_MAX_NAME_LEN     256
 #define REPORT_MAX_FILEPATH_LEN 4096
@@ -102,6 +103,12 @@ struct ReportConfiguration {
     int num_gids;                         // total number of gids
     int buffer_size;                      // hint on buffer size used for this report
     std::vector<int> target;              // list of gids for this report
+
+    ~ReportConfiguration() {
+        // Force cleanup of the std::vector and release the memory
+        std::vector<int>().swap(target);
+        malloc_trim(0);
+    }
 };
 
 void setup_report_engine(double dt_report, double mindelay);
--- a/coreneuron/io/reports/nrnreport.hpp
+++ b/coreneuron/io/reports/nrnreport.hpp
@@ -19,7 +19,6 @@
 #include <set>
 #include <unordered_map>
 #include <cstdint>
-#include <malloc.h>
 
 #define REPORT_MAX_NAME_LEN     256
 #define REPORT_MAX_FILEPATH_LEN 4096
@@ -103,12 +102,6 @@ struct ReportConfiguration {
     int num_gids;                         // total number of gids
     int buffer_size;                      // hint on buffer size used for this report
     std::vector<int> target;              // list of gids for this report
-
-    ~ReportConfiguration() {
-        // Force cleanup of the std::vector and release the memory
-        std::vector<int>().swap(target);
-        malloc_trim(0);
-    }
 };
 
 void setup_report_engine(double dt_report, double mindelay);
--- a/coreneuron/io/reports/nrnreport.hpp
+++ b/coreneuron/io/reports/nrnreport.hpp
@@ -101,7 +101,7 @@ struct ReportConfiguration {
     double stop;                          // stop time of report
     int num_gids;                         // total number of gids
     int buffer_size;                      // hint on buffer size used for this report
-    std::set<int> target;                 // list of gids for this report
+    std::vector<int> target;              // list of gids for this report
 };
 
 void setup_report_engine(double dt_report, double mindelay);
--- a/coreneuron/io/reports/report_event.hpp
+++ b/coreneuron/io/reports/report_event.hpp
@@ -20,7 +20,7 @@ namespace coreneuron {
 
 #if defined(ENABLE_BIN_REPORTS) || defined(ENABLE_SONATA_REPORTS)
 struct VarWithMapping {
-    int id;
+    uint32_t id;
     double* var_value;
     VarWithMapping(int id_, double* v_)
         : id(id_)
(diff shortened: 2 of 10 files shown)


coreneuron/io/reports/nrnreport.hpp
```diff
@@ -101,7 +102,13 @@ struct ReportConfiguration {
     double stop;                          // stop time of report
     int num_gids;                         // total number of gids
     int buffer_size;                      // hint on buffer size used for this report
-    std::set<int> target;                 // list of gids for this report
+    std::vector<int> target;              // list of gids for this report
+
+    ~ReportConfiguration() {
+        // Force cleanup of the std::vector and release the memory
+        std::vector<int>().swap(target);
+        malloc_trim(0);
```
pramodk, line 110: bit confused by this - I kind of see why this is done. But I wonder if this type of pattern we really need to do in C++ part (especially if we know we are releasing a large vector). Thoughts? 🤔

jorblancoa: I did this because otherwise I didnt see the memory being released...but maybe you are right and we should let the system figure it out

olupton: I'm opposed to scattering things like this throughout the code...

ohm314: I strongly agree with @olupton , we should not be doing this left and right. I can see how at one place at the end of the model building in neurodamus that made sense, but this should not become the norm.

coreneuron/io/reports/report_handler.cpp
```diff
@@ -13,18 +13,40 @@
 
 namespace coreneuron {
 
-void ReportHandler::create_report(double dt, double tstop, double delay) {
+std::vector<int> intersection_gids(const NrnThread& nt, std::vector<int>& target_gids) {
```
ohm314, line 16: maybe a stupid question, but you sure those should be `int`s and not `std::size_t` ?

ohm314 (changes requested):

coreneuron/io/reports/report_handler.cpp
jorblancoa: indeed at the end in libsonata-report we have a uint64_t for the node_id 
```
int sonata_add_node(const char* report_name,
                    const char* population_name,
                    uint64_t population_offset,
                    uint64_t node_id) {
```
However in neurodamus we write int from the CoreConfig.mod
```
int *gids = (int*) calloc(num_gids, sizeof(int));
...
fwrite(gids, sizeof(int), num_gids, fp);
```
And then the gids that each nrnthread has are also stored as an int:
```
class PreSyn: public ConditionEvent {
    ...
    int gid_{-1};
```
Maybe we can convert those to uint64_t for the intersection vector?

coreneuron/io/reports/report_handler.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include <memory>
 #include <vector>
+#include <unordered_set>
```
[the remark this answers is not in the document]
jorblancoa: indeed this was needed before for another implementation of the intersection that was using unordered_set

coreneuron/io/reports/report_event.hpp
```diff
@@ -20,15 +20,15 @@ namespace coreneuron {
 
 #if defined(ENABLE_BIN_REPORTS) || defined(ENABLE_SONATA_REPORTS)
 struct VarWithMapping {
-    int id;
+    uint32_t id;
     double* var_value;
     VarWithMapping(int id_, double* v_)
```
iomaganaris, line 25: Also I see the constructor being call with `int`s, so is this something that needs to be changed in all the calls to `VarWithMapping`?
I see the `section_id`s being `int`s around the code also. Is this because we pass these as `int`s from neurodamus? Should this be changed as well?

iomaganaris (commented):
Overall LGTM.
Some questions/suggestions for improvement

coreneuron/io/reports/report_handler.cpp
iomaganaris: Until we go to human brain scales I guess it's okay the gids to be `int`s?

coreneuron/io/reports/report_event.hpp
jorblancoa: We would have to change everywhere (also in neurodamus for the gids)
For now i just changed the obvious places before we call the libsonata-report functions

Do you think we should change it everywhere? @pramodk  do you have an opinion?

coreneuron/io/reports/report_handler.cpp
ohm314: Ah, so in sonata we are already storing the gids as `uint_64` ? Then probably having all in that type would be more consistent - but only danger here is that we end up needing more memory and we're already fighting this...

coreneuron/io/reports/report_event.hpp
ohm314: i think this is neuron legacy coming at us, but really for all these ids it would make much more sense to work with unsigned.

ohm314 (approved):
I think this looks good now!

coreneuron/io/reports/report_handler.cpp
```diff
@@ -13,18 +13,40 @@
 
 namespace coreneuron {
 
-void ReportHandler::create_report(double dt, double tstop, double delay) {
+std::vector<uint64_t> intersection_gids(const NrnThread& nt, std::vector<int>& target_gids) {
+    std::vector<int> thread_gids;
+    for (int i = 0; i < nt.ncell; i++) {
+        thread_gids.push_back(nt.presyns[i].gid_);
+    }
+    std::vector<uint64_t> intersection;
```
pramodk, line 21: as this is for saving memory - as vectors are int, `intersection` can be also int instead of uint64_t? (but I assume you used the later for future proofing)

pramodk (commented):
also one minor comment otherwise LGTM

coreneuron/io/reports/report_handler.cpp
ohm314: `uint32_t` you mean? :)

pramodk: make that template function and return with the same type as the arguments 😎

pramodk (approved):
LGTM

PR #886, opened 2022-11-23
Assertions should not have side effects

olupton:
**Description**
Some things I came across when debugging failures with `-DNDEBUG`.

### DIFF
--- a/coreneuron/io/core2nrn_data_return.cpp
+++ b/coreneuron/io/core2nrn_data_return.cpp
@@ -546,7 +546,8 @@ void core2nrn_tqueue(NrnThread& nt) {
     }
     // TQitems from binq_
     for (q = tqe->binq_->first(); q; q = tqe->binq_->next(q)) {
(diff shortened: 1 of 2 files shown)


coreneuron/io/core2nrn_data_return.cpp
```diff
@@ -546,7 +546,8 @@ void core2nrn_tqueue(NrnThread& nt) {
     }
     // TQitems from binq_
     for (q = tqe->binq_->first(); q; q = tqe->binq_->next(q)) {
-        assert(core2nrn_tqueue_item(q, sewm, nt) == false);
+        bool const result = core2nrn_tqueue_item(q, sewm, nt);
+        assert(result == false);
```
ohm314, line 550: why did you not replace here also with `nrn_assert`?

olupton: No strong reason, I think just that the header wasn't already included / it wasn't already used in this file...

PR #887, opened 2022-11-29
Use action for coverage instead of script

alkino:


pramodk:
@alkino : I suggest cancel this PR as this repo will be read-only from tomorrow/this week as we have migrated all code into NEURON: https://github.com/neuronsimulator/nrn/pull/2055

